PR #6, opened 2017-07-21
Issue 4

benjamin-work:
* Implement a `PrintLog` class.
* Add tests for callbacks

### DIFF
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -1,10 +1,15 @@
+from itertools import cycle
+from numbers import Number
 import operator
+import sys
 import time
 
 import numpy as np
 from sklearn.base import BaseEstimator
 from sklearn import metrics
+from tabulate import tabulate
 
+from inferno.utils import Ansi
 from inferno.utils import to_numpy
 from inferno.utils import to_var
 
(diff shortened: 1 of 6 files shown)


benjamin-work:
Addresses #4 
@ottonemo pls review

inferno/callbacks.py
```diff
@@ -139,11 +144,85 @@ def on_batch_end(self, net, X, y, train):
...
+        headers = [key for key in self.keys if not key.endswith('_best')]
+        return tabulate(
+            formatted,
+            headers=headers,
+            tablefmt=self.tablefmt,
+            floatfmt=self.floatfmt,
+        )
+
+    def on_epoch_end(self, net, *args, **kwargs):
+        data = net.history[-1:, self.keys]
```
ottonemo: What if the key does not exist in the history? This would raise a `KeyError` here.

benjamin-work: What would be the best solution here? There is no earlier point in time where we could detect missing keys.

ottonemo: True, but maybe we handle this case a bit more gracefully? I.e. writing a helpful error message explaining the issue instead. For example: "Key <x> was not found in the history which indicates that the event never occurred or you misspelled the key name."?

benjamin-work: Is this better now?

PR #10, opened 2017-07-21
Feature/neural net regressor

benjamin-work:
NeuralNetRegressor class implemented.

Addresses #8 
Partly hinges on https://github.com/pytorch/pytorch/issues/2178

PS: I failed at squashing, please forgive.


inferno/net.py
```diff
@@ -413,3 +417,29 @@ def __init__(
...
+            criterion=criterion,
+            *args,
+            **kwargs
+        )
+
+    def check_data(self, _, y):
+        # The problem with 1-dim float y is that the pytorch DataLoader will
+        # somehow upcast it to DoubleTensor
+        if get_dim(y) == 1:
+            raise ValueError("The target data shouldn't be 1-dimensional; "
```
ottonemo: Better `The target data shouldn't be 1-dimensional [because the pytorch DataLoader will upcast it to a DoubleTensor in this case]`?

inferno/utils.py
```diff
@@ -64,3 +64,10 @@ def check_history_slice(history, sl):
         msg = ("Key '{}' could not be found in history; "
                "maybe there was a typo?".format(exc.args[0]))
         raise KeyError(msg)
+
+
+def get_dim(y):
```
ottonemo: Does it return the dimensions or a specific one (which one?)?

inferno/net.py
benjamin-work: Don't you think that are too many implementation details here?

inferno/utils.py
benjamin-work: The dimensionality. I will add a docstring for clarification.

inferno/net.py
ottonemo: If we say the data "shouldn't be [this way]" then we should at least give a reason why this is the case. If this is done to prevent the user from harm even more so, so that the user understands the consequences and can remodel his data appropriately. If we just say 'your data is wrong' then there is no understanding what is happening.

benjamin-work: But then the explanation above is not satisfactory, since it neither explains why DataLoader is suddenly relevant, nor why applying DataLoader would have this consequence. For the latter, we don't even know the explanation, so we would have to say: 'Because that's how it is'. Which is why we can as well save the user some time and directly put it as is.

In my opinion, an exception does not need to give a reason, it should say what is wrong and what can be done about it. Reasons should be left to code comments.

ottonemo: > we don't even know the explanation, so we would have to say: 'Because that's how it is'

I was under the impression we do. In that case you are right and we shouldn't mention this at all then.

benjamin-work: > I was under the impression we do

We know the line of code that causes the problem but we don't even know if the effect is intended.

If it's fine with you, I merge.

PR #12, opened 2017-07-24
Add RNN language model example

ottonemo:
Needed to fix `to_tensor` to make it work.


examples/word_language_model/README.md
```diff
@@ -0,0 +1,18 @@
...
+checkpoints of the model in a file (default is `./model.pt`).
+
+### Generation:
+
+	python generate.py
+
+The default setting is that during training in `train.py` a checkpoint model
+is written to `./model.pt` by default. This model checkpoint is also used
+for generation in `generate.py`.
+
```
benjamin-work: Maybe add that by default, the output is written to the generated.txt file.

examples/word_language_model/learner.py
```diff
@@ -0,0 +1,92 @@
...
+        """Wraps hidden states in new Variables, to detach them from their history."""
+        if isinstance(h, Variable):
+            v = Variable(h.data)
+            return v.cuda() if self.use_cuda else v
+        else:
+            return tuple(self.repackage_hidden(v) for v in h)
+
+    def on_train_begin(self, net, *args, **kwargs):
+        self.hidden = self.module_.init_hidden(self.batch_size)
+        self.module_.cuda()
```
benjamin-work: Shouldn't this depend on `self.use_cuda`? And do we not need this in the base class?

examples/word_language_model/learner.py
```diff
@@ -0,0 +1,92 @@
...
+
+        hidden = self.module_.init_hidden(X.size(1))
+        output, hidden = self.module_(X, hidden)
+
+        word_weights = output.view(-1, self.ntokens)
+        word_idx = torch.multinomial(word_weights, 1)
+
+        return word_idx
+
+    def forward(self, X, training_behavior=False):
```
benjamin-work: This method is almost identical to the one it overrides. I wonder if that means that we should facilitate this. Maybe instead of having `self.module_(x)`, have `self.infer(x)` (== `evaluation_step`) that by default does the former? Then `forward` could use this and you would just have to change `self.infer`.

examples/word_language_model/learner.py
```diff
@@ -0,0 +1,92 @@
...
+        y_probas = []
+        for x in iterator:
+            x = inferno.utils.to_var(x, use_cuda=self.use_cuda)
+            y_probas.append(self.evaluation_step(x))
+        return torch.cat(y_probas, dim=0)
+
+    def score(self, X, y):
+        # TODO: we cannot use predict() directly as the y supplied by GridSearchCV
+        # is not a "valid" y and only based on the input given to fit() down below.
+        # Therefore we have to generate our own batches.
```
benjamin-work: Is there an easy way to make it so that the provided `y` is actually valid? Do we maybe need a method that just applies the batching to `X` and/or `y`?

examples/word_language_model/README.md
```diff
@@ -0,0 +1,18 @@
...
+Mostly adapted from [here](https://github.com/pytorch/examples/tree/master/word_language_model).
```
benjamin-work: Maybe a sentence or 2 about what is just re-used, what was added, what inferno contributed.

examples/word_language_model/train.py
```diff
@@ -0,0 +1,70 @@
...
+    {
+        'lr': [10,20,30],
+    },
+]
+
+pl = GridSearchCV(learner, params)
+pl.fit(corpus.train[:1000], corpus.train[:1000])
+
+import tabulate
+print(tabulate.tabulate(pl.cv_results_, headers='keys', tablefmt='psql'))
```
benjamin-work: At least for me, the resulting table looks rather ugly. Using a less spacious table format could help a little bit. Using fewer splits would also help (e.g. `cv=3`).

What is also nice is that you can pass the `cv_results_` directly to `pandas.DataFrame` to get a nice df and its prettier print output. However, we don't want a pandas dependency just for this ...

*Edit*: Maybe don't print the `cv_results_` but just `best_params_` and `best_score_`?

inferno/net.py
```diff
@@ -178,7 +178,7 @@ def _yield_callbacks(self):
                 if isinstance(cb, type):  # uninitialized:
                     name = cb.__name__
                 else:
-                    name = item.__class__.__name__
+                    name = cb.__class__.__name__
```
benjamin-work: Was this been a bug previously?

benjamin-work:
Just in general, can we have 2 empty lines after the last import and between class/function definitions?

inferno/net.py
ottonemo: No it is equivalent to the version before, it is just neater to look at.

examples/word_language_model/README.md
ottonemo: > what inferno contributed.

Yeah, as soon as we know for sure what that is and what it isn't. This is too much of a WIP to know for certain now.

examples/word_language_model/train.py
ottonemo: Fixed. Using `best_*` now.

examples/word_language_model/learner.py
ottonemo: In this case `evaluation_step` does a bit more than that than just calling the model, for example doing some post-processing to handle the probabilities. Maybe we should introduce this as well, analogous to `train_`- and `validate_`-step?

examples/word_language_model/learner.py
ottonemo: Yep this was fixed already.

examples/word_language_model/README.md
ottonemo: Fixed.

PR #16, opened 2017-07-25
docstrings

benjamin-work:
Addresses #15

### DIFF
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -46,6 +46,7 @@ class Callback:
     def set_params(self, **params):
         for key, val in params.items():
             setattr(self, key, val)
+        return self
 
 
 class EpochTimer(Callback):
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -1,3 +1,5 @@
+"""Contains callback base class and callbacks."""
+
 from itertools import cycle
 from numbers import Number
 import operator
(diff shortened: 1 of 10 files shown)


inferno/callbacks.py
```diff
@@ -16,25 +18,48 @@
 
 
 class Callback:
+    """Base class for callbacks.
+
+    All callbacks should inherit from this class. It implements all
+    existing methods on callbacks, but the methods are allowed not to
```
ottonemo: `are not allowed to `

inferno/callbacks.py
```diff
@@ -125,10 +190,35 @@ def on_epoch_end(self, net, **kwargs):
...
+
+    Parameters
+    ----------
+    name : str (default='myscore')
+      The name of the score. Determines the column name in the
+      history.
+
+    scoring : None, str, or callable (default=None)
+      If None, use the `score` method of the model. If str, it should
+      be a valid skearn metric (e.g. "f1_score", "accuracy_score"). If
```
ottonemo: *sklearn

inferno/callbacks.py
benjamin-work: But that would be a different meaning.

inferno/callbacks.py
```diff
@@ -162,6 +254,40 @@ def on_batch_end(self, net, X, y, train):
 
 
 class PrintLog(Callback):
+    """Print out useful information from the model's history.
+
+    By default, this will print the epoch, train loss, valid loss, and
+    epoch duration. In addition, `PrintLog` will take care of
+    highlighting the best loss at each epoch.
+
+    To determine the best loss, `PrintLog` looks for a keys that end
```
ottonemo: `for keys`

inferno/callbacks.py
ottonemo: OK got it now. Maybe better: "implementation of the methods is optional".

inferno/callbacks.py
```diff
@@ -162,6 +254,40 @@ def on_batch_end(self, net, X, y, train):
...
+    on `'_best'` and associates them with the corresponding
+    loss. E.g., `'train_loss_best'` will be matched with
+    `'train_loss'`. Therefore, `PrintLog` works best in conjunction
+    with the `BestLoss` callback.
```
ottonemo: "Therefore, `PrintLog` works best in conjunction with the `BestLoss` callback which adds these keys automatically." (give reason why it works best in conjunction)

inferno/tests/test_callbacks.py
```diff
@@ -205,6 +205,20 @@ def score_func(estimator, X, y):
         expected = [555] * 5
         assert np.allclose(acc, expected)
 
+    def test_scoring_func_none(self, scoring_cls, net):
+        net.score = Mock(return_value=345)
+        scoring = scoring_cls(
+            name='acc',
+            scoring=None,
+        )
+        for x, y in zip(np.arange(5), np.arange(5)[::-1]):
```
ottonemo: Personally I like `reversed(np.arange(5))` better for readability although I regularly fail to use it as well.

benjamin-work:
@ottonemo Please have a look at the changes.

ottonemo:
LGTM

PR #20, opened 2017-07-27
Dataset

benjamin-work:
Addresses #19

### DIFF
--- a/inferno/dataset.py
+++ b/inferno/dataset.py
@@ -1,0 +1,136 @@
+from functools import partial
+
+from sklearn.utils import safe_indexing
+import torch
+
+from inferno.utils import to_tensor
+
+
+def _apply_to_data(data, func, unpack_dict=False):
+    _apply = partial(_apply_to_data, func=func, unpack_dict=unpack_dict)
+    if isinstance(data, dict):
+        if unpack_dict:
+            return [_apply(v) for v in data.values()]
+        else:
+            return {k: _apply(v) for k, v in data.items()}
+    elif isinstance(data, (list, tuple)):
+        try:
+            # e.g.list/tuple of arrays
+            return [_apply(x) for x in data]
+        except TypeError:
+            return func(data)
+    return func(data)
+
+
+def _flatten(arr):
+    for item in arr:
+        if isinstance(item, (tuple, list)):
+            yield from _flatten(item)
+        else:
+            yield item
+
+
+def get_len(data):
+    lens = [_apply_to_data(data, len, unpack_dict=True)]
+    lens = list(_flatten(lens))
+    len_set = set(lens)
+    if len(len_set) != 1:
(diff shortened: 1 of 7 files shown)


inferno/dataset.py
```diff
@@ -0,0 +1,136 @@
...
+    if hasattr(data, 'iloc'):
+        # pandas NDFrame
+        return data.iloc[i]
+    # numpy ndarray, list
+    if isinstance(i, (int, slice)):
+        return data[i]
+    return safe_indexing(data, i)
+
+
+class Dataset(object):
```
ottonemo: Shouldn't this inherit `torch.utils.data.Dataset`?

inferno/dataset.py
```diff
@@ -0,0 +1,136 @@
...
+def get_len(data):
+    lens = [_apply_to_data(data, len, unpack_dict=True)]
+    lens = list(_flatten(lens))
+    len_set = set(lens)
+    if len(len_set) != 1:
+        raise ValueError("Dataset does not have consistent lengths.")
+    return list(len_set)[0]
+
+
+def multi_indexing(data, i):
```
ottonemo: This function should have a docstring as it is not clear at all what it does.

inferno/dataset.py
```diff
@@ -0,0 +1,136 @@
...
+            if len_y != len_X:
+                raise ValueError("X and y have inconsistent lengths.")
+        self._len = len_X
+
+    def __len__(self):
+        return self._len
+
+    def __getitem__(self, i):
+        X, y = self.X, self.y
+        if hasattr(X, 'iloc'):
```
ottonemo: Maybe we should refactor this into `is_pandas_dataframe(X)`? That would be a bit more descriptive. In case we want to support any data structure from pandas we could also name it `is_pandas_datastructure` :)

inferno/dataset.py
```diff
@@ -0,0 +1,136 @@
...
+    if isinstance(i, (int, slice)):
+        return data[i]
+    return safe_indexing(data, i)
+
+
+class Dataset(object):
+    """General dataset wrapper that can be used in conjunction with
+    pytorch's DataLoader.
+
+    The dataset will always yield two values, the data and the
```
ottonemo: "The dataset will always yield two values, the data (`X`) and the target (`y`). [...]"

inferno/dataset.py
benjamin-work: :+1:

inferno/dataset.py
benjamin-work: :+1:

inferno/dataset.py
benjamin-work: :+1:

inferno/dataset.py
benjamin-work: :+1:

PR #25, opened 2017-07-31
rework callbacks

benjamin-work:
I believe that especially if one looks at the changes in `net.py`, the new changes should result in cleaner user code.

### DIFF
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -202,7 +202,7 @@ class BestLoss(Callback):
             yield key, sign, loss
 
     def on_epoch_end(self, net, **kwargs):
-        sl = slice(-1, None), list(self.key_signs)
+        sl = np.s_[-1, list(self.key_signs)]
         check_history_slice(net.history, sl)
 
         history = net.history
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -12,9 +12,9 @@ from sklearn import metrics
 from tabulate import tabulate
 
 from inferno.utils import Ansi
+from inferno.utils import check_history_slice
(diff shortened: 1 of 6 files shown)


inferno/callbacks.py
```diff
@@ -140,46 +153,58 @@ class BestLoss(Callback):
...
                 loss = history[-1, key]
                 yield key, sign, loss
             except KeyError:
                 pass
+        for key, sign in self.key_signs.items():
+            loss = history[-1, key]
+            yield key, sign, loss
 
     def on_epoch_end(self, net, **kwargs):
+        sl = slice(-1, None), list(self.key_signs)
```
ottonemo: Why  not use `np.s_` here as well?

inferno/callbacks.py
```diff
@@ -267,6 +292,10 @@ class PrintLog(Callback):
     `BestLoss` callback takes care of creating those entries, which is
     why `PrintLog` works best in conjunction with that callback.
 
+    *Note*: `PrintLog` will not result in good outputs if the number
```
ottonemo: s/result in/yield/

inferno/callbacks.py
benjamin-work: :+1:

inferno/callbacks.py
benjamin-work: Not sure here. In Python docstrings, I prefer to use "yield" specifically to mean that a function `yield`s things, which is not the case here. Would the change result in more clarity?

ottonemo: Fair point. It is a minor issue at best so feel free to ignore. Pretty sure that people will understand what is meant.

PR #28, opened 2017-08-01
Feature/callbacks optional keys

benjamin-work:
Main changes:
* Callbacks now have an `keys_optional` keyword that allows to exclude keys from check.
* An additional check for duplicate keys has been added.
* Error messages have been improved.

*Note*: The idea of telling the user exactly how to `set_params` to a add a key to `keys_optional` does not work. For that, we would need to know the name of the callback, since the name is part of the `set_params` path. However, we don't know the callback name at the time of the call.

### DIFF
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -379,6 +379,14 @@ class PrintLog(Callback):
       on `'_best'` are used to determine the best corresponding loss
       (see above).
 
+    keys_optional: list of str or None (default=None)
+      If not None, this should be a list of keys whose presence is
+      optional. By default, keys indicated in `key` are mandatory, but
+      sometimes we want optional keys. E.g., if keys refer to
+      validation data, but validation data is not always present,
+      those keys should be optional. When a key that is not optional
+      is missing, an exception will be raised.
+
     sink : callable (default=print)
       The target that the output string is sent to. By default, the
       output is printed to stdout, but the sink could also be a
@@ -409,6 +417,7 @@ class PrintLog(Callback):
     def __init__(
             self,
             keys=None,
+            keys_optional=None,
             sink=print,
             tablefmt='simple',
             floatfmt='.4f',
@@ -417,13 +426,37 @@ class PrintLog(Callback):
             self.keys = []
         else:
             self.keys = [keys] if isinstance(keys, str) else keys
+
+        if keys_optional is None:
+            self.keys_optional = []
+        elif isinstance(keys_optional, str):
+            self.keys_optional = [keys_optional]
+        else:
+            self.keys_optional = keys_optional
+
         self.sink = sink
         self.tablefmt = tablefmt
         self.floatfmt = floatfmt
 
+        self._check_keys_duplicated()
+
+    def _is_optional(self, key):
+        return (key in self.keys_optional) or (key in self.default_keys)
+
+    def _check_keys_duplicated(self):
+        duplicates = _duplicate_items(self.default_keys, self.keys)
+        if duplicates:
+            raise ValueError("PrintLog found duplicate keys: {}"
+                             "".format(', '.join(sorted(duplicates))))
+
     def initialize(self):
         self.first_iteration_ = True
         return self
+
+    def _check_keys_missing(self, history):
+        check_keys = [key for key in self.keys if not self._is_optional(key)]
+        sl = np.s_[-1, check_keys]
+        check_history_slice(history, sl)
 
     def _yield_keys(self, data):
         for key in self.default_keys:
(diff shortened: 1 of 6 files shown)


inferno/callbacks.py
```diff
@@ -172,12 +217,30 @@ class BestLoss(Callback):
...
-    def initialize(self):
+        self._check_keys_duplicated()
+        self._check_signs()
+
+    def _is_optional(self, key):
+        return (key in self.keys_optional) or (key in self.default_key_signs)
+
+    def _check_keys_duplicated(self):
+        duplicates = duplicate_items(self.default_key_signs, self.key_signs)
+        if duplicates:
```
ottonemo: this can be merged into one line

inferno/callbacks.py
```diff
@@ -109,35 +111,70 @@ class AverageLoss(Callback):
...
+                    yield key_loss, list(zip(*row))
+                except KeyError:
+                    pass
+                continue
             row = history[-1, 'batches', :, (key_loss, key_size)]
             yield key_loss, list(zip(*row))
 
+    def _check_keys_duplicated(self):
+        duplicates = duplicate_items(self.default_key_sizes, self.key_sizes)
+        if duplicates:
```
ottonemo: this can be merged into one line

inferno/callbacks.py
```diff
@@ -109,35 +111,70 @@ class AverageLoss(Callback):
...
+        for key_loss, key_size in key_sizes:
+            if self._is_optional(key_loss) or self._is_optional(key_size):
+                try:
+                    row = history[-1, 'batches', :, (key_loss, key_size)]
+                    yield key_loss, list(zip(*row))
+                except KeyError:
+                    pass
+                continue
             row = history[-1, 'batches', :, (key_loss, key_size)]
             yield key_loss, list(zip(*row))
```
ottonemo: This whole inner-loop block can be simplified to

```
try:
    row = ...
    yield ...
except KeyError:
    if not self._is_optional(key_loss) and not self._is_optional(key_size):
        raise
```

inferno/callbacks.py
```diff
@@ -186,24 +249,32 @@ def initialize(self):
...
+        for key_loss, sign in key_signs:
+            if self._is_optional(key_loss):
+                try:
+                    loss = history[-1, key_loss]
+                    yield key_loss, sign, loss
+                except KeyError:
+                    pass
+                continue
+            loss = history[-1, key_loss]
+            yield key_loss, sign, loss
```
ottonemo: this can be simplified similar to the block above (i.e. raise on error, try always)

inferno/callbacks.py
```diff
@@ -303,6 +374,14 @@ class PrintLog(Callback):
       on `'_best'` are used to determine the best corresponding loss
       (see above).
 
+    keys_optional: list of str or None (default=None)
+      If not None, this should be a list of keys whose presence is
+      optional. By default, keys indicated in `key` are mandatory, but
```
ottonemo: "are mandatory [to exist in the history]"

inferno/callbacks.py
```diff
@@ -377,8 +481,16 @@ def table(self, row):
...
         for key, color in zip(self._yield_keys(row), colors):
+            if self._is_optional(key):
+                try:
+                    headers.append(key)
+                    formatted.append(self.format_row(row, key, color=color))
+                except KeyError:
+                    pass
+                continue
             headers.append(key)
             formatted.append(self.format_row(row, key, color=color))
```
ottonemo: This can be simplified similar to the blocks above

inferno/callbacks.py
```diff
@@ -341,14 +421,38 @@ def __init__(
...
         self.floatfmt = floatfmt
 
+        self._check_keys_duplicated()
+
+    def _is_optional(self, key):
+        return (key in self.keys_optional) or (key in self.default_keys)
+
+    def _check_keys_duplicated(self):
+        duplicates = duplicate_items(self.default_keys, self.keys)
+        if duplicates:
```
ottonemo: this can be merged into one line

inferno/callbacks.py
```diff
@@ -109,35 +111,70 @@ class AverageLoss(Callback):
...
 
+    def _check_keys_duplicated(self):
+        duplicates = duplicate_items(self.default_key_sizes, self.key_sizes)
+        if duplicates:
+            raise ValueError("AverageLoss found duplicate keys: {}"
+                             "".format(', '.join(sorted(duplicates))))
+
+    def _check_keys_missing(self, history):
+        check_keys = []
+        for key in flatten(self.key_sizes.items()):
```
ottonemo: Maybe add an explanation why we use `flatten(key_sizes.items())`, e.g.

    # Check value keys and size keys alike.

inferno/callbacks.py
benjamin-work: duplicate assignment is needed later, thus won't fix.

inferno/callbacks.py
benjamin-work: much better, ty :+1:

inferno/callbacks.py
benjamin-work: won't fix (s. above)

inferno/callbacks.py
benjamin-work: won't fix (s. above)

benjamin-work:
@ottonemo please look again

PR #29, opened 2017-08-02
Feature/train test split

benjamin-work:


inferno/dataset.py
```diff
@@ -168,3 +185,140 @@ def __getitem__(self, i):
...
+
+        cv_cls = StratifiedShuffleSplit if self.classifier else ShuffleSplit
+        try:
+            cv = check_cv(
+                cv_cls(test_size=self.cv),
+                y=y,
+                classifier=self.classifier,
+            )
+            stratified = True
+        except ValueError:
```
ottonemo: Where does this value error come from? This should be documented.

inferno/dataset.py
```diff
@@ -168,3 +185,140 @@ def __getitem__(self, i):
...
+                classifier=self.classifier,
+            )
+        return cv, stratified
+
+    def _check_cv(self, y):
+        try:
+            # for stratified split, y must be a numpy array
+            y_arr = to_numpy(y)
+        except AttributeError:
+            y_arr = y
```
ottonemo: Why is it a good idea to take the valuethat just caused an attribute error? Should be documented.

inferno/dataset.py
```diff
@@ -168,3 +185,140 @@ def __getitem__(self, i):
...
+        if isinstance(self.cv, Number):
+            cv_is_float = not float(self.cv).is_integer()
+            if cv_is_float:
+                return self._check_cv_float(y_arr)
+
+        # TODO: Find a better solution for this mess
+        stratified = False
+
+        if y is None:
+            cv = check_cv(self.cv, classifier=self.classifier)
```
ottonemo: If y=None is the default you can remove this special case.

inferno/dataset.py
```diff
@@ -168,3 +185,140 @@ def __getitem__(self, i):
...
+            return cv, stratified
+
+        try:
+            cv = check_cv(self.cv, y_arr, classifier=self.classifier)
+            stratified = True
+        except ValueError:
```
ottonemo: Why would a value error occur here?

inferno/net.py
```diff
@@ -502,17 +504,26 @@ def fit_loop(self, X, y, epochs=None):
...
                 self.notify('on_batch_begin', X=xi, y=yi, train=True)
                 loss = self.train_step(xi, yi, optimizer)
                 self.history.record_batch('train_loss', loss.data[0])
                 self.history.record_batch('train_batch_size', len(xi))
                 self.notify('on_batch_end', X=xi, y=yi, train=True)
 
-            for xi, yi in self.get_iterator(X, y, train=False):
+            if X_valid is None:
+                self.notify('on_epoch_end', X=X, y=y)
+                continue
```
ottonemo: I'm not a fan of repeating the callback call here. I would argue for wrapping the validation loop instead.

inferno/tests/test_dataset.py
```diff
@@ -364,3 +389,276 @@ def test_fit_predict_proba(self, net, data):
...
+                self.dense0 = nn.Linear(input_units, num_units)
+                self.nonlin = nonlin
+                self.dropout = nn.Dropout(0.5)
+                self.dense1 = nn.Linear(num_units, 10)
+                self.output = nn.Linear(10, 2)
+
+            def forward(self, X, **kwargs):
+                X = self.nonlin(self.dense0(X))
+                X = self.dropout(X)
+                X = F.relu(self.dense1(X))
```
ottonemo: Why not use the configured nonlin here as well? OTOH why have it configurable at all?

inferno/tests/test_dataset.py
```diff
@@ -364,3 +389,276 @@ def test_fit_predict_proba(self, net, data):
...
+        assert (mock.call_args_list[1][0][2] == data_split[3]).all()
+
+
+class TestCVSplit:
+    num_samples = 100
+
+    @pytest.fixture
+    def data(self):
+        X = np.random.random((self.num_samples, 10))
+        assert self.num_samples % 25 == 0
```
ottonemo: Why do we expect that?

inferno/tests/test_dataset.py
```diff
@@ -364,3 +389,276 @@ def test_fit_predict_proba(self, net, data):
...
+        y = np.hstack([np.repeat([0, 0, 0], num_expected),
+                       np.repeat([1], num_expected)])
+        X_train, X_valid, y_train, y_valid = cv_split_cls(
+            frac, classifier=True)(X, y)
+        assert y_train.sum() == 0.8 * num_expected
+        assert y_valid.sum() == 0.2 * num_expected
+
+    def test_fraction_no_y(self, cv_split_cls, data):
+        X = data[0]
+        cv = 0.2
```
ottonemo: Maybe parametrise this value?

inferno/net.py
```diff
@@ -312,6 +312,7 @@ def __init__(
             batch_size=128,
             iterator_train=DataLoader,
             iterator_test=DataLoader,
+            train_split=CVSplit(0.2),
```
ottonemo: I was not under the impression that a default split is our default case. Wouldn't you rather use grid searches primarily?

benjamin-work: This is the default in nolearn. Let's postpone the discussion of defaults.

inferno/net.py
benjamin-work: I'm a big fan :p 
No, I don't much care, but this preserves the symmetry.

ottonemo: Alright, we leave it as is for the time being.

inferno/dataset.py
```diff
@@ -72,6 +79,15 @@ def multi_indexing(data, i):
     2  3  6
 
     """
+    if isinstance(i, np.ndarray):
+        if i.dtype == bool:
+            i = i.nonzero()[0]
```
ottonemo: Why `[0]`? What if `i` is meant to influence both dim 0 and 1?

benjamin-work: Could a name an example? That way I could write a test that covers this.

benjamin-work: @ottonemo I think I get what you mean. Is this alright now?

ottonemo: Why not just use `i = i.nonzero()`?

benjamin-work: Because the output is a tuple of numpy arrays but pytorch wants a list of integers as index.

Technically, when the tuple contains more than one list, pytorch will complain anyway, so we could get away with only casting the first element to a list, but that would be kind of a silly optimization. And if pytorch one day allows a tuple of lists, our code does not need to be adjusted.

PR #30, opened 2017-08-02
`y=None` causes trouble in many cases

ottonemo:
in classifier `fit(X, y=None)` causes problems:
-  for inputs with shape `(n,m)` as it will create `y` with shape `(n,1)` whereas shape `(n,)` would be needed
- for inputs with shape `(n,)` as it will try to access the length of each element to create appropriate dummy values (but the child is a number, not a sequence)

I added tests and a fix for these cases, although regression is still missing.

### DIFF
--- a/inferno/net.py
+++ b/inferno/net.py
@@ -792,9 +792,6 @@ class NeuralNetClassifier(NeuralNet):
                              "(and your validation) and supply it using the "
                              "`iterator_train` and `iterator_valid` "
                              "parameters respectively.")
-        elif y is None:
-            # The user implements its own mechanism for generating y.
-            return
 
     def get_loss(self, y_pred, y, train=False):
         y_pred_log = torch.log(y_pred)
--- a/inferno/net.py
+++ b/inferno/net.py
@@ -785,6 +785,17 @@ class NeuralNetClassifier(NeuralNet):
             **kwargs
         )
 
+    def check_data(self, _, y):
+        if y is None and self.iterator_train is DataLoader:
+            raise ValueError("No y-values are given (y=None). You must "
+                             "implement your own DataLoader for training "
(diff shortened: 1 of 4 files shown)


ottonemo:
I think you can review now @benjamin-work.

inferno/net.py
```diff
@@ -785,6 +785,17 @@ def __init__(
...
         )
 
+    def check_data(self, _, y):
+        if y is None and self.iterator_train is DataLoader:
+            raise ValueError("No y-values are given (y=None). You must "
+                             "implement your own DataLoader for training "
+                             "(and your validation) and supply it using the "
+                             "`iterator_train` and `iterator_valid` "
+                             "parameters respectively.")
+        elif y is None:
```
benjamin-work: Why do we need this clause?

inferno/tests/test_dataset.py
```diff
@@ -49,6 +50,175 @@ def test_inconsistent_lengths(self, get_len, data):
...
+    ]
+
+    @pytest.fixture
+    def net_cls_1d_clf(self):
+        class MyModule(nn.Module):
+            def __init__(self):
+                super(MyModule, self).__init__()
+                self.dense = nn.Linear(1,1)
+            def forward(self, X):
+                return F.softmax(self.dense(X.float()))
```
benjamin-work: Maybe it's not realistic, but for these tests to run, we don't need the final nonlinearity. If you drop it, clf and reg would use the same module and you could safe some almost duplicate code.

inferno/net.py
ottonemo: To ignore the following check that access `y`.

ottonemo:
@benjamin-work please review

benjamin-work:
:+1:

PR #36, opened 2017-08-09
Bugfix: pandas test not skipped

benjamin-work:
The reason was that `pytest.mark.skipif` is simply ignored when used in conjuction with a `pytest.fixture`.

### DIFF
--- a/inferno/tests/test_dataset.py
+++ b/inferno/tests/test_dataset.py
@@ -336,9 +336,10 @@ class TestMultiIndexing:
         assert np.allclose(result[2].long().numpy(), np.arange(3, 5))
         assert result[3] == {'a': [4, 5], 'b': [7, 8]}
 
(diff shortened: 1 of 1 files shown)


inferno/tests/test_dataset.py
```diff
@@ -338,9 +336,10 @@ def test_mixed_data_slice(self, multi_indexing):
         assert np.allclose(result[2].long().numpy(), np.arange(3, 5))
         assert result[3] == {'a': [4, 5], 'b': [7, 8]}
 
-    @pytest.mark.skipif(not pandas_installed, reason='pandas is not installed')
```
ottonemo: Maybe precedence is important here?

benjamin-work: No, it just doesn't work: https://github.com/pytest-dev/pytest/issues/964

PR #38, opened 2017-08-09
Make it possible to continue training 

benjamin-work:
by using `cold_start=False` or `partial_fit`

### DIFF
--- a/inferno/tests/test_net.py
+++ b/inferno/tests/test_net.py
@@ -424,6 +424,51 @@ class TestNeuralNet:
         net.initialize()
         net.get_params()
 
+    def test_call_fit_twice_retrains(self, net_cls, module_cls, data):
+        X, y = data[0][:100], data[1][:100]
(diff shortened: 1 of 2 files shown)


inferno/tests/test_net.py
```diff
@@ -424,6 +424,51 @@ def test_get_params_with_uninit_callbacks(self, net_cls, module_cls):
...
+        net = net_cls(module_cls, cold_start=True).fit(X, y)
+        params_before = net.module_.parameters()
+
+        net.max_epochs = 0
+        net.fit(X, y)
+        params_after = net.module_.parameters()
+
+        assert len(net.history) == 0
+        for p0, p1 in zip(params_before, params_after):
+            assert (p0 != p1).data.any()
```
ottonemo: I'm not sure what's tested here. We don't iterate in `fit_loop` but we call fit. Do we expect it to not blow up? Maybe add a comment here.

benjamin-work: I added a comment, is it clear now?

PR #40, opened 2017-08-10
Feature: initialized modules

benjamin-work:
It is now possible to pass an initialized torch module to NeuralNet (e.g. an nn.Sequential). Iff a param is set on that module, it is re-initialized.

Addresses #39

### DIFF
--- a/inferno/tests/test_net.py
+++ b/inferno/tests/test_net.py
@@ -438,7 +438,7 @@ class TestNeuralNet:
         assert weight.shape[0] == 123
 
         stdout = capsys.readouterr()[0]
-        assert stdout.startswith("Re-initializing module!")
+        assert "Re-initializing module!" in stdout
 
     def test_with_initialized_module_non_default(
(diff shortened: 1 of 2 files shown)


inferno/tests/test_net.py
```diff
@@ -424,6 +424,76 @@ def test_get_params_with_uninit_callbacks(self, net_cls, module_cls):
...
+        module = module_cls(num_units=123)
+        net = net_cls(module, max_epochs=0, cold_start=False)
+        net.partial_fit(X, y)
+
+        for p0, p1 in zip(module.parameters(), net.module_.parameters()):
+            assert p0.data.shape == p1.data.shape
+            assert (p0 == p1).data.all()
+
+        stdout = capsys.readouterr()[0]
+        assert not stdout.startswith("Re-initializing module!")
```
ottonemo: I wonder whether we could implement this test better using property mocks on the initialised attribute for example.

PR #49, opened 2017-08-16
Simplify printlog and scoring

benjamin-work:
@ottonemo what do you think of the changes? If they are good, I will go ahead and replace old `Scoring` with new `WholesomeScoring` (including tests). Don't merge yet.

### DIFF
--- a/inferno/callbacks.py
+++ b/inferno/callbacks.py
@@ -490,3 +490,107 @@ class PrintLog(Callback):
         self.sink(tabulated.rsplit('\n', 1)[-1])
         if self.sink is print:
             sys.stdout.flush()
+
+
+class WholesomeScoring(Callback):
+    """Callback that performs generic scoring on predictions.
+
+    Parameters
+    ----------
+    name : str (default='myscore')
+      The name of the score. Determines the column name in the
+      history.
+
+    scoring : None, str, or callable (default=None)
+      If None, use the `score` method of the model. If str, it should
+      be a valid sklearn metric (e.g. "f1_score", "accuracy_score"). If
+      a callable, it should have the signature (model, X, y), and it
+      should return a scalar. This works analogously to the `scoring`
+      parameter in sklearn's `GridSearchCV` et al.
+
+    lower_is_better : bool (default=True)
+      TODO
+
+    on_train : bool (default=False)
+      Whether this should be called during train or validation.
+
+    target_extractor : callable (default=to_numpy)
+      This is called on y before it is passed to scoring.
+
+    pred_extractor : callable (default=to_numpy)
+      This is called on y_pred before it is passed to scoring.
+
+    """
+    _op_dict = {True: operator.lt, False: operator.gt}
+
+    def __init__(
+            self,
+            name='myscore',
+            scoring=None,
+            lower_is_better=True,
+            on_train=False,
+            target_extractor=to_numpy,
+            pred_extractor=to_numpy,
+    ):
+        self.name = name
+        self.scoring = scoring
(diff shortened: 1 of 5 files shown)


inferno/callbacks.py
```diff
@@ -402,55 +393,29 @@ class PrintLog(Callback):
...
-        if keys_optional is None:
-            self.keys_optional = []
-        elif isinstance(keys_optional, str):
-            self.keys_optional = [keys_optional]
-        else:
-            self.keys_optional = keys_optional
-
+        if isinstance(keys_ignored, str):
+            keys_ignored = [keys_ignored]
+        self.keys_ignored = keys_ignored
```
ottonemo: Maybe we should have something like 

    net.get_callback_by_name(name)

to modify this value later on?

inferno/callbacks.py
```diff
@@ -509,3 +490,107 @@ def on_epoch_end(self, net, *args, **kwargs):
...
+
+    def on_batch_end(self, net, X, y, train):
+        if train != self.on_train:
+            return
+
+        y = self.target_extractor(y)
+        if self.scoring is None:
+            score = net.score(X, y)
+        elif isinstance(self.scoring, str):  # TODO: make py2.7 compatible
+            # scoring is a string
```
ottonemo: This is bridge.

Maybe better: lookup and use sklearn metric for scorIng.

benjamin-work:
@ottonemo Please review again.

The main change is that before, we had different tasks (scoring, averaging, determining if best) distributed on different callbacks. Now all those tasks are performed by `Scoring`. However, this is less complex than the previous situation, because now each `Scoring` callback is responsible for exactly one score (train loss, accuracy, f1 etc.), where previously, each function was responsible for all scores. As a result, if I want to add one new score, I just have to add one more `Scoring` callback, instead of 3 (`Scoring`, `AverageLoss`, `BestLoss`).

To make the above work, `PrintLog` had to be changed to use black lists instead of white lists. Also, `PrintLog` is now by default the last callback to be called (so that all scores will have been written when it is called). Moreover, to get nicer results, `PrintLog` will always print the epoch first and the duration last.

inferno/tests/test_callbacks.py
```diff
@@ -56,267 +142,76 @@ def test_average_honors_weights(self, avg_loss, history):
...
+        is_best = history[:, 'mse_best']
+        assert is_best == [True, True, False]
+
     def test_other_score_and_name(self, scoring_cls, net):
+        """Test that we can change the scoring to accuracy score."""
         scoring = scoring_cls(
             name='acc',
             scoring='accuracy_score',
         )
         for x, y in zip(np.arange(5), reversed(np.arange(5))):
```
ottonemo: I think this warrants a comment what this x/y data is expected to give as a result. I know that the net is just the identity function but only because I read the whole test code. This is not clear from just looking at this test.

inferno/tests/test_callbacks.py
```diff
@@ -333,100 +229,164 @@ def score_func(estimator, X, y):
...
+
+    def test_is_best_ignored_when_none(self, scoring_cls, net, mock_data):
+        mse_scoring = scoring_cls(
+            name='mse',
+            scoring='mean_squared_error',
+            lower_is_better=None,
+        ).initialize()
+        history = get_history(mse_scoring)
+        with pytest.raises(KeyError):
+            # 'is_best' key should not be written
```
ottonemo: [...] not be written as lower_is_better is set to None

benjamin-work:
@ottonemo please review again.

PR #51, opened 2017-08-31
update to newer version of wheel file

soumith:
post1 had some crashes when loading in conjunction with ipython / tensorflow / zmq etc., post3 fixes these issues.


benjamin-work:
ty

PR #57, opened 2017-09-10
verbosity and example notebook

BenjaminBossan:
Verbosity is now a parameter on `NeuralNet`. It has not much use besides `PrintLog` and a single print statement in `NeuralNet`. However, it is really necessary for sane grid searches.

The example notebook just shows some basic usage. It should be extended in the future, and some content may be moved to an "extended usage" notebook.


BenjaminBossan:
@ottonemo Please review.

PR #58, opened 2017-09-10
pandas is not required in the setup.py and hence removed.

BenjaminBossan:
It should not have been in there, as usage with `pandas` is optional.

@ottonemo Please review.


ottonemo:
LGTM

Issue #56, opened 2017-09-10
[Bug] Binary classification on cuda fails


BenjaminBossan:
This is a similar issue as with regression and 1-dimensional target data, namely that `default_collate` unpacks the contents of the array (int64) and then the `.cuda()` call fails on int64.

It does not happen with 2-dimensional arrays, but for binary classifications, we can't use an n x 1 array, since that conflicts with `StratifiedKFold`.

ottonemo:
Fixed in #63.

PR #63, opened 2017-09-16
Fix bug that occurs with classification and cuda.

BenjaminBossan:
The problem was that `.cuda()` is called on a numpy scalar. To fix this, we have to wrap the scalar into a numpy array, but then `default_collate` will turn this into a n x 1 Tensor, which fails with `NLLLoss`. Consequently, `NeuralNetClassifier` intercepts n x 1 Tensors and flattens them.

@ottonemo pls review

### DIFF
--- a/inferno/net.py
+++ b/inferno/net.py
@@ -923,10 +923,22 @@ class NeuralNetClassifier(NeuralNet):
                              "`iterator_train` and `iterator_valid` "
                              "parameters respectively.")
 
+    def _prepare_target_for_loss(self, y):
+        # This is an ugly work-around (relating to #56), but
+        # currently, I see no solution that would result in a 1-dim
+        # LongTensor after passing through torch's DataLoader. If
+        # there is, we should use that instead.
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Addresses #56 
Probably also affects multi-class classification, not just binary.

inferno/net.py
```diff
@@ -917,10 +917,22 @@ def check_data(self, _, y):
                              "`iterator_train` and `iterator_valid` "
                              "parameters respectively.")
 
+    def _prepare_target_for_loss(self, y):
+        # This is an ugly work-around (relating to #56), but
+        # currently, I see no solution that would result in a 1-dim
+        # LongTensor after passing through torch's DataLoader. If
+        # there is, we should use that instead.
```
ottonemo: Add note that this goes away once pytorch has scalar types.

ottonemo:
Let's add this issue to the project board so we don't forget about it until  pytorch supports scalars or some other solution to this problem.

PR #64, opened 2017-09-29
Update dependencies to work with torch 0.2

benjamin-work:
Currently, dependencies don't work because we need numpy 1.13 for torch 0.2. That's why dependencies are updated here.

With new versions, a bug occurred in `__getstate__` and `__setstate__`, which is fixed.

Also, remove pytorch from environment.yml, because we can't know the user's cuda version. Update README to reflect this.


requirements-dev.txt
```diff
@@ -1,7 +1,8 @@
 jupyter
-matplotlib==2.0.0
+matplotlib>=2.0.2
 numpydoc
 openpyxl
+pandas
```
ottonemo: Pandas again? Where?

benjamin-work: Yes, I believe it makes sense here, since these are the requirements for developers, who should take care not to break the pandas functionality.

ottonemo: To prevent the case that pandas test cases are skipped during development. Gotcha.

Issue #70, opened 2017-10-06
Investigate whether a model trained on GPU can be loaded on CPU and vice versa, with pickle and state_dict. If that does not work, make it work.


benjamin-work:

benjamin-work:
It seems that at least with pickle, it doesn't work:

```
...
  File "/opt/conda/envs/blotto/lib/python3.6/site-packages/torch/cuda/__init__.py", line 266, in _lazy_new
    _lazy_init()
  File "/opt/conda/envs/blotto/lib/python3.6/site-packages/torch/cuda/__init__.py", line 84, in _lazy_init
    _check_driver()
  File "/opt/conda/envs/blotto/lib/python3.6/site-packages/torch/cuda/__init__.py", line 58, in _check_driver
    http://www.nvidia.com/Download/index.aspx""")
AssertionError: 
Found no NVIDIA driver on your system. Please check that you
have an NVIDIA GPU and installed a driver from
http://www.nvidia.com/Download/index.aspx

```

ottonemo:
This should be fixed in #70.

PR #71, opened 2017-10-06
Attempt at fixing #70

ottonemo:
Use `torch.load` so we can use the location mapping feature
which enables us to remap GPU storage to CPU storage.

### DIFF
--- a/inferno/exceptions.py
+++ b/inferno/exceptions.py
@@ -1,4 +1,4 @@
-"""Contains inferno-specific exceptions."""
+"""Contains inferno-specific exceptions and warnings."""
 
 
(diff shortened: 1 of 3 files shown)


benjamin-work:
This does not work currently because `optim_` has references to the model parameters, which are still on cuda. We need to treat `optim_` separately (maybe have a list of keys that need to be treated separately).

In addition, we should skip all tests that require cuda (`pytest.mark.skipf(...)`) when cuda is not available.

benjamin-work:
@ottonemo please have a look

ottonemo:
LGTM

ottonemo:
LGTM

Issue #77, opened 2017-10-11
*Examples* Add an example for using vanilla classifier and regressor


benjamin-work:

ottonemo:
We have two notebooks showing examples for this. See PR #78.

PR #78, opened 2017-10-11
Add more examples to notebook

benjamin-work:
Mainly adds a regression task to the Basic_Usage notebook. I moved the "custom callbacks" section to an Advanced_Usage notebook, where we should put all the stuff that involves custom classes etc.

During the task, I discovered a bug in `NeuralNetRegressor`, which is fixed now.

This addresses phantom issue #77

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,7 +7,7 @@ Example
 -------
 
 To see a more elaborate example, look `here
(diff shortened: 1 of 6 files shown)


benjamin-work:
* changed the custom callback
* add example on how to access callback parameters
* changed `optim` to `optimizer` for upcoming change

@ottonemo Please review again.

PR #80, opened 2017-10-11
Add checkpointing callback

### LINKED ISSUE #18: *Callbacks* A callback that saves the model periodically.
Requirements:

- the callback should not only save the weights but also the training parameters (i.e. pickling the learner and using `torch.save` to save the model data)
- option to disable 'best only' checking, we might want to do check-pointing at every epoch

Open questions:

- how much customization do we allow for selecting the 'best' run? Just make the key configurable (e.g. `key='valid_loss_best'`)?
- do we want to provide formatting options for the data path (e.g. `{epoch}` or `{unique_run_id}`)? Might be useful when doing grid search and runs would otherwise override checkpoints

ottonemo:
Fixes #18.

### DIFF
--- a/skorch/callbacks.py
+++ b/skorch/callbacks.py
@@ -14,6 +14,7 @@ from tabulate import tabulate
 from skorch.utils import Ansi
 from skorch.utils import to_numpy
 from skorch.utils import to_var
+from skorch.exceptions import SkorchException
 
 
 class Callback:
@@ -384,10 +385,11 @@ class Checkpoint(Callback):
     >>> net = MyNet(callbacks=[Checkpoint()])
     >>> net.fit(X, y)
 
-    Example using custom monitor:
+    Example using a custom monitor where only models are saved in
+    epochs where the validation *and* the train loss is best:
 
     >>> monitor = lambda net: all(net.history[-1, (
-    ...     'train_loss_best', 'valid_loss_best')]
+    ...     'train_loss_best', 'valid_loss_best')])
     >>> net = MyNet(callbacks=[Checkpoint(monitor=monitor)])
     >>> net.fit(X, y)
 
@@ -412,14 +414,14 @@ class Checkpoint(Callback):
       In case `monitor` is set to `None`, the callback will save
       the network at every epoch.
 
-      **Note:** If you supply a lambda expression as monitor,
-      checkpointing will not work anymore as these values cannot
-      be pickled. You have to use importable functions instead.
+      **Note:** If you supply a lambda expression as monitor, you cannot
+      pickle the wrapper anymore as lambdas cannot be pickled. You can
+      mitigate this problem by using importable functions instead.
     """
     def __init__(
-        self,
-        target='model.pt',
-        monitor='valid_loss_best',
+            self,
+            target='model.pt',
+            monitor='valid_loss_best',
     ):
         self.monitor = monitor
         self.target = target
@@ -430,7 +432,13 @@ class Checkpoint(Callback):
         elif callable(self.monitor):
             do_checkpoint = self.monitor(net)
         else:
-            do_checkpoint = net.history[-1, self.monitor]
+            try:
+                do_checkpoint = net.history[-1, self.monitor]
+            except KeyError as e:
+                raise SkorchException(
+                    "Monitor value '{}' cannot be found in history. "
+                    "Make sure you have validation data if you use "
+                    "validation scores for checkpointing.".format(e.args[0]))
 
         if do_checkpoint:
             target = self.target
@@ -440,4 +448,6 @@ class Checkpoint(Callback):
                     last_epoch=net.history[-1],
                     last_batch=net.history[-1, 'batches', -1],
                 )
+            if net.verbose > 0:
+                print("Checkpoint! Saving model to {}.".format(self.target)
             net.save_params(target)
--- a/skorch/callbacks.py
+++ b/skorch/callbacks.py
@@ -365,3 +365,79 @@ class PrintLog(Callback):
         self._sink(tabulated.rsplit('\n', 1)[-1], verbose)
         if self.sink is print:
             sys.stdout.flush()
+
+
+class Checkpoint(Callback):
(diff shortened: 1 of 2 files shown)


ottonemo:
@benjamin-work What do you think about the uncommented test case? What should happen here?

skorch/callbacks.py
```diff
@@ -365,3 +365,79 @@ def on_epoch_end(self, net, **kwargs):
...
+
+    Example:
+
+    >>> net = MyNet(callbacks=[Checkpoint()])
+    >>> net.fit(X, y)
+
+    Example using custom monitor:
+
+    >>> monitor = lambda net: all(net.history[-1, (
+    ...     'train_loss_best', 'valid_loss_best')]
```
benjamin-work: A closing parenthesis is missing at the end.

skorch/callbacks.py
```diff
@@ -365,3 +365,79 @@ def on_epoch_end(self, net, **kwargs):
...
+    You can also specify your own metric to monitor or supply a
+    callback that dynamically evaluates whether the model should
+    be saved in this epoch.
+
+    Example:
+
+    >>> net = MyNet(callbacks=[Checkpoint()])
+    >>> net.fit(X, y)
+
+    Example using custom monitor:
```
benjamin-work: ...that only saves the model when both train and valid loss improved.

skorch/callbacks.py
```diff
@@ -365,3 +365,79 @@ def on_epoch_end(self, net, **kwargs):
...
+      Value of the history to monitor or callback that determines whether
+      this epoch should to a checkpoint. The callback takes the network
+      instance as parameter.
+
+      In case `monitor` is set to `None`, the callback will save
+      the network at every epoch.
+
+      **Note:** If you supply a lambda expression as monitor,
+      checkpointing will not work anymore as these values cannot
+      be pickled. You have to use importable functions instead.
```
benjamin-work: I read this note to mean that if a lambda is used (as in the example above), checkpointing would fail. This is, however, not true, as `save_params` only saves the model parameters. Pickling would not work, but that is True no matter what argument is set to be a lambda.
If something else is meant, please clarify.

skorch/tests/test_callbacks.py
```diff
@@ -527,3 +527,75 @@ def test_print_skipped_if_not_verbose(self, capsys):
...
+        net = net_cls(callbacks=[
+            checkpoint_cls(monitor=None),
+        ])
+        net.fit(*data)
+
+        assert save_params_mock.call_count == len(net.history)
+
+    """
+    # TODO: should this fail with KeyError?
+    #       should the user be warned?
```
benjamin-work: I believe this should raise an error, since making no checkpoint could be very detrimental. The error should be more specific than `KeyError: 'valid_loss_best'`, though.

skorch/callbacks.py
```diff
@@ -365,3 +365,79 @@ def on_epoch_end(self, net, **kwargs):
...
+    Parameters
+    ----------
+    target : file-like object, str
+      File path to the file or file-like object.
+      See NeuralNet.save_params for details what this value may be.
+
+      If the value is a string you can also use format specifiers
+      to, for example, indicate the current epoch. Accessible format
+      values are `net`, `last_epoch` and `last_batch`.
+      Example to include last epoch number in file name:
```
benjamin-work: Is there precedence for this? I would not have expected this behavior (which is not to say it isn't useful).

skorch/tests/test_callbacks.py
```diff
@@ -527,3 +527,75 @@ def test_print_skipped_if_not_verbose(self, capsys):
...
+        net = net_cls(
+            callbacks=[
+                checkpoint_cls(),
+            ],
+            train_split=None
+        )
+        net.fit(*data)
+    """
+
+    def test_string_monitor_and_formatting(self, save_params_mock, net_cls, checkpoint_cls, data):
```
benjamin-work: Line too long

skorch/tests/test_callbacks.py
```diff
@@ -527,3 +527,75 @@ def test_print_skipped_if_not_verbose(self, capsys):
...
+            batch_size=10)
+
+    @pytest.fixture(scope='module')
+    def data(self):
+        # have 10 examples so we can do a nice CV split
+        X = np.zeros((10, 1), dtype='float32')
+        y = np.zeros((10, 1), dtype='float32')
+        return X, y
+
+    def test_none_monitor_saves_always(self, save_params_mock, net_cls, checkpoint_cls, data):
```
benjamin-work: Line too long

skorch/callbacks.py
```diff
@@ -365,3 +365,79 @@ def on_epoch_end(self, net, **kwargs):
...
+
+        if do_checkpoint:
+            target = self.target
+            if isinstance(self.target, str):
+                target = self.target.format(
+                    net=net,
+                    last_epoch=net.history[-1],
+                    last_batch=net.history[-1, 'batches', -1],
+                )
+            net.save_params(target)
```
benjamin-work: What do you think about printing a message if `net.verbose > 0`?

skorch/callbacks.py
ottonemo: I missed that we only save the state dict, that's fine then.

You *can* pickle attributes that hold a reference to a function though, as long as it is importable.

skorch/callbacks.py
ottonemo: Yes. if you have a better example I will replace this one. However, for the sake of showing how to use this callback this example suffices IMO.

benjamin-work: No, I just meant that the sentence should be extended to make it more transparent what happens.

skorch/callbacks.py
ottonemo: Yes. See https://keras.io/callbacks/#modelcheckpoint

ottonemo:
@benjamin-work please review again

benjamin-work:
:+1:

PR #86, opened 2017-10-13
Resolve #13

### LINKED ISSUE #13: Make initialization scheme consistent
Current state of different things (both = class and object supported, class = only class supported)

- callbacks (both)
- criterion (class)
- module (both)
- iterator_{train,test} (class)
- optimizer (class)

we should find a consistent scheme for this (either only initialized, always both, only class, ...)

ottonemo:
Moved `default_callbacks` to `get_default_callbacks()`


benjamin-work:
:+1:

PR #89, opened 2017-10-16
Feature/forward iter

benjamin-work:


skorch/net.py
```diff
@@ -612,8 +629,10 @@ def predict_proba(self, X):
         y_proba : numpy ndarray
 
         """
-        y_proba = self.forward(X, training=False)
-        y_proba = to_numpy(y_proba)
+        y_probas = []
+        for yp in self.forward_iter(X, training=False):
+            y_probas.append(to_numpy(yp))
```
ottonemo: Why not a list comprehension?

skorch/tests/test_net.py
```diff
@@ -705,13 +705,13 @@ def test_target_1d_raises(self, net, data):
...
+    def test_predict_predict_proba(self, net_fit, data):
         X = data[0]
-        y_pred = net.predict(X)
+        y_pred = net_fit.predict(X)
 
         # predictions should not be all zeros
         assert not np.allclose(y_pred, 0)
 
-        y_proba = net.predict_proba(X)
+        y_proba = net_fit.predict_proba(X)
```
ottonemo: Why this change and why didn't the previous version fail? This test doesn't do much if the unfitted net behaves the same?

benjamin-work: The `net` fixture is uninitialized, which is why `predict` shouldn't work. Through side effects of other tests, this still works, but calling this test in isolation fails.

skorch/net.py
benjamin-work: `y_probas = [to_numpy(yp) for yp in self.forward_iter(X, training=False)]` -> 80 chars with indents.

skorch/tests/test_net.py
ottonemo: And these side-effects are not a problem?

PR #101, opened 2017-10-25
Disable CUDA on load/unpickle if not available

### LINKED ISSUE #93: Disable `use_cuda` after loading a model on a non-CUDA machine
Currently we warn that CUDA is not supported but the model still has `use_cuda=True`.

ottonemo:
This resolves #93.


benjamin-work:
:+1:

PR #108, opened 2017-11-03
Introduce ProgressBar callback

ottonemo:
Useful when training time between epochs is quite long.


ottonemo:
@dnouri Do you think this is the best way to handle this? I also experimented with having only one progress bar so that they don't stack when in jupyter notebook mode, however it is quite hard to re-use a progress bar instance and I ran into all sorts of problems.

benjamin-work:
@ottonemo Could you add an example (notebook or not)?

PR #110, opened 2017-11-10
Remove optimizer parameter from train_step

ottonemo:
Benefits:
- Consistency with valid_step
- Consistency when using more than one optimizer


benjamin-work:
Broken example will be fixed separately.

PR #111, opened 2017-11-15
Revamp the word language model example

ottonemo:
Notable changes:

- Boiled down predict.py to a point where, if there are no
  further additions I would argue that it could be removed.
- Removed optimizer parameter in train_step
- Removed a lot of FIXME/TODO which solved themselves over time
- Removed statefulness of predict/validation
- Add manual seeding


ottonemo:
@benjamin-work the final scoring in each gridsearch step is quite memory exhaustive (600k+ samples, >32GiB RAM usage). Do you see a way to alleviate this or do we just warn the user?

ottonemo:
The memory consumption issue is addressed in #116.

Issue #112, opened 2017-11-15
EpochScoring is flaky when y_test is None


ottonemo:
Given a train split function like

    def my_train_split(X, y):
        return X[:10], X[10:], None, None

which results in `y_train` and `y_valid` being `None`, the `EpochScoring` will raise an exception:

```
 [...]
  File "/home/user/Code/skorch/skorch/net.py", line 557, in fit_loop
    self.notify('on_epoch_end', **on_epoch_kwargs)
  File "/home/user/Code/skorch/skorch/net.py", line 291, in notify
    getattr(cb, method_name)(self, **cb_kwargs)
  File "/home/user/Code/skorch/skorch/callbacks.py", line 294, in on_epoch_end
    y_test = self.target_extractor(y_test)
  File "/home/user/Code/skorch/skorch/utils.py", line 94, in to_numpy
    if X.is_cuda:
AttributeError: 'NoneType' object has no attribute 'is_cuda'
```

The issue here is that the scoring callback checks for `X_test` and not for `y_test`, but relies on both. Simply checking for both is not a viable workaround since this breaks the computation of the train and validation loss (which only relies on `X`).

We need to support unsupervised models that do not have a `y_test` but still want to compute metrics on `X` or the history of the net (e.g. train and validation loss).

ottonemo:
Is it a possible solution to not call the target extractor on `y_test` if it is `None`?

PR #114, opened 2017-11-16
Issues/112

ottonemo:
See #112


ottonemo:
Invalid commits in PR, also we'll probably not implement it this way. Abandoning PR.

PR #120, opened 2017-11-27
MNIST example for skorch

juripaern:
Main idea: use sci-kit learn for data processing and skorch/pytorch for neural-networks.

Contains:
- A simple dense NN
- A CNN

### DIFF
--- a/notebooks/MNIST.ipynb
+++ b/notebooks/MNIST.ipynb
@@ -91,7 +91,7 @@
    "source": [
     "## Preprocessing Data\n",
     "Each image of the MNIST dataset is encoded in a 784 dimensional vector, representing a 28 x 28 pixel image. Each pixel has a value between 0 and 255, corresponding to the grey-value of a pixel.<br />\n",
-    "If the above ```featch_mldata``` is used to load MNIST, ```data``` and ```target``` are encoded as ```uint8```, which cannot be processed by Skorch. Skorch accepts only ```float32``` or ```int64``` as types. As a consequence ```data``` is transformed to ```float32``` and scaled to [0, 1]. ```target``` is transformed to ```int64```. "
+    "The above ```featch_mldata``` method to load MNIST returns ```data``` and ```target``` as ```uint8```, which cannot be processed by PyTorch. PyTorch accepts only ```float32``` or ```int64``` as types. As a consequence ```data``` is transformed to ```float32``` and ```target``` to ```int64```. "
    ]
   },
   {
@@ -102,6 +102,13 @@
    "source": [
     "X = mnist.data.astype('float32')\n",
     "y = mnist.target.astype('int64')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we will use ReLU as activation in combination with softmax over the output layer, we need to scale `X` down. An often use range is [0, 1]."
    ]
   },
   {
@@ -121,7 +128,7 @@
     {
      "data": {
       "text/plain": [
-       "(0.0, 0.13092543, 0.30844888, 1.0)"
+       "(0.0, 1.0)"
       ]
      },
      "execution_count": 7,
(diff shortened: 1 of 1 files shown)


ottonemo:
> If the above featch_mldata is used to load MNIST, data and target are encoded as uint8, which cannot be processed by Skorch. Skorch accepts only float32 or int64 as type

If anything, PyTorch cannot handle these data types. Skorch is pretty agnostic about input data types and even adds some more to the supported list (e.g. dictionaries). However, in this case even `torch.Tensor(mnist.target)` works without problems so I don't know what your point is here?

> `>>> X.min(), X.mean(), X.std(), X.max()`
> (0.0, 0.13092543, 0.30844888, 1.0)

This would be a good point on stating whether or not you are normalizing your inputs.

> `np.mean(cnn_pred == y_test)`
> 0.99085714285714288

Can't you train a bit longer for more accuracy? :)

juripaern:
> If anything, PyTorch cannot handle these data types. Skorch is pretty agnostic about input data types and even adds some more to the supported list (e.g. dictionaries). 

Good to know, changed text accordingly.
I tried `float32`, `float64`, `uint8`, `int32` and `int64`. For me, only `float32` and `int64` works.  

> However, in this case even torch.Tensor(mnist.target) works without problems so I don't know what your point is here?

My idea was to use PyTorch only for modeling a network. Everything else should be done with NumPy/SciKit learn. However, if you think it makes more sense to use `torch.Tensor` I'm happy to change it.

> This would be a good point on stating whether or not you are normalizing your inputs.

I added text to explain, why a scaling is necessary for the notebook to work and a short note that data is not normalized.

>  You should explain what you are doing here (1 channel, image height x width, etc.).

Good point. Added a little description.

> Can't you train a bit longer for more accuracy? :)

To be honest: I can't :-(
Starting from epoch 17 (or so) loss will be `nan`.
Haven't had the time to look into it.

ottonemo:
OK LGTM for the sake of tihs example. Thanks!

asmith26:
~~How was 1600 calculated?~~ It took me a while to calculate this, but I think I've worked it out:

According to [CS231n course notes](http://cs231n.github.io/convolutional-networks/#conv), we have:

- Dimensions resulting from convolution: `height = width = (width - kernel_size + 2*padding) / stride + 1`, `num_channels = num_filters`
- Dimensions resulting from pooling: `height = width = (width - kernel_size) / stride + 1`

Thus for:
- `input_shape = (num_channels, height, width) = (1, 28, 28)` 
- Layers definitions:
  - `x = F.relu(F.max_pool2d(self.conv1(x), kernel_size=2, stride=kernel_size, padding=0))` with `self.conv1 = nn.Conv2d(1, num_filters=32, kernel_size=3, stride=1, padding=0)` (also using [`Conv2d` defaults](http://pytorch.org/docs/master/nn.html#torch.nn.Conv2d) and [`max_pool2d` defaults](http://pytorch.org/docs/master/nn.html#torch.nn.functional.max_pool2d))
  - Same again except `self.conv2 = nn.Conv2d(32, num_filters=64)`

We get:
- 1st convolution: `num_channels = 32` and `height = width = (28 - 3 + 2*0) / 1 + 1 = 26`
- 1st pooling: `num_channels = 32` and `height = width = (26 - 2) / 2 + 1 = 13` (note stride=kernel_size=2 accoring to [these docs](http://pytorch.org/docs/master/nn.html#torch.nn.MaxPool2d))
- 2nd convolution: `num_channels = 64` and `height = width = (13 - 3 + 2*0) / 1 + 1 = 11`
- 2nd pooling: `num_channels = 64` and `height = width = (11 - 2) / 2 + 1 = 5.5` and rereading the torch doc, we take the floor of this giving 5.

Hence `number of pixels = num_channels * height * width = 64 * 5 * 5 = 1600`

Thanks for this great example, and for the excellent library!

asmith26:
In fact, it also looks like you can compute this automatically (may want to do this in this/another example?)

See e.g.  [this suggestion](https://discuss.pytorch.org/t/inferring-shape-via-flatten-operator/138/4) or [this one](https://discuss.pytorch.org/t/inferring-shape-via-flatten-operator/138/6). Some also mentions:

>Note that you can add a global operation (like global max/average pooling) just before your view layer, so that you know precisely the number of inputs that the linear layer will receive (as you can see in the [resnet model definition](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L149), where the kernel size for the pooling can be computed on the fly using the functional interface).

PR #121, opened 2017-11-29
Fix issue 112: EpochScoring flaky when y_test is None.

benjamin-work:
The current solution is to not call `target_extractor`. That way, None is passed through, which will then lead sklearn to call the scoring function without y. Therefore, if the scoring function supports that, the user is fine.

The reason why `BatchScoring` does work here is because we currently replace non-existing y with a dummy tensor during batching. Of course, this is not ideal but not in the scope of this fix. It _may_ also solve itself with updates in pytorch.


benjamin-work:
Addresses #112

Issue #45, opened 2017-08-15
Consolidate names


ottonemo:
- [x] train vs training_behavior (e3d215e9275cf3d057033f02448ac2f9da86947e)
- [x] optim vs. optimizer (286b370c7a6e8d7f118c6f016477c6acb7bf8f02)
- [x] valid vs test (aeafd7e29ae6e54d573323c3b84422f7f1c25103)
- [x] xi vs Xb vs Xi (155df7bc5770d57d5057656f9a31bbf5eb045675)
- [x] params vs kwargs vs *_kwargs
- [x] <strike>change net.py to base.py</strike>
- [x] stay with `NeuralNet`?

ottonemo:
Possibly rename `Dataset` to something less akin to `pytorch.utils.data.Dataset`?

ottonemo:
re valid vs. test: pro train/test: sklearn has `train_test_split` as a function. con: is never used for test, always for validation. I think we should rename 'test' occurrences to 'valid'.

benjamin-work:
:+1: 
Also, "train" and "valid" have the same number of characters, so that code will align more often.

benjamin-work:
Regarding `Dataset`: I was thinking about calling it `DatasetXy` instead, to signify that it takes and returns pairs of X and y. I could foresee a `DatasetX` for use cases where there is no y.

ottonemo:
What would a `DatasetX` be used for?

ottonemo:
We keep `Dataset` as is for now. Conflicts are unlikely to occur and easily fixed by the user.

PR #123, opened 2017-11-29
Feature/train vs training

ottonemo:
As discussed in #45 we wanted to consolidate names, such as training/train. This PR does that for `get_loss` and callbacks.


ottonemo:
Done, please review again.

PR #128, opened 2017-12-08
Add helper SliceDict to simplify work with sklearn.

benjamin-work:
If this is approved, I would add an example somewhere first before merging.

### DIFF
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -1,0 +1,74 @@
+"""Helper functions and classes for users.
+
+They should not be used in skorch directly.
+
+"""
+
+
+class SliceDict(dict):
+    """Wrapper for Python dict that makes it sliceable across values.
+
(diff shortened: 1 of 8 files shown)


skorch/helper.py
```diff
@@ -0,0 +1,74 @@
...
+    -------
+    >>> X = {'key0': val0, 'key1': val1}
+    >>> search = GridSearchCV(net, params, ...)
+    >>> search.fit(X, y)  # raises error
+    >>> Xs = SliceDict(key0=val0, key1=val1)  # or Xs = SliceDict(**X)
+    >>> search.fit(Xs, y)  # works
+
+    """
+    def __init__(self, **kwargs):
+        shapes = [value.shape[0] for value in kwargs.values()]
```
ottonemo: Please rename this to `lengths` instead of `shapes`. It is a bit misleading, especially in the error message.

PR #133, opened 2017-12-15
Add a notebook that shows how best to use RNNs with skorch.

benjamin-work:


benjamin-work:
@ottonemo ready for review

ottonemo:
> First make sure to install an additional package, dstoolbox

This sounds like dstoolbox is mandatory for the use with skorch.

    out = F.softmax(self.output(drop))

Dimension parameter is missing.

benjamin-work:
@ottonemo Points were addressed, pls review again.

PR #143, opened 2018-01-31
Resolve #142

### LINKED ISSUE #142: train_batch_size wrong in history with multi X inputs
Hey, thanks for your lib, skorch is cool !

https://github.com/dnouri/skorch/blob/master/skorch/net.py
line 566
`                self.history.record_batch('train_batch_size', len(Xi))
`
len(Xi) does not return correct batch size if Xi is a dict
`
Xi = { 'x': <FloatTensor> }
`

ottonemo:
Use `get_len` instead of `len` to correctly handle dictionaries as input.

### DIFF
--- a/skorch/callbacks.py
+++ b/skorch/callbacks.py
@@ -14,6 +14,7 @@ from tabulate import tabulate
 import tqdm
 from torch.nn.utils import clip_grad_norm
 
+from skorch.dataset import get_len
 from skorch.utils import Ansi
 from skorch.utils import to_numpy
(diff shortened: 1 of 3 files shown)


ottonemo:
I think 

https://github.com/dnouri/skorch/blob/master/skorch/callbacks.py#L605

needs fixing as well right?

benjamin-work:
Looks like it. Would you do it within this PR?

The rest LGTM

ottonemo:
Done.

Issue #141, opened 2018-01-19
Make improvements to forward and forward_iter


benjamin-work:
IMO, [`forward_iter`](https://github.com/dnouri/skorch/blob/9382a83f30d9ba0b1dae5bcc3896feee5e2a0dd5/skorch/net.py#L667) could do with some improvements:

* The docstring "Yield forward results from batches derived from data." is unclear -- it should explain in what conditions to use it.
* The method sets `self.module_.train(training)` but then calls `evaluation_step`, which does the same.

Furthermore, I would adjust [forward](https://github.com/dnouri/skorch/blob/9382a83f30d9ba0b1dae5bcc3896feee5e2a0dd5/skorch/net.py#L728) this way:

```
            y_infer = list(self.forward_iter(X, training=training))
            -->
            y_infer = [yp for yp, *_ in self.forward_iter(X, training=training)]
```

The result of this would be that if `forward_iter` returns more than 1 value (say, like the output of a torch RNN layer), it discards all but the first output. Currently, if more than 1 value is returned, `forward` (and thus `predict` and `predict_proba`) just break, which is discouraging. The change could be a little bit surprising for the user but would make this common use case a lot more comfortable.

ottonemo:
A problem I see is that I would not expect forward to modify the output of my model and it would certainly be helpful if I can rely on the fact that a call to forward on a skorch net will yield the same result as a direct call to my model.

benjamin-work:
I see the problem with the name `forward`. From your proposals, the 2nd makes more sense to me. If we change [this line](https://github.com/dnouri/skorch/blob/9382a83f30d9ba0b1dae5bcc3896feee5e2a0dd5/skorch/net.py#L785) in `predict_proba` from:

`for yp in self.forward_iter(X, training=False):`

to:

`for yp, *_ in self.forward_iter(X, training=False):`

(and the same for `predict`), would that not be enough? I don't see why that would require more memory and neither do I think that we would need a custom collate function. We would just specify that the contract is:

* to use `predict(_proba)`, the first output from `module_.forward` can be cast to numpy and concatenated via `np.concatenate`
* to use `forward`, all outputs must be able to be concatenated by `torch.cat`. (And maybe we need the option to pull batches from GPU to CPU if memory is a concern.)

benoua:
Hi, I had the same issue for a multilabel task. When I compute a recall target at each epoch, the whole predict_proba was not fitting in GPU and I had to do the stacking of results on CPU. 

[...]

PR #144, opened 2018-02-02
Clean up of inference methods and allow multioutput

benjamin-work:
Addresses #141 and then some.

The following things were changed:

* Remove redundant `training` call in `forward_iter`
* There was no explicit test for `NeuralNet.foward` -- added it
* *breaking change*: `NeuralNet.predict` no longer performs argmax -- this makes only sense for a classifier, which `NeuralNet` isn't; as a consequence, `NeuralNetClassifier.predict` is now charged with taking the argmax.
* `NeuralNet.forward` can now deal with multiple outputs from `module_.forward`: If they are returned as a tuple, they are separated and `torch.cat` is called on each of them individually.
* `predict` and `predict_proba`  can now deal with multiple outputs from `module_.forward`: They discard all but the first output if outputs are tuples.
* Improve and adjust docstrings, documentation

Not changed:

* Allow to pull batches to CPU when calling `NeuralNet.foward`. I would like to think a bit more about how to handle the parameters before taking on this issue (e.g. `store_on='cpu'` vs `store_on_cpu=True`).

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -723,14 +723,10 @@ class NeuralNet(object):
           The result from the forward step.
 
         """
-        y_infer = []
-        is_multioutput = None
-        for batch in self.forward_iter(X, training=training):
-            if is_multioutput is None:
-                is_multiouput = isinstance(batch, tuple)
-            y_infer.append(batch)
+        y_infer = list(self.forward_iter(X, training=training))
 
-        if is_multiouput:
+        is_multioutput = len(y_infer) > 0 and isinstance(y_infer[0], tuple)
+        if is_multioutput:
             return tuple(map(torch.cat, zip(*y_infer)))
         return torch.cat(y_infer)
 
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -692,8 +692,13 @@ class NeuralNet(object):
             yield yp
 
     def forward(self, X, training=False):
-        """Perform a forward step on the module with batches derived
-        from data.
+        """Gather and concatenate the output from forward call with
+        input data.
+
+        The outputs from ``self.module_.forward`` are gathered and
+        then concatenated using ``torch.cat``. If multiple outputs are
+        returned y ``self.module_.forward``, each one of them must be
+        able to be concatenated this way.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


skorch/net.py
```diff
@@ -720,8 +723,16 @@ def forward(self, X, training=False):
...
+        y_infer = []
+        is_multioutput = None
+        for batch in self.forward_iter(X, training=training):
+            if is_multioutput is None:
+                is_multiouput = isinstance(batch, tuple)
+            y_infer.append(batch)
+
+        if is_multiouput:
+            return tuple(map(torch.cat, zip(*y_infer)))
+        return torch.cat(y_infer)
```
ottonemo: Can't we simplify this to 
```
is_multiout = len(y_infer) > 0 and isinstance(y_infer[0], tuple)
```

skorch/tests/test_net.py
```diff
@@ -905,6 +917,54 @@ def test_set_params_grad_clip_and_norm_deprecated(
...
+        assert set(y_pred) == {0, 1}
+
+    def test_multiouput_predict_proba(self, multiouput_net, data):
+        X = data[0]
+        n = len(X)
+
+        # does not raise
+        y_proba = multiouput_net.predict_proba(X)
+
+        assert y_proba.shape == (n, 2)
```
ottonemo: Needs explanation what is expected here (as above)

skorch/tests/test_net.py
```diff
@@ -144,7 +145,18 @@ def test_net_learns(self, net_fit, data):
...
         assert accuracy_score(y, y_pred) > 0.7
 
-    def test_predict_proba(self, net_fit, data):
+    def test_forward(self, net_fit, data):
+        X = data[0]
+        n = len(X)
+        y_forward = net_fit.forward(X)
+
+        assert is_torch_data_type(y_forward)
+        assert y_forward.shape == (n, 2)
```
ottonemo: A comment what you expect would be nice, e.g.

    # Expecting (number of samples, number of output units)

skorch/tests/test_net.py
```diff
@@ -905,6 +917,54 @@ def test_set_params_grad_clip_and_norm_deprecated(
...
+        n = len(X)
+        y_infer = multiouput_net.forward(X)
+
+        assert isinstance(y_infer, tuple)
+        assert len(y_infer) == 3
+        for arr in y_infer:
+            assert is_torch_data_type(arr)
+
+        # 1st output is original, 2nd is only col 0, 3rd is only evey other row
+        assert y_infer[0].shape == (n, 2)
```
ottonemo: Needs explanation what is expected here (as above)

skorch/tests/test_net.py
```diff
@@ -905,6 +917,54 @@ def test_set_params_grad_clip_and_norm_deprecated(
...
+        assert y_infer[2].shape == (n // 2, 2)
+
+    def test_multioutput_predict(self, multiouput_net, data):
+        X = data[0]
+        n = len(X)
+
+        # does not raise
+        y_pred = multiouput_net.predict(X)
+
+        assert y_pred.shape == (n,)
```
ottonemo: We expect only `(n,)` since we are doing a predict, not because we are indexing `X` right? A short, clarifying comment would be nice.

benjamin-work:
@ottonemo pls review again

Issue #146, opened 2018-02-06
make it possible to fit with a Dataset instance


benjamin-work:
The idea is to allow something like this:

```
dataset = MyDataset(...)
net = NeuralNet(...)
net.fit(dataset)
```

This should allow to do things like loading the training data on the fly (when they don't fit memory, for instance).

Acceptance criteria:
* Make something like the [cifar tutorial](https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/cifar10_tutorial.py) work

Open questions:
* How to differentiate between train and valid data?
* Do we need a `dataset_train` and a `dataset_valid`?
* What to do with the `net.dataset` parameter?

ottonemo:
> How to differentiate between train and valid data?
> Do we need a dataset_train and a dataset_valid?

There are many scenarios where a train and a test dataset is provided but no validation. So, initially I thought we need to be able to compute a validation split from the train dataset, but for more complex splits (even a stratified split, really) this is impossible since we have no guarantee that we can inspect all targets from the dataset. Therefore, we cannot provide a functionality that computes a CV split from a given train dataset, we can only offer functionality to specify a separate validation dataset (this also means we raise an error if `X` is a dataset and the `train_split` is `CV`).

The current solution for this would look something like this:
```python
from skorch.helper import dataset_validation_split

net = Net(train_split=dataset_validation_split(validation_dataset))
net.fit(train_dataset)
```
with
```python
def dataset_validation_split(dataset_valid):
    def split(dataset_train, _):
        return dataset_train, dataset_valid, None, None
    return split
```

After that we can simply check in `fit_loop` whether `X_train` and `X_valid` are datasets and set `dataset_train` and `dataset_valid` accordingly.

> What to do with the `net.dataset` parameter?

Remove. Not needed since it can be substituted with a partial iterator anyway.

ottonemo:
One thing we have to care about is that datasets encapsulate both X and y which makes them hard to use in callbacks that use the data supplied in each epoch (this data would be the dataset). Callbacks such as the epoch scoring callback would cease to function. This is OK since the user can always specify a custom scoring function that deals with datasets or use batch scoring but we should at least throw a meaningful exception in epoch scoring.

Issue #54, opened 2017-09-04
Callbacks: Generalize `yield_callbacks` into processing and output callbacks


githubnemo:
Currently, `net._yield_callbacks` discerns between `PrintLog` and other callbacks with the effect that `PrintLog` is added last to the list of callbacks so it has access to all processed values added by other callbacks. Maybe we should generalize this by classifying callbacks into two groups: processing callbacks and output callbacks.

Output callbacks (identified by inheriting from an abstract subclass of `Callback`) are by default appended to the end of the callback list. This would pave the way for other output callbacks besides `PrintLog` such as TensorBoard logging callbacks.

ottonemo:
Maybe `PostponedCallback`? `PostAggregation`?

PR #149, opened 2018-02-09
Scores that don't require a forward prop each

dnouri:
- on_batch_end now also gets 'y_pred' and 'loss'
- Add EpochMetrics and BatchMetrics callbacks

  These make use of the 'y_pred' parameter that NeuralNet now passes
  into callbacks.  Using 'y_pred', it's no longer necessary to do a
  separate forward run of the network just to produce scores, when a
  forward run has already been performed to calculate the loss.

  Refactoring existing code to use this and deprecate BatchScoring and
  EpochScoring is a TODO.  Plus possibly more improvements.


ottonemo:
We should refrain from adding things to callback parameters that are available through the net's history, like loss for example.

dnouri:
> We should refrain from adding things to callback parameters that are available through the net's history, like loss for example.

Meh.  How about we remove the special case `self.history.record_batch` calls inside the training loop here and move them out into a callback.

ottonemo:
> Meh. How about we remove the special case self.history.record_batch calls inside the training loop here and move them out into a callback.

So every scoring callback adds the value to the history? That would only work if we have some sort of scoring callback bucket that's prefilled with a loss metric and can be extended by the user. Otherwise this would only be a more complicated change.

dnouri:
> So every scoring callback adds the value to the history? That would only work if we have some sort of scoring callback bucket that's prefilled with a loss metric and can be extended by the user. Otherwise this would only be a more complicated change.

Isn't it the case already that the sorting of callbacks is a bit delicate?  I remember someone saying that the `PrintLog` has to come after the other callbacks etc.  I'd suggest we look for a general solution for this kind of dependency problem, e.g. a simple `priority` (int) attribute that callbacks are sorted by before firing.  This way we could ensure that what we want to come first, will come first.  User defined callbacks could default to coming before the `PrintLog` and after whatever else we want to run before.

Also, now that `loss` and `Xi` (and thus `len(Xi)`) are available as parameters in callbacks, it would seem unnecessary for new callbacks to rely on getting those values from the history, yes?

ottonemo:
> Isn't it the case already that the sorting of callbacks is a bit delicate? 

It is and we probably need to address that sooner or later (see #54). I think we don't have to decide on where to move this bit of code here and now. Instead, we should focus on how this PR came to be, what the benefits are and what is still to do.

There are two angles for why we might consider a different scoring approach:
1. support datasets as inputs which break the assumption that we can always provide `X` and `y` with `on_epoch_begin`, this breaks `EpochScoring` but `BatchScoring` is fine
2. re-use of predictions to improve performance when using multiple scores

(1) would benefit from any solution that can switch to a different scoring level (epoch vs. batch) depending on the input data type OR caches all inputs for the whole epoch (probably not wise). This is not directly associated with this PR but should be considered when rewriting the scoring callback. 

(2) is the main reason for this PR and has several implications. First we cannot use sklearn scorers for this as they require the estimator and call `predict`/`predict_proba` depending on the chosen scoring. This makes it hard to introduce caching and any solution (proxy estimator, overwriting `predict`, ...) is probably to messy. So we need to switch to metrics directly (which only depend on `y_true` and `y_pred`). The next problem is that the predictions need to be stored without redundancy (`y_pred` might be quite large) where we have the option to store it in the history object or, as suggested with your change, in a central callback that handles the individual metrics themselves. I see one problem with this approach, namely that it is hard to have default metrics while still being able to add new callbacks, e.g.:
```
class Net:
    default_callbacks = Metrics(['accuracy'])
Net(callbacks=Metrics([....])
```
In this case we would have to override the metrics completely which might be a bit of a pain if there are more complicated metrics supplied by default and the user has to re-type some code from the defaults. While this is probably not a problem it might not be a concern but we should at least discuss it first.

dnouri:
Thanks for your comments.

> There are two angles for why we might consider a different scoring approach:
> 1. support datasets as inputs which break the assumption that we can always provide `X` and `y` with `on_epoch_begin`, this breaks `EpochScoring` but `BatchScoring` is fine
> 2. re-use of predictions to improve performance when using multiple scores

For (2), note that you already get a speed benefit from this patch even when calculating a single score!  Since the first scorer (such as the default `valid_acc`) already incurs an extra forward pass.  And it gets worse as you add scorers.

> (1) would benefit from any solution that can switch to a different scoring level (epoch vs. batch) depending on the input data type OR caches all inputs for the whole epoch (probably not wise). This is not directly associated with this PR but should be considered when rewriting the scoring callback.

I'd prefer having the user control which scoring function to use and refrain from automatic switching, but it's not a strong opinion.

So `BatchMetrics` doesn't need to collect the dataset before it does the scoring, just like `BatchScoring`.  Neither `BatchMetrics` nor `EpochMetrics` depend on the full `X` and `y` at `on_epoch_begin` time, so there's a path to supporting datasets where we don't have the complete `X` and `y` in memory.  (Mentioning this only for clarity.)

If you think keeping all targets and predictions in memory (like `EpochMetrics` does it) is problematic, consider that today we're requiring users of `EpochScoring` to load all training data into memory, so it's not really worse than the status quo; it's keeping all predictions and targets in memory vs. keeping all data and targets in memory.  Of course targets can be as big as inputs, but often the result will be a _smaller_ memory footprint than what we have today.

`BatchMetrics` should be the default for obvious reasons.  User can override with `EpochMetrics` if they care about it, such as when calculating F1.  No danger of blowing people's memory up if they don't change the defaults.

> (2) [...] I see one problem with [your approach of using a single callback for all metrics], namely that it is hard to have default metrics while still being able to add new callbacks, e.g.:
> ```
> class Net:
>     default_callbacks = Metrics(['accuracy'])
> Net(callbacks=Metrics([....])
> ```
> In this case we would have to override the metrics completely which might be a bit of a pain if there are more complicated metrics supplied by default and the user has to re-type some code from the defaults. While this is probably not a problem it might not be a concern but we should at least discuss it first.

Not concerned.  ;-)

Although with this patch it becomes viable to use multiple scores/metrics per default, I don't see us defining lots of default metrics.  (Though I would suggest calculating train _and_ valid accuracy for classification.)

ottonemo:
> I'd prefer having the user control which scoring function to use and refrain from automatic switching, but it's not a strong opinion.

It should definitely be user-controlled but the case epoch metrics + dataset should raise an error.

> So BatchMetrics doesn't need to collect the dataset before it does the scoring, just like BatchScoring. Neither BatchMetrics nor EpochMetrics depend on the full X and y at on_epoch_begin time, so there's a path to supporting datasets where we don't have the complete X and y in memory. (Mentioning this only for clarity.)

Right, this also means that we should probably change the signature of `on_epoch_begin` and `on_epoch_end` to accept a dataset (internal or provided) instead of `X` and `y`. This is more related to #146 but the scoring callback currently depends on `X`/`y` at `on_epoch_end`.

> `BatchMetrics` should be the default for obvious reasons. User can override with `EpochMetrics` if they care about it, such as when calculating F1. No danger of blowing people's memory up if they don't change the defaults.

Not sure about that. If we use batch-level by default then it is very easy to screw up F1 scores. I think we should make the safest setting (in terms of correctness) the default which would be epoch-level scoring.

> Although with this patch it becomes viable to use multiple scores/metrics per default, I don't see us defining lots of default metrics. (Though I would suggest calculating train and valid accuracy for classification by default.)

I'm fine with a solution that requires re-setting the accuracy metrics if the user wants to specify additional metrics.

Another point is that we the `Metrics` callback should support `set_params`, i.e. the user should be able to change aspects of the individual metrics contained in the callback , for example, `callback__metrics__accuracy__lower_is_better=False`.

ottonemo:
> I'm fine with a solution that requires re-setting the accuracy metrics if the user wants to specify additional metrics.

Actually I'm not fine with this unless we find a good way to integrate train and valid loss scoring because those need to be there by default and I would get annoyed rather quickly if I had to specify them all over again when adding new metrics.

ottonemo:
OK I think the default metric problem is not an actual problem. Every network instance will have default `Metrics` for train and valid loss and, in case of classification, could potentially have an accuracy metric. All these metrics are batch-level metrics which don't store any additional information except the metric score. Therefore defining a new `Metrics` instance at network initialization and having two `Metrics` callbacks in the list of active callbacks is most likely not a problem.

ottonemo:
Closing in favor of #171.

PR #154, opened 2018-02-27
Move callbacks into own submodule

ottonemo:
We have too many callbacks in one file. We should move them into a submodule.


benjamin-work:
Just for discussion: We could also manage the imports so that we just import * in `__init__.py` and move the burden to specify what to import to the respective modules by requiring them to define `__all__`. See e.g. here: https://github.com/Lasagne/Lasagne/tree/master/lasagne/layers

Not a strong opinion, though.

ottonemo:
I like that. Changed to defining `__all__`.

benjamin-work:
@ottonemo After resolving the conflict, go ahead and merge.

PR #155, opened 2018-02-27
Base Dockerfile and specific one for example

ottonemo:
We should think about setting up an automatic docker cloud build as well.


benjamin-work:
👍

PR #156, opened 2018-02-28
on_grad_computed gets more generic named_parameters

ottonemo:
This makes it far easier to do simple filtering on specific
parameters for some quick development work.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -493,7 +493,7 @@ class NeuralNet(object):
         y_pred = self.infer(Xi, **fit_params)
         loss = self.get_loss(y_pred, yi, X=Xi, training=True)
         loss.backward()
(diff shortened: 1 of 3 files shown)


skorch/net.py
```diff
@@ -493,7 +493,7 @@ def train_step(self, Xi, yi, **fit_params):
         y_pred = self.infer(Xi, **fit_params)
         loss = self.get_loss(y_pred, yi, X=Xi, training=True)
         loss.backward()
-        self.notify('on_grad_computed', parameters=self.module_.parameters())
```
benjamin-work: Should the argument not also be named `named_parameters` then?

ottonemo: Done

PR #157, opened 2018-02-28
Support (wildcard) param groups in initialization

ottonemo:


docs/user/neuralnet.rst
```diff
@@ -114,6 +114,16 @@ optimizer
 This should be a PyTorch optimizer,
 e.g. ``torch.optim.SGD``. After initializing the ``NeuralNet``, the
 initialized optimizer will stored in the ``optimizer_`` attribute.
+During initialization you can already define param groups to define
+different parametrizations for certain parameters identified by
+strings with globbing:
```
benjamin-work: This sentence is not very clear, maybe you could extend it a little bit (e.g. by giving an example or providing a link to the pytorch docs).

ottonemo: I wasn't sure if this is the best place to document this in detail. I can improve upon the existing text by adding some more details but I have the feeling that this is not the best place to explain param groups in all their glory.

benjamin-work: Nonetheless, I find the sentence a little awkward. Maybe read it out loud and you may see what I mean.

ottonemo:
@benjamin-work better?

Issue #161, opened 2018-03-09
RuntimeError: Expected object of type Variable[torch.LongTensor] but found type Variable[torch.IntTensor]


BrendanMartin:
When I try to run the intro example on the README, I get the error

```RuntimeError: Expected object of type Variable[torch.LongTensor] but found type Variable[torch.IntTensor] for argument #1 'target'```

What could be the problem? I haven't changed anything from the example.

I'm on Windows 10, running in a Jupyter notebook.

The full traceback:

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-90-8f9dd3f6d8ef> in <module>()
     36 )
     37 
---> 38 net.fit(X, y)
     39 y_proba = net.predict_proba(X)

~\Anaconda3\lib\site-packages\skorch\net.py in fit(self, X, y, **fit_params)
   1170         # this is actually a pylint bug:
   1171         # https://github.com/PyCQA/pylint/issues/1085
-> 1172         return super(NeuralNetClassifier, self).fit(X, y, **fit_params)
   1173 
   1174 

~\Anaconda3\lib\site-packages\skorch\net.py in fit(self, X, y, **fit_params)
    630             self.initialize()
    631 
--> 632         self.partial_fit(X, y, **fit_params)
    633         return self
    634 

~\Anaconda3\lib\site-packages\skorch\net.py in partial_fit(self, X, y, classes, **fit_params)
    595         self.notify('on_train_begin')
    596         try:
--> 597             self.fit_loop(X, y)
    598         except KeyboardInterrupt:
    599             pass

~\Anaconda3\lib\site-packages\skorch\net.py in fit_loop(self, X, y, epochs)
    540             for Xi, yi in self.get_iterator(dataset_train, training=True):
    541                 self.notify('on_batch_begin', X=Xi, y=yi, training=True)
--> 542                 loss = self.train_step(Xi, yi)
    543                 self.history.record_batch('train_loss', loss.data[0])
    544                 self.history.record_batch('train_batch_size', len(Xi))

~\Anaconda3\lib\site-packages\skorch\net.py in train_step(self, Xi, yi)
    463         self.optimizer_.zero_grad()
    464         y_pred = self.infer(Xi)
--> 465         loss = self.get_loss(y_pred, yi, X=Xi, training=True)
    466         loss.backward()
    467 

~\Anaconda3\lib\site-packages\skorch\net.py in get_loss(self, y_pred, y_true, X, training)
[...]

PR #162, opened 2018-03-13
Explicitly cast y to 64 bit for 32 bit archs

ottonemo:
Addresses #161. Some stacks with 32 bit components (e.g. numpy)
may default to 32 bit integers for `y` in `make_classification`
but torch expects 64 bit values.


benjamin-work:
Could you adjust the quickstart section of the docs as well (https://github.com/dnouri/skorch/blob/master/docs/user/quickstart.rst)?

ottonemo:
Done

PR #166, opened 2018-03-20
Callback leveraging livelossplot

ottonemo:
Live plotting in Jupyter Notebook.

This is a WIP but comments are welcome.

Uses https://github.com/stared/livelossplot/ to provide this feature.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -290,15 +290,19 @@ class LivePlot(skorch.callbacks.Callback):
       the default is to use training and validation loss.abs
     """
     def __init__(self, metrics=None):
+        # Import here to raise error on model instantiation
+        # when dependency not available.
+        import livelossplot as _
+
         self.metrics = metrics
+        self.plot = None
         if self.metrics is None:
             self.metrics = ['train_loss', 'valid_loss']
 
-    def on_train_begin(self, net, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


benjamin-work:
Is there any meaningful way to add a test to this feature? Or maybe add an example to a notebook?

benjamin-work:
I tried this:

```
# in Advanced_Usage.ipynb
net = NeuralNetClassifier(
    ClassifierModule,
    max_epochs=20,
    lr=0.1,
    callbacks=[
        ('plot', LivePlot()),
    ],
)
net.fit(X, y)
```

Interestingly, it only plotted the training loss, even though a valid loss exists. Furthermore, the output from `PrintLog` is cleared by the plot, but I guess we have to accept that.

`LivePlot` also had a 3x overhead for this specific example. This is not too relevant, but one might think about a feature to only update the plot every N epochs.

thomasjpfan:
If there is still interest in this, there is another similar library that may be useful: https://github.com/fastai/fastprogress

ottonemo:
I don't see that `fastprogress` does more than what we support now. The plotting is independent from the progress bar as far as I am concerned. I'd still be interested in something like `livelossplot` but I think that it would be better to have a better integration with plotting tools like visdom (for example).

Since this PR is open for quite a while I think we should abandon this feature in favor of better support for visdom or use a more powerful library such as https://github.com/waleedka/hiddenlayer.

PR #169, opened 2018-04-03
Resolve #145 by implementing location parameter

### LINKED ISSUE #145: forward / forward_iter use up GPU memory
Currently, calls to `forward` with large input data will process the data in batches but stores the final, concatenated result of all batches on the GPU which is problematic for most users (GPU memory is very limited). There are two relatively simple solutions to this problem:

1. We default to using CPU memory when storing predictions in `forward`
2. We introduce a parameter to choose the storage location in `forward` and `forward_iter`

The current state of storing all results on the GPU has the advantage that it is simple to further process the data outside of the module while still harnessing the performance gain of the GPU. Examples for this might be post-processing using CRF/HMM models or chaining other skorch models. For systems that need this sort of post-processing or have performance constraints it would be really good to have this feature. Therefore, it is probably better to be able to make this opt-in by using a parameter:

    outputs = net.forward(X, keep_on_gpu=True)

The parameter name should be easily differentiable to the other CUDA/GPU related parameters since this does not influence where results are calculated in `forward` but only determines the storage location.

githubnemo:
You can now pass a `location` paramter to `forward` which
indicates on which device to store the intermediate results.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -669,6 +669,8 @@ class NeuralNet(object):
 
     def forward_iter(self, X, training=False, location='cpu'):
         """Yield outputs of module forward calls on each batch of data.
+        The storage location of the yielded tensors is determined
+        by the ``location`` parameter.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


ottonemo:
This still needs a test but I would like to get some feedback on the general approach @benjamin-work.

benjamin-work:
Code LGTM, though I'm not sure what `torch.serialization.default_restore_location` does exactly (any overhead when using the default?).

Tests could be tricky here.

`forward` needs an update of the docstring (same as `forward_iter`).

ottonemo:
No overhead, see https://github.com/pytorch/pytorch/blob/master/torch/serialization.py#L62
For CUDA it would simply translate to a string split and `.cuda(device_id)`.

ottonemo:
@benjamin-work pls review

PR #170, opened 2018-04-04
fit with datasets

benjamin-work:
DON'T MERGE.

Addresses #146 

This is a proposal for how to implement fitting with datasets. To see how usage would look in practice, open the `proposal_datasets.ipynb`.


benjamin-work:
@ottonemo pls review again

@dnouri This is the proposal for making skorch fitting work with `Dataset`s. Maybe you would like to test whether this covers your use cases? There are some caveats with using `Dataset`s in relation to sklearn, have a look at `Advanced_Usage.ipynb`, which covers those. Fortunately, once we have implemented the discussed changes to `EpochScoring`, most of the caveats should vanish.

skorch/utils.py
```diff
@@ -136,6 +143,71 @@ def flatten(arr):
...
+     2
+    [torch.FloatTensor of size 2]
+
+    >>> multi_indexing([[1, 2, 3], [4, 5, 6]], np.s_[:2])
+    [[1, 2], [4, 5]]
+
+    >>> multi_indexing({'a': [1, 2, 3], 'b': [4, 5, 6]}, np.s_[-2:])
+    {'a': [2, 3], 'b': [5, 6]}
+
+    >>> multi_indexing(pd.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]}))
```
ottonemo: missing index parameter

skorch/utils.py
```diff
@@ -177,3 +249,31 @@ def params_for(prefix, kwargs):
...
+    X, y = _none, _none
+
+    if isinstance(dataset, Subset):
+        X, y = data_from_dataset(dataset.dataset)
+        X = multi_indexing(X, dataset.indices)
+        y = multi_indexing(y, dataset.indices) if y is not None else y
+    else:
+        if hasattr(dataset, 'X') and hasattr(dataset, 'y'):
+            X, y = dataset.X, dataset.y
+        elif hasattr(dataset, 'dataset'):
```
ottonemo: Which input does this case handle?

skorch/utils.py
benjamin-work: fixed

benjamin-work:
@ottonemo pls review again

Issue #175, opened 2018-04-11
set_params with unknown attribute does not raise


benjamin-work:
We have an inconsistency with sklearn, namely that calling ~~a `NeuralNet`'s or~~ a `Callback`'s `set_params` with an unknown attribute just sets this attribute. In sklearn, a `ValueError` is raised:

```ValueError: Invalid parameter foo for estimator LogisticRegression. Check the list of available parameters with `estimator.get_params().keys()`.```

ottonemo:
Waiting for an answer @ sklearn.

PR #178, opened 2018-04-12
set_params with unknown key on callback should raise

benjamin-work:
Previously, this parameter was just set as an attribute. The correct
behavior is to raise a ValueError, as sklearn would.

Addresses #175

### DIFF
--- a/skorch/callbacks/base.py
+++ b/skorch/callbacks/base.py
@@ -66,6 +66,5 @@ class Callback:
         return BaseEstimator.get_params(self, deep=deep)
 
     def set_params(self, **params):
-        for key, val in params.items():
-            setattr(self, key, val)
+        BaseEstimator.set_params(self, **params)
         return self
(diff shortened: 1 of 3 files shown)


benjamin-work:
Let's put this PR on hold for now. It seems that there was a [change](https://github.com/scikit-learn/scikit-learn/commit/75763cfe7842af6a579bb2f57c5b030d5c529115) in sklearn 0.19 -> 0.19.1 that results in a different output text from the ValueError. I actually prefer the old output text (not sure why that was changed or whether it was on purpose).

Old:

```
Invalid parameter foo for estimator LogisticRegression. Check the list of available parameters with `estimator.get_params().keys()`.
```

new

```
Invalid parameter foo for estimator LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
          intercept_scaling=1, max_iter=100, multi_class='ovr', n_jobs=1,
          penalty='l2', random_state=None, solver='liblinear', tol=0.0001,
          verbose=0, warm_start=False). Check the list of available parameters with `estimator.get_params().keys()`.```

So either we have to live with the new, verbose message or we have to raise our own ValueError and customize the message.

benjamin-work:
asked amueller about whether this is intended: https://github.com/scikit-learn/scikit-learn/commit/75763cfe7842af6a579bb2f57c5b030d5c529115

benjamin-work:
superseded by #188

PR #180, opened 2018-04-14
ENH: Adds CyclicLR

thomasjpfan:
Closes https://github.com/dnouri/skorch/issues/168.

1. Adds `on_batch_begin` to `LRScheduler` to step through `CyclicLR`. This uses `len(net.history[-1, 'batches'])` to get the batch index for the current epoch. Once there are more than one epoch, the total number of batches per epoch is obtained by `len(net.history[-2, 'batches'])` . 
2. `CyclicLR` is a subclass of `_LRScheduler` which allows for `LRScheduler(CyclicLR)`.
3. Tests were refactored a little to use `@pytest.mark.parametrize`. Additional tests were added to confirm that `LRScheduler` is calling `step` or `batch_step`  the expected number of times.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -58,7 +58,7 @@ class LRScheduler(Callback):
             self._lr_scheduler.step(epoch)
 
     def on_batch_begin(self, net, **kwargs):
-        if isinstance(self._lr_scheduler, CyclicLR):
+        if hasattr(self._lr_scheduler, 'batch_step') and callable(self._lr_scheduler.batch_step):
             batch_idx = self._get_batch_idx(net)
             self._lr_scheduler.batch_step(batch_idx)
 
@@ -258,23 +258,8 @@ class CyclicLR(object):
             raise TypeError('{} is not an Optimizer'.format(
                 type(optimizer).__name__))
         self.optimizer = optimizer
-
-        if isinstance(base_lr, (list, tuple)):
-            if len(base_lr) != len(optimizer.param_groups):
-                raise ValueError("expected {} base_lrs, got {}".format(
-                    len(optimizer.param_groups), len(base_lr)))
-            self.base_lrs = list(base_lr)
-        else:
-            self.base_lrs = [base_lr] * len(optimizer.param_groups)
-
-        if isinstance(max_lr, (list, tuple)):
-            if len(max_lr) != len(optimizer.param_groups):
-                raise ValueError("expected {} max_lrs, got {}".format(
-                    len(optimizer.param_groups), len(max_lr)))
-            self.max_lrs = list(max_lr)
-        else:
-            self.max_lrs = [max_lr] * len(optimizer.param_groups)
-
+        self.base_lrs = self._format_lr('base_lr', optimizer, base_lr)
+        self.max_lrs = self._format_lr('max_lr', optimizer, max_lr)
         self.step_size = step_size
 
         if mode not in ['triangular', 'triangular2', 'exp_range'] \
@@ -301,8 +286,18 @@ class CyclicLR(object):
         self.batch_step(last_batch_idx + 1)
         self.last_batch_idx = last_batch_idx
 
+    def _format_lr(self, name, optimizer, lr):
+        """Return correctly formatted lr for each param group."""
+        if isinstance(lr, (list, tuple)):
+            if len(lr) != len(optimizer.param_groups):
+                raise ValueError("expected {} values for {}, got {}".format(
+                    len(optimizer.param_groups), name, len(lr)))
+            return np.array(lr)
+        else:
+            return lr * np.ones(len(optimizer.param_groups))
+
     def step(self, epoch=None):
-        """Not used by ``CyclicLR``"""
+        """Not used by ``CyclicLR``, use batch_step instead."""
         pass
 
     def batch_step(self, batch_idx=None):
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -69,7 +69,7 @@ class LRScheduler(Callback):
             scheduler_kwargs["last_epoch"] = last_epoch
 
         if policy is CyclicLR and \
-           "batch_idx" not in scheduler_kwargs:
+           "last_batch_idx" not in scheduler_kwargs:
             last_batch_idx = self._get_batch_idx(net) - 1
             scheduler_kwargs["last_batch_idx"] = last_batch_idx
         return policy(net.optimizer_, **scheduler_kwargs)
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -75,7 +75,7 @@ class LRScheduler(Callback):
         return policy(net.optimizer_, **scheduler_kwargs)
 
     def _get_batch_idx(self, net):
-        if len(net.history) == 0:
+        if not net.history:
             return -1
         epoch = len(net.history) - 1
         current_batch_idx = len(net.history[-1, 'batches'])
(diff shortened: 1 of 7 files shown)


BenjaminBossan:
This looks like a great addition. Also thanks for improving the tests. I will have a more thorough look at it soon, but I noticed one thing right away: Would you mind changing the docstring to numpy style, like in the rest of skorch? Thank you

thomasjpfan:
I just updated this PR with the following changes:

1. Converted `CyclicLR`'s docstring to numpy style.
2. Adds `__all__` to `callbacks/__init__.py` so that sphinx's `automodule: skorch.callbacks` can detect the docstrings.
3. Fixes most warnings from sphinx doc generation.

benjamin-work:
Thank you for the amendments.

I have a question regarding usability. Currently, a lot of work in `CyclicLR` and `WarmRestartLR` is done during `__init__`. My proposal is to move that to the `initialize` method of the callback instead (while saving the parameters as they come during `__init__`). This would have two advantages in my opinion:

* It works better with `set_params(...)`. This method does not use `__init__`, so all the work done in `__init__` (such as parameter validation) would be skipped. Therefore, the change would prevent users from setting parameters via `set_params` that cause trouble even though they work during `__init__`. `set_params` is used for things such as `GridSearchCV`.
* The change would allow us to leave the `optimizer` parameter to be None when initializing the callbacks. The callback would then take the net's optimizer during `callback.initialize` (something we would need to add). This is more user friendly, since typically in skorch, we don't have an initialized optimizer at the time we instantiate the callbacks, making the current usage slightly awkward.

The proposed changes would allow code like this to work:

```
net = NeuralNetClassifier(
    MyModule,
    callbacks=[CyclicLR(optimizer=None)],
)
net.fit(X, y)
```

(We would still allow to pass the optimizer explicitely.)

We can always implement these changes at a later stage. My question is rather whether you see a problem with that.

thomasjpfan:
It feels like skorch likes to have `LRScheduler` be the user facing interface for LR scheduling, i.e.:

```python
net = NeuralNetClassifier(
    MyModule,
    callbacks=[LRScheduler('CyclicLR')],
)
net.fit(X, y)
```

Nevertheless, I think it is a good idea to move the following lines in `LRScheduler`:

https://github.com/dnouri/skorch/blob/218716e0631ab33c820562a8b18c60380ba0f6ff/skorch/callbacks/lr_scheduler.py#L46-L49

to the `initialize` callback. This would require the `optimizer` to be initialized before the callbacks are initialized. The order of initialization is defined here:

https://github.com/dnouri/skorch/blob/218716e0631ab33c820562a8b18c60380ba0f6ff/skorch/net.py#L428-L440

We would need to move the `self.initialize_optimizer()` call before the `self.initialize_callbacks()`. Is there any issue with changing the order of initialization?

thomasjpfan:
There are two places where `_LRScheduler` stores state:

1. It uses last_epoch to remember the last epoch. `LRscheduler` avoids using this internal state by always passing in the current epoch to the `step` function.

2. It stores `initial_lr` in `optimizer.param_groups`, this must be done once when the `last_epoch` parameter is -1. My latest updates uses `net.history` to get this information and passes it along.

All other parameters are stored in `LRScheduler`'s kwargs which works with `set_params`. With the latest updates, I think leaving the initialization in `on_train_begin` is okay, and we won't need change the `callback.initialize` interface.

I made a few more additional tweaks:

1. The accuracy score in `test_net_learns` went to 0.693 so I lowered the test from 0.7 to 0.65. The `test_net.py` tests does not use LRScheduler, so this was a bit strange.
2. I converted `CyclicLR` to not be a subclass of `_LRScheduler` just like how `ReduceLROnPlateau` is not a subclass. A user would have to pass in the strings: "CyclicLR" or "ReduceLROnPlateau" to `LRScheduler` to use them.

benjamin-work:
Hi Thomas,
Thanks for putting so much thought into this PR. I will probably not have the time to review your changes this week, but will start next week as soon as possible. I hope that's okay.
Cheers,
Benjamin

skorch/callbacks/lr_scheduler.py
```diff
@@ -140,3 +162,176 @@ def get_lr(self):
...
+    def _triangular_scale_fn(self, x):
+        return 1.
+
+    def _triangular2_scale_fn(self, x):
+        return 1 / (2. ** (x - 1))
+
+    def _exp_range_scale_fn(self, x):
+        return self.gamma**(x)
+
+    def get_lr(self):
```
benjamin-work: A short docstring would be nice here.

skorch/callbacks/lr_scheduler.py
```diff
@@ -140,3 +162,176 @@ def get_lr(self):
...
+        step_size = float(self.step_size)
+        cycle = np.floor(1 + self.last_batch_idx / (2 * step_size))
+        x = np.abs(self.last_batch_idx / step_size - 2 * cycle + 1)
+
+        lrs = []
+        param_lrs = zip(self.optimizer.param_groups, self.base_lrs, self.max_lrs)
+        for param_group, base_lr, max_lr in param_lrs:
```
benjamin-work: `param_group` is not used, can be replaced by `_`.

thomasjpfan:
I have updated the PR with your suggestions!

skorch/callbacks/lr_scheduler.py
```diff
@@ -140,3 +162,192 @@ def get_lr(self):
...
+        else:
+            self.base_lrs = [base_lr] * len(optimizer.param_groups)
+
+        if isinstance(max_lr, (list, tuple)):
+            if len(max_lr) != len(optimizer.param_groups):
+                raise ValueError("expected {} max_lrs, got {}".format(
+                    len(optimizer.param_groups), len(max_lr)))
+            self.max_lrs = list(max_lr)
+        else:
+            self.max_lrs = [max_lr] * len(optimizer.param_groups)
```
ottonemo: The `base_lr` and `max_lr` initialization could be refactored in a similar way to `WarmRestartLR._format_lr` but we can do this later.

skorch/tests/test_lr_scheduler.py
```diff
@@ -63,6 +43,50 @@ def _lr_callback_init_policies(
...
+    @pytest.mark.parametrize('policy, kwargs', [
+            ('CyclicLR', {}),
+        ])
+    def test_lr_callback_batch_steps_correctly(self, classifier_module, policy, kwargs):
+        num_examples = 1000
+        batch_size = 100
+        max_epochs = 2
+
+        X, y = make_classification(num_examples, 20, n_informative=10, random_state=0)
+        X = X.astype(np.float32)
```
ottonemo: You can use the `classifier_data` fixture defined in `conftest.py` for both of these operations.

skorch/tests/test_lr_scheduler.py
```diff
@@ -63,6 +43,50 @@ def _lr_callback_init_policies(
...
+            ('StepLR', {'step_size': 30}),
+            ('MultiStepLR', {'milestones': [30, 90]}),
+            ('ExponentialLR', {'gamma': 0.1}),
+            ('ReduceLROnPlateau', {}),
+            ('WarmRestartLR', {}),
+        ])
+    def test_lr_callback_steps_correctly(self, classifier_module, policy, kwargs):
+        max_epochs = 2
+        X, y = make_classification(1000, 20, n_informative=10, random_state=0)
+        X = X.astype(np.float32)
```
ottonemo: You can use the `classifier_data` fixture defined in `conftest.py` for both of these operations.

skorch/callbacks/lr_scheduler.py
```diff
@@ -56,16 +57,37 @@ def on_epoch_begin(self, net, **kwargs):
         else:
             self._lr_scheduler.step(epoch)
 
+    def on_batch_begin(self, net, **kwargs):
+        if isinstance(self._lr_scheduler, CyclicLR):
```
ottonemo: Maybe `hasattr(self._lr_scheduler, 'batch_step') and callable(self._lr_scheduler.batch_step)` would be better suited here?

thomasjpfan:
This PR is updated with @ottonemo's suggestions.

benjamin-work:
thx @thomasjpfan

PR #185, opened 2018-04-26
ENH: Adds pytorch 0.4 support

thomasjpfan:
Closes https://github.com/dnouri/skorch/issues/184


skorch/net.py
```diff
@@ -1306,14 +1307,9 @@ def check_data(self, X, y):
             raise ValueError(msg)
 
     def _prepare_target_for_loss(self, y):
-        # This is a temporary, ugly work-around (relating to #56), but
-        # currently, I see no solution that would result in a 1-dim
-        # LongTensor after passing through torch's DataLoader. If
-        # there is, we should use that instead. Otherwise, this will
-        # be obsolete once pytorch scalars arrive.
```
ottonemo: This comment is still valid though. This is a workaround for the case where you pass a tensor of shape `(n,)` as `y` to the data loader which will result in `y.shape = (n,1)` which we correct to its original shape `(n,)` here.

You can verify for yourself that this is still the case:

    X = torch.zeros(10, 5)
    y = torch.zeros(10)
    d = TensorDataset(X, y)
    l = DataLoader(d)
    Xi, yi = next(iter(l))
    assert len(yi.size()) == 1 # fails, is 2

benjamin-work: For me it's 1!

thomasjpfan: `len(yi.size())` is 1 in pytorch 0.3.1 and 0.4.

githubnemo:
@BenjaminBossan I think this PR is ready. Do you think we can merge this so at least the master branch works with pytorch 0.4.0 until we release the next version?

BenjaminBossan:
I can't merge before Wednesday ;) There are some references to Variables still left that could be removed, as well as a typo in your deprecation warning message, but that can be fixed later.

benjamin-work:
@ottonemo I fixed the points I addressed earlier. If it's fine by you, I would merge now.

PR #209, opened 2018-05-09
lr scheduler can now monitor other fields

benjamin-work:
`LRScheduler` now allows you to monitor vars other than train_loss.

The class now takes a 'monitor' argument (like `Checkpoint`). If this is
a string, this value from the history is passed to step. If it is a
callable, the value returned by calling this callable with net is is
passed to step.

Moreover, some minor clean ups.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -29,9 +29,10 @@ class LRScheduler(Callback):
       Learning rate policy name or scheduler to be used.
 
     monitor : str or callable (default=None)
-      Value of the history to monitor or callback that determines whether
-      this epoch should to a checkpoint. The callback takes the network
-      instance as parameter.
+      Value of the history to monitor or function/callable that. In
+      the latter case, the callable receives the net instance as
+      argument and is expected to return the score (float) used to
+      determine the learning rate adjustment.
 
     kwargs
       Additional arguments passed to the lr scheduler.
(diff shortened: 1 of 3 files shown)


skorch/callbacks/lr_scheduler.py
```diff
@@ -34,13 +28,19 @@ class LRScheduler(Callback):
     policy : str or _LRScheduler class (default='WarmRestartLR')
       Learning rate policy name or scheduler to be used.
 
+    monitor : str or callable (default=None)
+      Value of the history to monitor or callback that determines whether
+      this epoch should to a checkpoint. The callback takes the network
+      instance as parameter.
```
ottonemo: Maybe add a note what values to expect (in case of `Checkpoint` we want boolean values, here we want real values, right?)

benjamin-work:
@ottonemo pls review again.

skorch/callbacks/lr_scheduler.py
```diff
@@ -29,9 +29,10 @@ class LRScheduler(Callback):
       Learning rate policy name or scheduler to be used.
 
     monitor : str or callable (default=None)
-      Value of the history to monitor or callback that determines whether
-      this epoch should to a checkpoint. The callback takes the network
-      instance as parameter.
+      Value of the history to monitor or function/callable that. In
```
ottonemo: -that

benjamin-work: Can't write...

PR #210, opened 2018-05-09
Resolve #205: Add EarlyStopping callback

### LINKED ISSUE #205: Earlystop callback for skorch
Is there any example of showing earlystop callback in skorch?

ottonemo:


ottonemo:
You are right, this is certainly a nice feature to have.

We could also add an epsilon parameter to `Scoring`. One can always add a second score with a higher tolerance if one wants both scores.

benjamin-work:
IMO, it would be nice if the callback had similar parameters as [`ReduceLROnPlateau`](https://pytorch.org/docs/master/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau), with the difference being that training is stopped instead of learning rate reduced.

PR #211, opened 2018-05-09
Resolve #207: Support optimizers requiring closures

### LINKED ISSUE #207: Support for optimizers that require a closure
Right now, the [toy examples](https://nbviewer.jupyter.org/github/dnouri/skorch/blob/master/notebooks/Basic_Usage.ipynb) break if the `torch.optim.LBFGS` optimizer is used (likely also some others, e.g. conjugate gradients).

As per https://pytorch.org/docs/stable/optim.html#optimizer-step-closure it would seem that the closure way of calling the optimizer is preferred (and certainly required by those optimizers)

Right now, for my purposes, I've subclassed the `skorch.NeuralNetRegressor` class, overriding the `train_step` method with the following code. It seems to work, I'm just not sure if it interferes with some other skorch machinery. 
```
    def train_step(self, Xi, yi, **fit_params):
        y_pred = None
        loss = None
        def train_closure():
            nonlocal y_pred
            nonlocal loss
            self.module_.train()
            self.optimizer_.zero_grad()
            y_pred = self.infer(Xi, **fit_params)
            loss = self.get_loss(y_pred, yi, X=Xi, training=True)
            loss.backward()
            self.notify(
            'on_grad_computed',
            named_parameters=list(self.module_.named_parameters())
            )
            return loss

        self.optimizer_.step(train_closure)
        return {
            'loss': loss,
            'y_pred': y_pred,
            }
```

Are there any downsides to this solution? Should I prepare a pull request?

BTW, it could be useful to disable batching, as e.g. LBFGS does not use them.
There is the obvious workaround to set batch size to len(training set), but being able to pass None as batch_size would be clearer, I think

tpietruszka:
Should resolve #207 
Please comment!

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -434,10 +434,11 @@ class NeuralNet(object):
 
         self.optimizer_ = self.optimizer(*args, **kwargs)
 
-        # some optimizers require scoring closures to call them multiple times during each step. Those need special
-        # treatment in train_step() - the actually used
-        optimizer_step_signature = inspect.signature(self.optimizer_.step)
-        self.optimizer_is_iterative_= optimizer_step_signature.parameters['closure'].default is inspect.Parameter.empty
+        # optimizers that require a closure argument have to be treated
+        # differently. See `NeuralNet.train_step()` for details
+        optim_sig = inspect.signature(self.optimizer_.step)
+        self.optimizer_is_iterative_ = \
+            optim_sig.parameters['closure'].default is inspect.Parameter.empty
 
     def initialize_history(self):
         """Initializes the history."""
@@ -512,6 +513,15 @@ class NeuralNet(object):
         loss = None
 
         def loss_closure():
+            """ Calculate the loss function - called by the optimizer.
+
+            https://pytorch.org/docs/stable/optim.html#optimizer-step-closure
+            states: 'Some optimization algorithms such as Conjugate Gradient
+            and LBFGS need to reevaluate the function multiple times, so you
+            have to pass in a closure that allows them to recompute your model.
+            The closure should clear the gradients, compute the loss, and
+            return it.'
+            """
             nonlocal y_pred
             nonlocal loss
             self.module_.train()
@@ -528,7 +538,8 @@ class NeuralNet(object):
         self.optimizer_.step(loss_closure)
 
         if self.optimizer_is_iterative_:
-            # re-computing as we cannot know if the results of the last loss_closure() were actually used
+            # re-computing, as we cannot know if the results of the last
+            # loss_closure() call were the ones actually used
             y_pred = self.infer(Xi, **fit_params)
             loss = self.get_loss(y_pred, yi, X=Xi, training=True)
 
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -6,6 +6,7 @@ from functools import partial
 import re
 import tempfile
 import warnings
+import inspect
 
 import numpy as np
 from sklearn.base import BaseEstimator
@@ -433,6 +434,11 @@ class NeuralNet(object):
 
         self.optimizer_ = self.optimizer(*args, **kwargs)
 
+        # some optimizers require scoring closures to call them multiple times during each step. Those need special
+        # treatment in train_step() - the actually used
+        optimizer_step_signature = inspect.signature(self.optimizer_.step)
+        self.optimizer_is_iterative_= optimizer_step_signature.parameters['closure'].default is inspect.Parameter.empty
+
     def initialize_history(self):
         """Initializes the history."""
         self.history = History()
@@ -502,22 +508,34 @@ class NeuralNet(object):
           the module and to the train_split call.
 
         """
-        self.module_.train()
-        self.optimizer_.zero_grad()
-        y_pred = self.infer(Xi, **fit_params)
-        loss = self.get_loss(y_pred, yi, X=Xi, training=True)
-        loss.backward()
+        y_pred = None
+        loss = None
 
-        self.notify(
-            'on_grad_computed',
-            named_parameters=list(self.module_.named_parameters())
-        )
+        def loss_closure():
+            nonlocal y_pred
+            nonlocal loss
+            self.module_.train()
+            self.optimizer_.zero_grad()
+            y_pred = self.infer(Xi, **fit_params)
+            loss = self.get_loss(y_pred, yi, X=Xi, training=True)
+            loss.backward()
+            self.notify(
+                'on_grad_computed',
+                named_parameters=list(self.module_.named_parameters())
+            )
+            return loss
 
-        self.optimizer_.step()
+        self.optimizer_.step(loss_closure)
+
+        if self.optimizer_is_iterative_:
+            # re-computing as we cannot know if the results of the last loss_closure() were actually used
+            y_pred = self.infer(Xi, **fit_params)
+            loss = self.get_loss(y_pred, yi, X=Xi, training=True)
+
         return {
             'loss': loss,
             'y_pred': y_pred,
-            }
+        }
 
     def evaluation_step(self, Xi, training=False):
         """Perform a forward step to produce the output used for
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -1232,7 +1232,7 @@ class TestNeuralNet:
         # TODO: check error message more precisely, depending on what
         # the intended message shouldb e from sklearn side
         assert exc.value.args[0].startswith('Invalid parameter foo for')
-        
+
     @pytest.fixture()
     def sequence_module_cls(self):
         """Simple sequence model with variable size dim 1."""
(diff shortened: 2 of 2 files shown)


BenjaminBossan:
Thanks for `step`ping in, we didn't have this problem on the radar.

Overall, this looks good to me but I wonder about the use of inspecting here. We should think about the pros and cons of other options, namely 1) always using closures or 2) giving the user a choice of whether to use it (or some other way).

skorch/net.py
```diff
@@ -502,22 +509,44 @@ def train_step(self, Xi, yi, **fit_params):
...
+            y_pred = self.infer(Xi, **fit_params)
+            loss = self.get_loss(y_pred, yi, X=Xi, training=True)
+            loss.backward()
+            self.notify(
+                'on_grad_computed',
+                named_parameters=list(self.module_.named_parameters())
+            )
+            return loss
+
+        self.optimizer_.step(loss_closure)
```
BenjaminBossan: If I understood @ottonemo correctly, we can get the loss here.

tpietruszka: Indeed, we can get the loss returned from `self.optimizer_.step(loss_closure)` - except we still need y_pred. I thought getting both the same way would be clearer. 

On top of that, LBFGS seems to return the "wrong" (in a way) loss value. (from before it did the optimizations, I believe). The one we actually want is neither that, nor the last result of the closure() call.

To clarify:
- we are using the closure-type call for all optimizers
- some optimizers (just 1 of the default ones) actually require the closure argument - those might call it multiple times. I think we can make less assumptions about how they operate. We cannot know results of which call were finally used - so we re-calculate loss and y_pred for those optimizers.

We could calculate only y_pred in those cases, and get loss as you have mentioned, but again - it would be problematic for LFBGS.

benjamin-work: I have to admit that I haven't wrap my head around all possible implications of using LBFGS (or other optimizers that call the closure more than once), but here is a proposal, maybe you can comment on that:

```
class TrainClosure:
    def __init__(self, net, Xi, yi, **fit_params):
        self.net = net
        self.Xi = Xi
        self.yi = yi
        self.fit_params = fit_params

        self.losses = []
        self.y_preds = []
        self.num_calls = 0

    def __call__(self):
        self.num_calls += 1

        self.net.module_.train()
        self.net.optimizer_.zero_grad()

        y_pred = self.net.infer(self.Xi, **self.fit_params)
        loss = self.net.get_loss(y_pred, self.yi, X=self.Xi, training=True)
        loss.backward()
        self.net.notify(
            'on_grad_computed',
            named_parameters=list(self.net.module_.named_parameters())
        )

        self.losses.append(loss)
        self.y_preds.append(y_pred)

        return loss

# in NeuralNet
    def train_step(self, Xi, yi, **fit_params):
        y_preds = []
        losses = []

        from skorch.utils import LossClosure
        loss_closure = LossClosure(
            net=self,
            Xi=Xi,
            yi=yi,
            losses=losses,
            y_preds=y_preds,
            **fit_params
        )

        self.optimizer_.step(loss_closure)
        if loss_closure.num_calls == 1:
            loss = loss_closure.losses[0]
            y_pred = loss_closure.y_preds[0]
        else:
            # do something smart
            pass

        return {
            'loss': loss,
            'y_pred': y_pred,
        }
```

That way, we would collect all losses and predictions generated by calling the "closure". Would this allow us to do something smart (e.g. use last loss and prediction) and skip the whole inspection-thing?

ottonemo: I also dislike the use of `inspect` and I would like to keep the ability to easily overwrite the training step while focusing on the module's behavior rather than doing bookkeeping.
A solution would be to create a new train step function `train_step_meta` (working title)
which creates a wrapper closure similar to @benjamin-work 's approach which is
then able to return one set of losses depending on choice the user can make 
(e.g. mean, median, first, last, ...):

```
class TrainClosure:
    def __init__(self):
        self.losses = []
        self.y_preds = []
        self.num_calls = 0

    def __call__(self, f):
        def step():
            self.num_calls += 1

            step = f()

            self.losses.append(step['loss'])
            self.y_preds.append(step['y_pred'])

            return loss
        return step

    @staticmethod
    def extract_values(cls, losses, y_preds):
        if train_closure.num_calls == 1:
            loss = train_closure.losses[0]
            y_pred = train_closure.y_preds[0]
        else:
            # do something smart
            pass
        return loss, y_pred




# in NeuralNet

    def train_step(self, Xi, yi, **fit_params):
        self.module_.train()
        self.optimizer_.zero_grad()

        y_pred = self.infer(Xi, **fit_params)

        loss = self.get_loss(y_pred, yi, X=Xi, training=True)
        loss.backward()

        self.notify(
            'on_grad_computed',
            named_parameters=list(self.module_.named_parameters())
        )

        return {
            'loss': loss,
            'y_pred': y_pred,
        }

    # called in fit_loop instead of train_step
    def train_step_meta(self, Xi, yi, **fit_params):
        from skorch.utils import TrainClosure
        train_closure = TrainClosure()
        step = train_closure(self.train_step)

        self.optimizer_.step(step)

        loss, y_pred = train_closure.extract_values(
            train_closure.losses,
            train_closure.y_preds,
        )

        return {
            'loss': loss,
            'y_pred': y_pred,
        }
```

benjamin-work: I like this approach the most so far. I would consider having the `extract_values` method on `NeuralNet` instead, returning the last element by default (which is probably what we want for LBFGS).

@tpietruszka any thoughts?

tpietruszka: Sorry for not getting back to you earlier.
This is certainly better - less hacky -  than my original code. 

I think until now I misunderstood what the `loss` and `y_pred` returned by `train_step` should be. 

Now I think that those values should be from before the "current" step's optimization was applied (this is what happens for the typical case of optimizers not requiring closures)

If I'm correct, even if an optimizer calls the closure multiple times, we should always return the first values of loss and y_pred.
Would you agree? This assumption should simplify the code above - `extract_values` should always do the same thing

ottonemo: Yes, the default case should match the current behaviour as you stated. That means we have a parametrised train closure that is potentially able to collect multiple results over time (but doesn't) and an extraction method on the net that extracts loss and prediction from the closure. (Unless I forgot something, @benjamin-work)

ottonemo: Proposal:
```python
class TrainClosure:
    def __init__(self):
        self.step = None

    def __call__(self, f):
        def step(*args, **kwargs):
            step = f(*args, **kwargs)

            if self.step is None:
                self.step = step

            return step['loss']
        return step

    def unpack(self):
        return [self.step]


# in NeuralNet

    def train_step(self, Xi, yi, **fit_params):
        self.module_.train()
        self.optimizer_.zero_grad()

        y_pred = self.infer(Xi, **fit_params)

        loss = self.get_loss(y_pred, yi, X=Xi, training=True)
        loss.backward()

        self.notify(
            'on_grad_computed',
            named_parameters=list(self.module_.named_parameters())
        )

        return {
            'loss': loss,
            'y_pred': y_pred,
        }

    # called in fit_loop instead of train_step
    def train_step_meta(self, Xi, yi, **fit_params):
        train_closure = self.get_train_closure()
        step = train_closure(self.train_step)

        self.optimizer_.step(step)
        loss, y_pred = self.unpack_train_closure(train_closure)

        return {
            'loss': loss,
            'y_pred': y_pred,
        }
```

benjamin-work: This looks good. Some comments:

* What does `get_train_closure` do, only returning an instance of `TrainClosure`? If so, would it make sense to have `get_train_step` instead that returns the `step_fn`?
* Does `unpack` return a list for the case that I want to return more than one step?
* How about simply calling the new method `_train_step`?

ottonemo: > What does get_train_closure do, only returning an instance of TrainClosure? If so, would it make sense to have get_train_step instead that returns the step_fn?

Yes, that would make sense.

> Does unpack return a list for the case that I want to return more than one step?

Exactly. `unpack_train_closure` can then decide what to do with multiple results, the naïve implementation would simply return the first element.

> How about simply calling the new method _train_step?

Don't like it. The name should convey some sense of hierarchy. Maybe `outer_train_step` or something like that.

ottonemo: @tpietruszka Do you think this proposal works for you and this PR? If you want we can also split this PR and make the necessary training closure changes ourselves, considering how much work you already put in.

tpietruszka: I think this approach works great.

About method naming - how about keeping the "outer" method `train_step` , and calling the "inner" e.g. `compute_gradient`? That's what it actually does, I think.

On that note, maybe the TrainClosure could be something like GradientClosure? 

A question here: how do you think a user should supply their own logic to handle multiple loss values/predicted values?

Would there be an additional parameter to the `skorch.net.NeuralNet` class constructor - with the default value of skorch.net.TrainClosure? Or did you have something else in mind? (there are quite many constructor parameters already, and this would not be a common customization. Some option would be to just subclass `skorch.net.NeuralNet` and override `get_train_step`)

Anyway, I can change  this pull request using your approach. Would be ready this week for sure.

But since it's a totally different approach than I'd proposed originally you can also close this PR and start a new one. Please let me know!

benjamin-work:
See #243

PR #213, opened 2018-05-09
Add intersphinx extension to link to PyTorch/Sklearn documentation

taketwo:


BenjaminBossan:
Thank you for the great addition. I saw a couple of instances where links were still missing (nn.Module, DataLoader, Dataset), but that can be fixed later.

I will be back in office on Monday, so can't merge before that.

taketwo:
Yeah, there are some instances that I did not update. My idea was to create a link the first time an external Pytorch class is mentioned, but keep it plain text in subsequent text. Because of the namespacing, Pytorch links are a bit verbose and look heavy (`torch.utils.data.DataLoader`), so I thought this would negatively affect the readability.

That being said, I can link all of them, if that is preferred.

BenjaminBossan:
Ah okay, I didn't think of that. I could see either working. @ottonemo what is your opinion?

taketwo:
Actually, in Sphinx it is possible to [strip the namespaces from links](https://stackoverflow.com/a/43686845/1525865). Thus I propose the following approach: always link PyTorch classes, stripping namespaces. When such a class is mentioned for the first time in the current context, qualify with plain text "PyTorch". For example:

image

BenjaminBossan:
That looks like a very good compromise.

taketwo:
I went ahead and added hyperlinks not only to PyTorch classes, but also Scikit-learn and Skorch own classes, as well as methods. There is one paragraph though where I am not sure if I referred to correct methods. Please double-check, I'll leave inline comment.

docs/user/neuralnet.rst
```diff
@@ -317,43 +330,50 @@ predict(X) and predict_proba(X)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 These methods perform an inference step on the input data and return
-``numpy array``\s. By default, ``predict_proba`` will return whatever
-it is that the ``module``\'s ``forward`` method returns, cast to a
-``numpy array``. If ``forward`` returns multiple outputs as a tuple,
+``numpy array``\s. By default,
+:func:`~skorch.net.NeuralNet.predict_proba` will return whatever it is
+that the ``module``\'s :func:`~torch.nn.Module.forward` method
```
taketwo: Here and in the next paragraph, please check if I picked correct classes' forward/forward_iter methods.

BenjaminBossan: I didn't catch any error here.

PR #214, opened 2018-05-10
Resolve #205: Add EarlyStopping callback - modified

### LINKED ISSUE #205: Earlystop callback for skorch

tpietruszka:
Based on #210 should resolve #205 

A solution for ignoring very small changes in the EarlyStopping callback. 
As suggested by @benjamin-work  - parameters similar to [ReduceLROnPlateau](https://pytorch.org/docs/master/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau), altough I used `lower_is_better` as in Scoring callbacks. 

I am not very happy about how `threshold_mode` is passed - should be a boolean flag `threshold_relative` instead?

**Please comment on the approach and the interface, if it is okay I will prepare more tests etc.**

I thought about ignoring small improvements inside of Scoring callbacks, but found it awkward to set `valid_loss_best` flag to False when it is actually best. 

Maybe we can just calculate scores in Scoring and then interpret them in Early Stopping. Perhaps other callbacks will need to interpret them differently.

Plus, I think in this version most use cases won't require additional Scoring callbacks.


skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,71 @@
...
+        return EarlyStopping
+
+    @pytest.fixture
+    def net_clf_cls(self):
+        from skorch import NeuralNetClassifier
+        return NeuralNetClassifier
+
+    @pytest.fixture
+    def broken_classifier_module(self, classifier_module):
+        """Return a classifier that does not imporve over time."""
```
BenjaminBossan: typo "imporve"

BenjaminBossan:
Thank you for your contribution. I will be back in office on Monday and then will look more thoroughly at this PR. Overall, though, I think the interface looks very usable and intuitive.

> I thought about ignoring small improvements inside of Scoring callbacks, but found it awkward ...

Yes, best leave Scoring as is.

> Maybe we can just calculate scores in Scoring and then interpret them in Early Stopping

This sounds like the right way.

> I am not very happy about how threshold_mode is passed - should be a boolean flag threshold_relative instead?

At the moment, this sounds like it should be boolean, but it would conflict with the pytorch parameter. I could imagine that there could be other modes in the future, in which case having this as str now is the better option.

> if it is okay I will prepare more tests etc.

I'm positive this will go through, so please go ahead and add more tests to cover the remaining cases.

One question I have is whether `misses_` should be reset `on_train_begin` or not rather at `initialize`. Say `patience=5` and I fit the model for X epochs, after which I have 3 misses. Then I continue fitting with `partial_fit` for Y epochs. Would I not expect that after 2 additional misses, training is stopped? If so `misses_` should not be reset `on_train_begin`. With the current implementation, I get different results when first training X, then Y epochs, than I would get when immediately training X+Y epochs.

benjamin-work:
@tpietruszka are you still working on this?

benjamin-work:
Glad you're still working on this. Take your time.

You raise some valid concerns about the use case of `partial_fit`. Personally, I often use it while working interactively, when I just want to continue training for a couple of epochs using the same data. There are thus arguments for both possibilities.

Maybe @ottonemo can break the tie :)

ottonemo:
`partial_fit` allows for manually fitting the model without re-initialization and, for any case I can think of, it would be surprising to me if partial fit was interrupted by early stopping just because the previous `fit` call reached that point (maybe I'm calling partial fit to advance just a little bit further to see where the results go, for example). So I'd vote for resetting state in `on_train_begin`.

While I see Benjamin's point that it is generally a good idea to satisfy `20*[partial_fit()] <=> fit(epochs=20)` I am not sure if that even holds right now considering, for example, learning rate schedulers which reset their state on train begin which is called with every `partial_fit()`. There is also the issue that `partial_fit` is currently *exactly* like `fit` with the exception of doing a warm-start. So for 'real' partial fitting you'd have to set `max_epochs=1` before calling `partial_fit`.

skorch/callbacks/training.py
```diff
@@ -90,3 +91,88 @@ def on_epoch_end(self, net, **kwargs):
...
+    patience : int (default=5)
+      Number of epochs to wait for improvement of the monitor value until
+      the training process is stopped.
+
+    threshold : int (default=1e-4)
+      Ignore score improvements smaller than `threshold`.
+
+    threshold_mode : str (default='rel')
+        One of `rel`, `abs`. In `rel` mode,
+        dynamic_threshold = best * ( 1 + threshold ) in 'max'
```
taketwo: Usage of 'min' and 'max' mode terminology here is confusing, because it sounds like these are two more possible options for `threshold_mode`.

tpietruszka: Fixed (i hope)

skorch/tests/callbacks/test_early_stopping.py
tpietruszka: Fixed

tpietruszka:
I've fixed the comments and added some tests. 
My understanding was that we are leaving re-setting the `misses_` field as is.

Is there anything else to be done?

skorch/callbacks/training.py
```diff
@@ -90,3 +91,79 @@ def on_epoch_end(self, net, **kwargs):
...
+        self.misses_ = 0
+        self.dynamic_threshold_ = None
+
+    # pylint: disable=arguments-differ
+    def on_train_begin(self, net, **kwargs):
+        if self.threshold_mode not in ['rel', 'abs']:
+            raise ValueError("Invalid threshold mode: '{}'"
+                             .format(self.threshold_mode))
+        self.misses_ = 0
+        self.dynamic_threshold_ = \
```
benjamin-work: No need to break this line, it fits within 1 line.

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+        net = net_clf_cls(
+            classifier_module,
+            callbacks=[
+                early_stopping_cb,
+            ],
+            max_epochs=max_epochs,
+        )
+        net.fit(*classifier_data)
+
+        ok_run = (
```
benjamin-work: This can be abbreviated to `assert len(net.history) == max_epochs`

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+        net = net_clf_cls(
+            broken_classifier_module,
+            callbacks=[
+                early_stopping_cb,
+            ],
+            max_epochs=max_epochs,
+        )
+        net.fit(*classifier_data)
+
+        bad_run = (
```
benjamin-work: Can we not be more precise here? I.e. instead of `len(net.history) != max_epochs`, we know that `assert len(net.history) == patience + 1 < max_epochs`.

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+            classifier_module,
+            callbacks=[
+                scoring_cb,
+                early_stopping_cb,
+            ],
+            max_epochs=max_epochs,
+        )
+        net.fit(*classifier_data)
+
+        ok_run = (
```
benjamin-work: again, a one-liner should be enough here

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
+"""Tests for early stopping callback"""
+
+import numpy as np
```
benjamin-work: np is not needed

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+
+        stopped_run = (
+            len(net.history) != max_epochs
+        )
+        assert stopped_run
+
+    def test_wrong_threshold_mode(
+            self, net_clf_cls, classifier_module, classifier_data,
+            early_stopping_cls):
+        with pytest.raises(ValueError) as e_info:
```
benjamin-work: You assign `e_info` but do not use it. You could test that the message is correct:

`assert exc.value.args[0] == expected_msg`

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+            broken_classifier_module,
+            callbacks=[
+                scoring_cb,
+                early_stopping_cb,
+            ],
+            max_epochs=max_epochs,
+        )
+        net.fit(*classifier_data)
+
+        stopped_run = (
```
benjamin-work: again, one-liner is better here

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+        net.fit(*classifier_data)
+
+        bad_run = (
+            len(net.history) != max_epochs
+        )
+
+        assert bad_run
+
+    def test_custom_scoring_nonstop(
+        self, net_clf_cls, classifier_module, classifier_data,
```
benjamin-work: This line and the next should be indented on the same level (preferably the one of the second line).

skorch/tests/callbacks/test_early_stopping.py
```diff
@@ -0,0 +1,170 @@
...
+        net = net_clf_cls(
+            classifier_module,
+            callbacks=[
+                early_stopping_cb,
+            ],
+            max_epochs=max_epochs,
+        )
+        net.fit(*classifier_data)
+
+        stopped_run = (
```
benjamin-work: Again, one-liner is enough here.

Also, as above, the should know that `assert len(net.history) == patience + 1 < max_epochs`.

benjamin-work:
@tpietruszka I have taken the liberty to commit the changes I asked for and to also solve the outstanding merge conflicts. If you don't object, I will merge this PR as is.

tpietruszka:
Of course, thank you, sorry I could not do it myself

skorch/callbacks/training.py
```diff
@@ -101,3 +102,102 @@ def _sink(self, text, verbose):
...
+    threshold : int (default=1e-4)
+      Ignore score improvements smaller than `threshold`.
+
+    threshold_mode : str (default='rel')
+        One of `rel`, `abs`. Decides whether the `threshold` value is
+        interpreted in absolute terms or as a fraction of the best
+        score so far (relative)
+
+    sink : callable (default=print)
+      The target that the information about created checkpoints is
```
taketwo: There are no check-points here ;)

taketwo:
Suppose a user adds a pair of callbacks: a scoring callback to compute the score, and an early stopping callback to monitor this score. He will need to specify the `lower_is_better` parameter two times (and not make an accidental mistake).

I think `EarlyStopping` can figure out whether lower is better automatically, by checking two consecutive history entries and their "best" status. What do you think? Or we can not assume that each score will have a corresponding "best" column?

benjamin-work:
> I think EarlyStopping can figure out whether lower is better automatically, by checking two consecutive history entries and their "best" status.

I agree that in theory, indicating this twice should not be necessary. However, I'm not sure if automatic detection cannot have some unintended consequences in edge cases.

I would propose to leave the implementation as is for this PR. In a later PR, we could implement a `lower_is_better='auto'` option that will automatically determine what the right direction is. How is that?

taketwo:
Sounds good. (If implemented, `lower_is_better="auto"` should become the default I think.)

Issue #174, opened 2018-04-11
allow scoring callbacks to work with sklearn metrics


benjamin-work:
The scoring callbacks don't accept sklearn metrics as scoring function (e.g. `sklearn.metrics.accuracy`). Maybe we should make this possible by internally calling `sklearn.metrics.make_scorer` if we detect this (if that is possible).

This requires to resolve #173 first.

PR #216, opened 2018-05-10
ENH: Enable sklearn make_scorer

### LINKED ISSUE #173: scoring callbacks fail with sklearn make_scorer
Using a scoring callback with `make_scorer` fails:

```
from sklearn.metrics import accuracy_score, make_scorer

net = NeuralNetClassifier(..., callbacks=[('accuracy', BatchScoring(make_scorer(accuracy)))])
```
The reason is that the name of the `make_scorer` object cannot be determined. To mitigate this, pass the `name` argument to the scoring callback.

This bug is especially annoying as users might believe it is possible to pass an sklearn metric as scoring function, which fails with the prompt to use `make_scorer`, which will then fail with the error above.

thomasjpfan:
Fixes #173, #174

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -4,7 +4,8 @@ from contextlib import contextmanager
 from functools import partial
 
 import numpy as np
-from sklearn.metrics.scorer import check_scoring, _BaseScorer
+from sklearn.metrics.scorer import (
+    check_scoring, _BaseScorer, make_scorer)
 from sklearn.model_selection._validation import _score
 
 from skorch.utils import data_from_dataset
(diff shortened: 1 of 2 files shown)


skorch/callbacks/scoring.py
```diff
@@ -34,6 +35,19 @@ def cache_net_infer(net, use_caching, y_preds):
     del net.__dict__['infer']
 
 
+def convert_sklearn_metric_function(scoring):
+    """If ``scoring`` is a sklearn metric function, convert it to a
+    sklearn scorer and return it. Otherwise, return ``scoring`` unchanged."""
+    if callable(scoring):
+        module = getattr(scoring, '__module__', None)
+        if (hasattr(module, 'startswith') and
```
BenjaminBossan: Could you please change the indentation of the conditions to increase readability:

```
if (
        cond0 and
        cond1 and ...
):
    return ...
```

skorch/callbacks/scoring.py
```diff
@@ -64,10 +78,13 @@ def _get_name(self):
             return self.scoring
         if isinstance(self.scoring, partial):
             return self.scoring.func.__name__
+        if isinstance(self.scoring, _BaseScorer):
+            return self.scoring._score_func.__name__
         return self.scoring.__name__
 
     def initialize(self):
         self.best_score_ = np.inf if self.lower_is_better else -np.inf
+        self.scoring = convert_sklearn_metric_function(self.scoring)
```
BenjaminBossan: Could you change the new variable to `scoring_`, so that the original `scoring` remains untouched?

BenjaminBossan:
Thanks for the contribution. It looks very good to me, just 2 minor comments. I will be back in office on Monday and then give it a final review.

benjamin-work:
TY

Issue #218, opened 2018-05-13
Should use torch.no_grad in validation_step()?


taketwo:
I'm very new to both PyTorch and Skorch. I have started with the quickstart example and it worked fine, so I proceeded to implement an actual network that I care about (Pointnet). Unfortunately, `fit()` fails with "out of memory" error:

```
RuntimeError: cuda runtime error (2) : out of memory at /pytorch/aten/src/THC/generic/THCStorage.cu:58
```

I narrowed down the source to be the validation step. After each validation batch the GPU memory consumption increases until nothing in left. On the PyTorch issue tracker I found a [suggestion](https://github.com/pytorch/pytorch/issues/958#issuecomment-383625711) that validation should be run within `torch.no_grad()` context. Grepping through skorch codebase yields no matches. I tried wrapping the contents of [`validation_step()`](https://github.com/dnouri/skorch/blob/ec6903b5899a98eaefc38e774f6dce791c0adc95/skorch/net.py#L477-L479) into the `no_grad` context, and it solved my problem.

Do I overlook or misunderstand something? How does this work for others (on non-toy examples) without my fix?

BenjaminBossan:
Very good catch, we failed to change this when pytorch 0.4 was introduced. I made a PR (#219) to fix this.

BenjaminBossan:
`get_loss` currently doesn't use it in the skorch code base, but we encountered situations where it makes a difference.

taketwo:
Fixed with #219. What about releasing 0.2.1 already? I think this bug is critical enough.

benjamin-work:
There are still some pending changes that we would like to include before the next release. Meanwhile, you can install directly from [github](https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch) if you need the fix now.

PR #219, opened 2018-05-13
Bugfix/no grad during valid

BenjaminBossan:
Addresses #218

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -475,8 +475,9 @@ class NeuralNet(object):
 
         """
         self.module_.eval()
-        y_pred = self.infer(Xi, **fit_params)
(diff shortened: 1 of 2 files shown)


taketwo:
I've just noticed there is an `evaluation_step()` function. Perhaps this one should also be modified?

pranavdheram:
Hi! I am still having issues with my GPU running out of memory when I am changing to model.eval(). I see that we need to wrap the code around torch.no_grad() but I am not sure what this means. Any leads would be appreciated. Thanks

qlwang25:
@pranavdheram 
I think it is right:
```
with torch.no_grad():
    evaluate_code(arg)
```

PR #225, opened 2018-05-20
Default train/valid loss callbacks should have no-op target extractor

taketwo:
Both of these callbacks retrieve already computed losses from the history, thus they do not need target and they do not need to process it in any way.

I care about this case for the following reason. I am implementing a multi-task network which has several outputs. The target values are actually dictionaries with the ground truth output for these several tasks. Naturally, I implemented a custom loss function which takes care of this. Now I expect the default scoring to be able to report training loss (because it was computed already). Right now it is not possible, because the default `target_extractor` is `to_numpy()`, which does not work with dictionaries (which my `y`s are).

In future the lambda can be replaced with Python [`noop()`](https://www.python.org/dev/peps/pep-0559/) built-in.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -24,9 +24,10 @@ from skorch.history import History
 from skorch.utils import duplicate_items
 from skorch.utils import get_dim
 from skorch.utils import is_dataset
+from skorch.utils import noop
+from skorch.utils import params_for
 from skorch.utils import to_numpy
 from skorch.utils import to_tensor
-from skorch.utils import params_for
 
 
 # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
This looks like a reasonable change. Could you please also change the defaults in `NeuralNetClassifier` as well?

https://github.com/dnouri/skorch/blob/735e02296f82c0c607d5905dd4882133d9143282/skorch/net.py#L1397

Furthermore, please change the `lambda`s to actual functions (e.g. forward port noop) so that the callbacks can be pickled.

taketwo:
> PS: It appears that the noop PEP was rejected.

Indeed, didn't pay attention to the status.

Where would you like our `noop()` to be placed? Right there in `net.py`? Should it be `noop` or `_noop`?

BenjaminBossan:
I think `noop` in `net.py` works fine. It will need a `# pylint: disable=unused-argument`.

benjamin-work:
LGTM, let's see what @ottonemo thinks.

taketwo:
Moved to "utils.py" and added a docstring.

Issue #177, opened 2018-04-11
Complications with placeholder value in skorch Dataset


ottonemo:
See https://github.com/dnouri/skorch/blob/d21c96489f592e8fd70f5f4f2153cc3cd806c83a/skorch/dataset.py#L128-L132

PR #226, opened 2018-05-20
ENH: Handle placeholder in BatchScoring

thomasjpfan:
Addresses https://github.com/dnouri/skorch/issues/177

This uses the same technique as `EpochScoring` to deal with placeholder `y`. The user would have to provide a scoring function that takes `net` as a `NeuralNet`, `X` as a tensor and `y=None`.


BenjaminBossan:
Looks good. @ottonemo what do you think?

ottonemo:
Thanks for the PR. This fixes a problem with `BatchScoring` so in that respect the PR LGTM.

But: the PR does not address the issue that the dummy value is confusing for callbacks that operate on batch level so the FIXME is still valid in my opinion. I would keep #177 open and leave the FIXME in place.

@benjamin-work what do you think?

thomasjpfan:
Another solution would be to add `y_is_placeholder` to the `on_batch_begin` and `on_batch_end` during the fit_loop:

https://github.com/dnouri/skorch/blob/dd3318c961caef9d3e12fdb09d76043c8863f637/skorch/net.py#L586-L591

Line 586 becomes `self.notify('on_batch_begin', X=Xi, y=yi, training=True, y_is_placeholder=True)`,
and Line 591 becomes `self.notify('on_batch_end', X=Xi, y=yi, training=True,  y_is_placeholder=True, **step)`.

thomasjpfan:
I have updated this PR by moving `y_is_placeholder` to `fit_loop`. Any callback listening to `on_batch_begin` and `on_batch_end` can use the `y_is_placeholder` keyword to deal with placeholders.

benjamin-work:
I have thought a little bit about this and have to say that I like the first implementation better. The reason is that currently, the `fit_loop` looks very clean -- I believe most users will understand what happens there after a quick glance, there is little distracting noise. With the 2nd implementation, we suddenly have code in there that deals with a fringe case but takes up a disproportionate number of lines. If I wanted to override this method, the number of things I have to keep in mind becomes ever larger.

tldr: I like implementation 1 better because it keeps the `fit_loop` cleaner.

ottonemo:
I mean we could do something like `notify('on_batch_begin', y=self.resolve_placeholder(dataset, y))` where `resolve_placeholder` returns `None` if it determines that `y` is a placeholder value.

benjamin-work:
Does `resolve_placeholder` need to be a method on the net?

thomasjpfan:
I just rebased this PR with the master branch. The idea of resolving `yi` to `None` is good. I think calling `resolve_placeholder` on every batch may lead to lower performance by calling `isinstance` multiple times.

I added `dataset_uses_y_placeholder` to `utils.py` that returns True if the dataset is using a y placeholder. Before the for loop, `dataset_uses_y_placeholder` is called on the train and valid datasets. In the loop, the `yi` is resolved to None when a placeholder is used.

benjamin-work:
@ottonemo are you okay with a merge?

PR #227, opened 2018-05-21
ENH: batch_size=-1 uses whole dataset

### LINKED ISSUE #208: Option to disable batching (e.g. batch_size=None)
We should think about introducing an option to disable batching, e.g. by setting `batch_size=None`.
This could be implemented by either setting the `batch_size` to `len(dataset)` in that case or 
explicitly setting `batch_sampler=SequentialSampler` in the data loader.

A possible use case is the usage of LBFGS, see #207.

thomasjpfan:
Resolves #208

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -102,7 +102,9 @@ class NeuralNet(object):
     batch_size : int (default=128)
       Mini-batch size. Use this instead of setting
       ``iterator_train__batch_size`` and ``iterator_test__batch_size``,
-      which would result in the same outcome.
+      which would result in the same outcome. If ``batch_size`` is None,
+      a single batch with all the data will be used during training
+      and validation.
 
     iterator_train : torch DataLoader
       The default PyTorch :class:`~torch.utils.data.DataLoader` used for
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thank you very much for adding this feature.

I believe the case of specifically setting `iterator_train(/valid)__batch_size=None` is not covered by this, is it?

Additionally, I wonder whether it would be a good addition to allow setting `batch_size=-1` with the same effect -- the reason being that in sklearn, `-1` is often used as a placeholder to mean "all" (e.g. `n_jobs=-1` means use all cores).

thomasjpfan:
Currently `iterator_train(/valid)__batch_size=None` sets the batch_size to `self.batch_size`. This behavior should be kept for backwards compatibility. 

I think your `batch_size=-1` idea for "all" is a superior interface. I have updated this PR to use `iterator_train(/valid)__batch_size=-1` and `batch_size=-1` instead of `None`.

benjamin-work:
I like this solution. My first thought was to have `batch_size=-1` and `batch_size=None` to have the same effect, but that would be more confusing.

Overall, I'm ready to merge. I just wonder whether there is a way to get rid of some of the repetition in the unit tests (maybe there is no good way, then it's fine).

thomasjpfan:
I updated this PR by using `pytest.mark.parametrize` to get rid of some repetition.

PR #230, opened 2018-05-23
ENH: Adds history loading and saving

thomasjpfan:
Adds saving and loading history from a `NeutralNet`.

This feature can be used by:

```python
before = NeuralNetClassifier(mymodule)
before.save_history('history.json')
before.fit(X, y)
after = NeuralNetClassifier(mymodule).initialize()
after.load_history('history.json')
```

Use Cases:

1. A user wants to continue training a `NeutralNet` while keeping the history of the previous run. This allows for `LRScheduler` to work correctly during the second run, since `LRScheduler` relies on `len(history)`.
2. After training is finished, the `history.json` file can be used to analyze the run.

### DIFF
--- a/docs/user/neuralnet.rst
+++ b/docs/user/neuralnet.rst
@@ -419,9 +419,10 @@ initialize a :class:`.NeuralNet` to load the parameters again:
     new_net.initialize()  # This is important!
     new_net.load_params('some-file.pkl')
 
-In addition to saving the model, the history can persist by
-calling :func:`~skorch.net.NeuralNet.save_history` and
-:func:`~skorch.net.NeuralNet.load_history` methods on
+In addition to saving the model parameters, the history
+can be saved and loaded by calling the
+:func:`~skorch.net.NeuralNet.save_history`
+and :func:`~skorch.net.NeuralNet.load_history` methods on
 :class:`.NeuralNet`. This feature can be used to
 continue training:
 
@@ -529,5 +530,5 @@ total loss:
             loss += self.lambda1 * sum([w.abs().sum() for w in self.module_.parameters()])
             return loss
 
-.. note:: This example also reguralizes the biases, which you typically
+.. note:: This example also regularizes the biases, which you typically
     don't need to do.
--- a/docs/user/neuralnet.rst
+++ b/docs/user/neuralnet.rst
@@ -419,6 +419,37 @@ initialize a :class:`.NeuralNet` to load the parameters again:
     new_net.initialize()  # This is important!
     new_net.load_params('some-file.pkl')
 
+In addition to saving the model, the history can persist by
+calling :func:`~skorch.net.NeuralNet.save_history` and
+:func:`~skorch.net.NeuralNet.load_history` methods on
+:class:`.NeuralNet`. This feature can be used to
+continue training:
+
+.. code:: python
+
+    net = NeuralNet(
+        module=MyModule
+        criterion=torch.nn.NLLLoss,
+    )
+
+    net.fit(X, y, epochs=2) # Train for 2 epochs
+
+    net.save_params('some-file.pkl')
+    net.save_history('history.json')
+
+    new_net = NeuralNet(
+        module=MyModule
+        criterion=torch.nn.NLLLoss,
+    )
+    new_net.initialize() # This is important!
+    new_net.load_params('some-file.pkl')
+    new_net.save_history('history.json')
+
+    new_net.fit(X, y, epochs=2) # Train for another 2 epochs
+
+.. note:: In order to use this feature, the history
+    must only contain JSON encodable Python data structures.
+    Numpy and PyTorch types should not be in the history.
 
 Special arguments
 -----------------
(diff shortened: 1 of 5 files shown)


skorch/utils.py
```diff
@@ -5,14 +5,18 @@
...
 from enum import Enum
 from functools import partial
+import sys
 
 import numpy as np
 from sklearn.utils import safe_indexing
 import torch
 from torch import nn
 from torch.utils.data.dataset import Subset
+if sys.version_info[0] == 3:
```
benjamin-work: Not technically necessary, since we don't support Python 2

skorch/utils.py
```diff
@@ -274,3 +278,18 @@ def noop(*args, **kwargs):
...
     """
     pass
+
+
+@contextmanager
+def _with_file_like(f, mode):
+    new_fd = False
+    if (
+        isinstance(f, str) or
+        sys.version_info[0] == 2 and isinstance(f, unicode) or
```
benjamin-work: Again, we don't support Python 2

skorch/utils.py
```diff
@@ -274,3 +278,18 @@ def noop(*args, **kwargs):
...
+        sys.version_info[0] == 3 and isinstance(f, pathlib.Path)
+    ):
+        new_fd = True
+        f = open(f, mode)
+    yield f
+    if new_fd:
+        f.close()
```
benjamin-work: Do we need a `try ... finally` here in case we open the file but something goes wrong?

skorch/net.py
```diff
@@ -1325,6 +1327,44 @@ def load_params(self, f):
...
+        ----------
+        f : file-like object or str
+
+        Examples
+        --------
+        >>> before = NeuralNetClassifier(mymodule)
+        >>> before.save_history('path/to/file.json')
+        >>> before.fit(X, y)
+        >>> after = NeuralNetClassifier(mymodule).initialize()
+        >>> after.load_history('path/to/file.json')
```
ottonemo: I don't think copying the example from above is a good idea. Maybe reference the example from `save_history`?

skorch/net.py
```diff
@@ -1325,6 +1327,44 @@ def load_params(self, f):
...
```
ottonemo: This example is broken or at least misleading: if `mymodule` is a class then `after` is a freshly initialized net/module with the history of a trained one which will surely result in unexpected behavior. I think it would be better to show this functionality with `net.load`/`net.save`.

skorch/utils.py
thomasjpfan: If there is an error opening the file, I think the caller should handle the exception.

thomasjpfan:
@benjamin-work When a non JSON serializable object is passed to `json.dump`, it will raise a `TypeError` with a message similar to: "Object of type 'ndarray' is not JSON serializable". We can adjust this message, but I think the message is enough to tell the user what the issue is.

benjamin-work:
Yes, the message is already quite good. It would, however, be nice to at least mention somewhere that if the user wants to save the history, they should avoid numpy or torch types. Maybe there should be a mention of this (and generally about this new feature) in the docs/history section.

thomasjpfan:
I updated this PR with documentation in the **saving and loading** section of docs/user/neuralnet.rst.

docs/user/neuralnet.rst
```diff
@@ -419,6 +419,37 @@ initialize a :class:`.NeuralNet` to load the parameters again:
     new_net.initialize()  # This is important!
     new_net.load_params('some-file.pkl')
 
+In addition to saving the model, the history can persist by
```
benjamin-work: > the history can persist by

Should be "the history can be persisted and loaded by".

docs/user/neuralnet.rst
```diff
@@ -498,5 +529,5 @@ total loss:
             loss += self.lambda1 * sum([w.abs().sum() for w in self.module_.parameters()])
             return loss
 
-*Note*: This example also reguralizes the biases, which you typically
- don't need to do.
+.. note:: This example also reguralizes the biases, which you typically
```
benjamin-work: While you're at it, could you please fix the "reguralizes" typo? (That probably was me)

skorch/utils.py
```diff
@@ -274,3 +276,16 @@ def noop(*args, **kwargs):
     target extractor.
     """
     pass
+
+
+@contextmanager
+def open_file_like(f, mode):
+    """Wrapper for opening a file and y"""
```
benjamin-work: > and y

That is not clear to me.

thomasjpfan: That is a typo

Atry:
Is there a Callback to call `save_history` in each epoch?

taketwo:
In #246 we reached a tentative agreement that saving history should be added to the `Checkpoint` callback. I have it on my TODO, but haven't had time to discuss the API and implement yet.

thomasjpfan:
@Atry In the mean time, the following callback would work for saving history per epoch:

```python
class HistorySaver():
    def __init__(self, target='history.json'):
        self.target = target

        net.save_history(self.target)
```

PR #231, opened 2018-05-24
BUG: Fixes off by one batch index

thomasjpfan:
The batch index should be zero indexed. Consider the following learning rate evolution of `CyclicLR`:

screen shot

This PR moves the starting point from the red dot to the gray dot (on the bottom left).


benjamin-work:
Looks good, thank you.

PR #232, opened 2018-05-25
API: Removes device casting from Dataset

thomasjpfan:
Resolves https://github.com/dnouri/skorch/issues/223.

1. Removes `to_tensor` and `self.device` from `Dataset`.

2. `torch.utils.data.DataLoader`'s `collate_fn=default_collate` function is able to handle our types without needing to covert them to tensors first. During training, validation, or evaluation, `NeuralNet.get_loss` and `NeuralNet.infer` will call `to_tensor` to move data to `self.device`.

3. When `X` and `y` are yielded from a `DataLoader`, their numerical values are already converted into tensors. `get_loss` and `infer` will call `to_tensor` with tensor-filled `X` and `y` values.  With this in mind, I refactored `to_tensor` to check the tensor type first and return early.

This PR frees `Dataset` from worrying about `device` and allows `NeutralNet` to handle device transfers. This makes `Dataset` be inline with datasets in `torchvision.datasets`, which doesn't involve the GPU when yielding data.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -15,7 +15,6 @@ from skorch.utils import flatten
 from skorch.utils import is_pandas_ndframe
 from skorch.utils import multi_indexing
 from skorch.utils import to_numpy
-from skorch.utils import to_tensor
 
 
 def _apply_to_data(data, func, unpack_dict=False):
@@ -73,9 +72,6 @@ class Dataset(torch.utils.data.Dataset):
     y : see above or None (default=None)
       Everything pertaining to the target, if there is anything.
 
-    device : str, torch.device (default='cpu')
-      Which computation device to use (e.g., 'cuda').
-
     length : int or None (default=None)
       If not None, determines the length (``len``) of the data. Should
       usually be left at None, in which case the length is determined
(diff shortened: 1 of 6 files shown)


skorch/dataset.py
```diff
@@ -86,12 +82,10 @@ def __init__(
             self,
             X,
             y=None,
-            device='cpu',
```
benjamin-work: Could you please leave this here for now and down below raise a deprecation warning that "device is no longer needed by Dataset and will be ignored." with a comment to remove the warning in a later release.

benjamin-work:
Excellent changes, thank you. My only wish is to add a deprecation warning for the removal of `device` in `Dataset`.

@ottonemo pls also have a look.

ottonemo:
@thomasjpfan Can you add a deprecation message. Then we can merge this PR.

thomasjpfan:
I updated this PR with a deprecation warning.

BenjaminBossan:
Thanks, again, for your excellent contribution.

@ottonemo can you merge?

benjamin-work:
Thanks again.

Issue #224, opened 2018-05-19
Benchmark skorch overhead


BenjaminBossan:
[good first issue] [enhancement]

We should have a benchmark script that measures the skorch performance overhead.

Ideally, it should be run on a *real* dataset but still be reasonably quick (say MNIST) so that we could possibly include it in CI. This script should ascertain that skorch does not slow down training and inference too much, and that approximately the same loss is reached.

PR #234, opened 2018-05-25
mnist benchmark

benjamin-work:
addresses #224


benjamin-work:
@ottonemo ready for review

PR #235, opened 2018-05-26
fixing torch.nn.utils.clip_grad Import error

8key:
it raised an import error, the code fixed for my local


8key:
was __init__.py issue, so not an issue anymore

PR #239, opened 2018-06-01
ENH: Adds NeuralNetBinaryClassifier

### LINKED ISSUE #200: Provide NeuralNetBinaryClassifier
While `NeuralNetClassifier` supports binary classification when formulated as a multi-class classification problem with two classes (`shape == (n, 2)`), it does not support modeling the prediction as vector (`shape == (n,)`) and using a threshold (i.e. 0.5) to determine the class (see also: #137).

Since this way of modeling binary classification differs from multi-class in several ways we will not change the existing class to support binary classification but will introduce a new class that can specifically handle all intricacies that may arise in this use-case.

thomasjpfan:
Resolves #200. I made the following design choices when creating `NeuralNetBinaryClassifier`:

1. Uses `BCEWithLogitsLoss` by default, which means `self.module` must output the logit of probability. Based on the pytorch documentation, `BCEWithLogitsLoss` is more numerical stable to use than `BCELoss`.

2. `self.module` is expected to output values with shape: (`batch_size`, ). I am open to changing this to (`batch_size`, 1).

3. Subclassed `NeuralNetClassifier` since the `predict`, `predict_proba`, and `fit` docstrings are relevant for  `NeuralNetBinaryClassifier`. 

4. `NeuralNetClassifier.check_data` is reused, combined with a check on the y's dimension.

5. I had to override `get_loss` and use `NeuralNet.get_loss` instead. In its current form, `NeuralNetBinaryClassifier` should not need to take the log of any values. This is the downside of subclassing `NeuralNetClassifier`.

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -202,7 +202,11 @@ neural_net_binary_clf_criterion_text = """
 
     criterion : torch criterion (class, default=torch.nn.BCEWithLogitsLoss)
       Binary cross entropy loss with logits. Note that the module should return
-      the logit of probabilities with shape (batch_size, )."""
+      the logit of probabilities with shape (batch_size, ).
+
+    threshold : float (default=0.5)
+      Probabilities above this threshold is classified as 1. ``threshold``
+      is used by ``predict`` and ``predict_proba`` for classification."""
 
 
 def get_neural_net_binary_clf_doc(doc):
@@ -222,6 +226,7 @@ class NeuralNetBinaryClassifier(NeuralNet):
             module,
             criterion=torch.nn.BCEWithLogitsLoss,
             train_split=CVSplit(5, stratified=True),
+            threshold=0.5,
             *args,
             **kwargs
     ):
@@ -232,6 +237,7 @@ class NeuralNetBinaryClassifier(NeuralNet):
             *args,
             **kwargs
         )
+        self.threshold = threshold
 
     @property
     def _default_callbacks(self):
@@ -277,7 +283,7 @@ class NeuralNetBinaryClassifier(NeuralNet):
         # https://github.com/PyCQA/pylint/issues/1085
         return super().fit(X, y, **fit_params)
 
-    def predict(self, X, threshold=0.5):
+    def predict(self, X):
         """Where applicable, return class labels for samples in X.
 
         If the module's forward method returns multiple outputs as a
@@ -306,7 +312,7 @@ class NeuralNetBinaryClassifier(NeuralNet):
         y_pred : numpy ndarray
 
         """
-        return (self.predict_proba(X) > threshold).astype('uint8')
+        return (self.predict_proba(X) > self.threshold).astype('uint8')
 
     # pylint: disable=missing-docstring
     def predict_proba(self, X):
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -224,10 +224,10 @@ class NeuralNetBinaryClassifier(NeuralNet):
     def __init__(
             self,
             module,
+            *args,
             criterion=torch.nn.BCEWithLogitsLoss,
             train_split=CVSplit(5, stratified=True),
             threshold=0.5,
-            *args,
             **kwargs
     ):
         super().__init__(
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -16,7 +16,6 @@ from skorch.utils import get_dim
 from skorch.utils import is_dataset
 from skorch.utils import noop
 from skorch.utils import to_numpy
-from skorch.utils import to_tensor
 from skorch.utils import train_loss_score
 from skorch.utils import valid_loss_score
 
@@ -215,7 +214,7 @@ def get_neural_net_binary_clf_doc(doc):
 
 
 # pylint: disable=missing-docstring
-class NeuralNetBinaryClassifier(NeuralNetClassifier):
+class NeuralNetBinaryClassifier(NeuralNet):
     __doc__ = get_neural_net_binary_clf_doc(NeuralNet.__doc__)
 
     def __init__(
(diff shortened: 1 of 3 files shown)


benjamin-work:
Thank you for your contribution.

Some general comments:

> Uses BCEWithLogitsLoss by default

Makes sense. Users can change this if they like.

> self.module is expected to output values with shape: (batch_size, )

I wonder whether we could accept both and squeeze on the other end if necessary.

> Subclassed NeuralNetClassifier

I'd rather have that all net classes inherit from `NeuralNet`. There will be some code duplication, but if it mostly affects docstrings, I'm fine with it. If it affects actual methods, we can think about using mixins. The disadvantage of subclassing `NeuralNetClassifier` is that it can get tedious and confusing to find out where methods actually come from. Flatter hierarchies should keep things simple (or at least simple-ish).

thomasjpfan:
I agree with delaying this PR for #240. `net.py` is getting pretty long.

benjamin-work:
With #240, you could now move your code to `classifier.py`.

I saw that you hardcoded the threshold to 0.5. Do you think that making this a parameter would be useful?

thomasjpfan:
This PR has been updated:

1. Adds `NeuralNetBinaryClassifier` to `classifier.py`.
2. `NeuralNetBinaryClassifier` subclasses `NeutralNet`.
3. Adds `threshold` to `predict`: `predict(self, X, threshold=0.5)`.

skorch/classifier.py
```diff
@@ -188,3 +189,160 @@ def predict(self, X):
...
+        set to ``None`` in case it is derived dynamically from
+        ``X``.
+
+        """
+        # pylint: disable=useless-super-delegation
+        # this is actually a pylint bug:
+        # https://github.com/PyCQA/pylint/issues/1085
+        return super().fit(X, y, **fit_params)
+
+    def predict(self, X, threshold=0.5):
```
ottonemo: I think this needs to be a class attribute since passing kwargs in predict is enerally unsupported by sklearn. This would not work in a pipeline for example.

skorch/classifier.py
```diff
@@ -188,3 +189,160 @@ def predict(self, X):
...
+
+        Returns
+        -------
+        y_proba : numpy ndarray
+
+        """
+        y_probas = []
+        for yp in self.forward_iter(X, training=False):
+            yp = yp[0] if isinstance(yp, tuple) else yp
+            yp = torch.nn.functional.sigmoid(yp)
```
ottonemo: As with the multi-class classifier the module should return what the user wants to see in `predict_proba`. Consequently, I don't think that the call to sigmoid is justified here. Maybe it would be better to expect values in the range from 0 to 1 from the module.

thomasjpfan: I decided to use `BCEWithLogitsLoss` as the loss function, which expects the module to output the logit of probability with a range of (-inf, inf). I choose this loss function because it is more numerically stable than `BCELoss`.

Since logit of probability is the module output, sigmoid will convert it to a probability.

benjamin-work: I see why you made that decision but wonder if we need a check for the type of loss function used like in [NeuralNetClassifier](https://github.com/dnouri/skorch/blob/12842307309765b8a0caefa00441df88ec53ce0d/skorch/classifier.py#L103).

skorch/tests/test_classifier.py
```diff
@@ -117,3 +117,94 @@ def test_takes_no_log_without_nllloss(self, net_cls, module_cls, data):
...
+            assert expected_keys.issubset(row)
+
+    @pytest.mark.parametrize('threshold', [0, 0.25, 0.5, 0.75, 1])
+    def test_predict_predict_proba(self, net, data, threshold):
+        X, y = data
+        net.threshold = threshold
+        net.fit(X, y)
+
+        y_pred_proba = net.predict_proba(X)
+        assert len(y_pred_proba.shape) == 1
```
benjamin-work: Shorter: `assert y_pred_proba.ndim == 1`

skorch/classifier.py
thomasjpfan: `NeutralNetClassifier` assumes that the module output is probabilities and `NLLLoss` expects log probabilities, thus that check and log call needs to be in place.

`NeuralNetBinaryClassifier` assumes that the module outputs logits, and `BCEWithLogitsLoss` expects logits, thus no checks need to be in place.

I see how this can confuse a user, where `NeutralNetClassifier` and `NeuralNetBinaryClassifier` expects different outputs from the module. 

It is a choice that needs to be made:

1. Have both `NeutralNet*Classifier`s expect the module to output probabilities for a cleaner API, at the cost of `BCEWithLogitsLoss` using `BCELoss`.
2. Have `NeutralNetClassifier` expect the module to output probabilities, and `NeuralNetBinaryClassifier` expect logits. This allows `NeuralNetBinaryClassifier` to use `BCEWithLogitsLoss` to be numerically stable. The cost would be a less clean API.

thomasjpfan: Ah I see what you mean. I updated this PR with your suggestion.

benjamin-work:
Could you please add a test for the case that `BCEWithLogitsLoss` is not used? Then we should be fine.

thomasjpfan:
Added tests to check that `torch.nn.functional.sigmoid` is not called for criterion that is not `BCEWithLogitsLoss`.

skorch/tests/test_classifier.py
```diff
@@ -208,3 +208,27 @@ def test_target_2d_raises(self, net, data):
...
+        monkeypatch.setattr(torch.nn.functional, "sigmoid", mock)
+
+        net = net_cls(module_cls, max_epochs=20, lr=0.1, criterion=nn.MSELoss)
+        X, y = data
+        net.fit(X, y)
+
+        net.predict_proba(X)
+        mock.assert_not_called()
+
+    def test_default_loss_does_calls_sigmoid(
```
benjamin-work: `does_calls` -> `does_call`

skorch/tests/test_classifier.py
```diff
@@ -208,3 +208,27 @@ def test_target_2d_raises(self, net, data):
...
+            self, net_cls, data, module_cls, monkeypatch):
+        mock = Mock(side_effect=lambda x: x)
+        monkeypatch.setattr(torch.nn.functional, "sigmoid", mock)
+
+        net = net_cls(module_cls, max_epochs=20, lr=0.1, criterion=nn.MSELoss)
+        X, y = data
+        net.fit(X, y)
+
+        net.predict_proba(X)
+        mock.assert_not_called()
```
benjamin-work: Minor pet peeve of mine: I don't like methods on mock such as `assert_not_called`. The reason is that if there is a typo (e.g. `mock.asert_not_called()`), the test just passes. For this particular method, I like `mock.call_count == 1` better for the given reason.

skorch/tests/test_classifier.py
```diff
@@ -208,3 +208,27 @@ def test_target_2d_raises(self, net, data):
 
         assert exc.value.args[0] == (
             "The target data should be 1-dimensional.")
+
+    def test_custom_loss_does_not_call_sigmoid(
+            self, net_cls, data, module_cls, monkeypatch):
+        mock = Mock(side_effect=lambda x: x)
+        monkeypatch.setattr(torch.nn.functional, "sigmoid", mock)
+
+        net = net_cls(module_cls, max_epochs=20, lr=0.1, criterion=nn.MSELoss)
```
benjamin-work: Do we need 20 epochs here? Probably not a huge loss of time, but there seems to be no benefit.

skorch/tests/test_classifier.py
```diff
@@ -208,3 +208,27 @@ def test_target_2d_raises(self, net, data):
...
+            self, net_cls, data, module_cls, monkeypatch):
+        mock = Mock(side_effect=lambda x: x)
+        monkeypatch.setattr(torch.nn.functional, "sigmoid", mock)
+
+        net = net_cls(module_cls, max_epochs=20, lr=0.1)
+        X, y = data
+        net.fit(X, y)
+
+        net.predict_proba(X)
+        mock.assert_called()
```
benjamin-work: See comment above.

PR #247, opened 2018-06-11
Tpietruszka feature/support optimizers requiring closure arguments

benjamin-work:
I didn't manage to push on the existing PR, so here is a new one -- sorry for that.

### DIFF
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -1427,23 +1427,18 @@ class TestNeuralNet:
 
     def test_accumulator_that_returns_last_value(
             self, net_cls, module_cls, data):
-        # We define a loss that is called 3 times and an accumulator
-        # that returns the last of those calls. We then test that the
-        # correct values were stored.
+        # We define an optimizer that calls the step function 3 times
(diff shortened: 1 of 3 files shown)


skorch/tests/test_net.py
```diff
@@ -1409,6 +1419,70 @@ def test_batch_size_neg_1_uses_whole_dataset(
...
+            """Call NLLLoss 3 times."""
+            loss = torch.nn.NLLLoss()
+            def __call__(self, *args, **kwargs):
+                loss = self.loss(*args, **kwargs)
+                side_effect.append(float(loss))
+                loss = self.loss(*args, **kwargs)
+                side_effect.append(float(loss))
+                loss = self.loss(*args, **kwargs)
+                side_effect.append(float(loss))
+                return loss
```
ottonemo: This should be a custom optimizer instead of a criterion as the additional loss values do not contribute to the test.

benjamin-work:
@ottonemo pls review again

benjamin-work:
great work @tpietruszka @ottonemo I believe the long discussion was worth it at the end.

PR #248, opened 2018-06-12
Remove superfluous data call

ottonemo:
This is most probably(tm) not needed anymore.


benjamin-work:
It seems the call was unnecessary. I also found one in mnist.py, which I removed as well.

Issue #240, opened 2018-06-01
Clean up modules:


benjamin-work:
* split net.py, since it's too long (e.g. into base.py and net.py)
* the callback test modules are currently all over the place

benjamin-work:
Proposals for `net.py`:

* Split `net.py` into 2, one for the base and one for the derived classes. Makes more sense when the number of derived classes is low (classifier, binary classifier, regressor), is less consistent.
* Create a subfolder, analogously to callbacks, with a `base.py`, `classifier.py`, etc. Makes more sense when the number of derived classes grows larger, and is more consistent, but perhaps overkill.

Regarding the callback tests, they should mirror the module structure of callbacks.

ottonemo:
Done with #249 and #251 merged.

PR #249, opened 2018-06-12
RFC: Splits callback tests into seperate files

thomasjpfan:
Refactors callback tests as suggested in #240.


benjamin-work:
Thanks, I will look into this.

Just so that you know, I already started working on the other part (splitting net.py).

benjamin-work:
We saw that you moved the tests for some classes to their own submodule (e.g. `test_print_log.py`). Could you instead mirror the module structure of `callbacks` for consistency?

At least for now, we want to keep the current structure. If in the future, we find that files become too crowded, we might change that, but for now, the existing structure looks good enough.

thomasjpfan:
I updated this PR to mirror the `callbacks` structure.

benjamin-work:
Looks good to me. Only the error is left to fix. I guess changing `from .conftest import` to `from ..conftest import` and puting an `__init__.py` into `skorch/tests/callbacks/` should do the job.

thomasjpfan:
I refactored the two tests to use the pytest fixtures directly. This fixes the tests that was failing before.

benjamin-work:
Yes, that's the better way.

Many thanks.

Issue #246, opened 2018-06-06
Checkpoint callback verbosity


taketwo:
Currently, the `Checkpoint` callback prints a message to terminal every time the model is saved. These messages get mixed with the `PrintLog` output, breaking the neat progress log table. One can set `net.verbosity=False`, but this disables output from both callbacks.

What's your opinion on adding a constructor argument to `Checkpoint` to control its verbosity?

One other idea I had is to let `Checkpoint` write whether the model was saved or not to the network history (say, under 'checkpoint' key). `PrintLog` will then have a separate column indicating that. Even fancier, it can analyze these keys (the same way it automatically analyzes '_best' keys) and highlight epoch number (with bold or some color) if checkpoint occurred.

**Edit:**  while we are at this, what about adding an option to save the history? Useful for resuming training later.

benjamin-work:
Yes, there should definitely be a way to control this.

> What's your opinion on adding a constructor argument to Checkpoint to control its verbosity?

We've had our discussions about how to control verbosity in the past. I don't like this suggestion because now we suddenly have verbosity parameters at several places. Who is really in control?

One possibility (which would be consistent with sklearn) is to have several verbosity levels and changing `Checkpoint` so that it only prints if `verbosity >= 2`. This solution is, however, very opaque. What if only want the `Checkpoint` message but not `PrintLog`?

Another solution is to have `Checkpoint` have a `sink` argument that controls where the print message goes to. By default, we would have `sink=print`, but the user may change it to `sink=logger.info` or `sink=noop` (which would suppress the messages).

This would be an indirect way to address the original concern, but it has some advantages. Most notably, it gives the user control about what should happen to the message, but it also makes testing easier (no more `capsys` necessary).

Of course, if we made that change, it should affect all instances where we currently use print.

taketwo:
You second solution with `sink` sounds reasonable. Is this something you want/have time to implement yourself, or should I give it a try?

Regarding history saving, I am aware of that new feature. I was proposing to make use of it (by adding an option to `Checkpoint` to control whether it should save history as well).

benjamin-work:
> You second solution with sink sounds reasonable. Is this something you want/have time to implement yourself, or should I give it a try?

We are always happy to get some help. But first I would would for @ottonemo's input regarding this idea.

> Regarding history saving, I am aware of that new feature. I was proposing to make use of it (by adding an option to Checkpoint to control whether it should save history as well).

I see, this is probably a sensible addition.

[...]

PR #252, opened 2018-06-14
Satisfy pylint

benjamin-work:
* non-functional changes for pylint
* allow attribute names of type "X_*"
* allow line lengths up to 88


taketwo:
I finally got around to start implementing the things discussed in #246. What line length should I format for? In general, how do you format code? Yapf?

BenjaminBossan:
We don't currently have strict rules in place but if you go with [black](https://github.com/ambv/black) you probably won't go wrong. There might be some minor things I disagree with, but overall it looks very reasonable.

PR #253, opened 2018-06-14
Add support for configurable sink in Checkpoint

taketwo:
Partially addresses #246.

I'm not particularly confident with unit testing in Python, so I probably did something wrong, please let me know.

### DIFF
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -56,14 +56,21 @@ class Checkpoint(Callback):
       **Note:** If you supply a lambda expression as monitor, you cannot
       pickle the wrapper anymore as lambdas cannot be pickled. You can
       mitigate this problem by using importable functions instead.
+
+    sink : callable (default=print)
+      The target that the information about created checkpoints is
+      sent to. By default, the output is printed to stdout, but the
+      sink could also be a logger or :func:`~skorch.utils.noop`.
     """
     def __init__(
             self,
             target='model.pt',
             monitor='valid_loss_best',
+            sink=print,
     ):
         self.monitor = monitor
         self.target = target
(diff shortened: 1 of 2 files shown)


skorch/callbacks/training.py
```diff
@@ -87,6 +94,9 @@ def on_epoch_end(self, net, **kwargs):
                     last_epoch=net.history[-1],
                     last_batch=net.history[-1, 'batches', -1],
                 )
-            if net.verbose > 0:
-                print("Checkpoint! Saving model to {}.".format(target))
+            self._sink("Checkpoint! Saving model to {}.".format(target), net.verbose)
             net.save_params(target)
+
+    def _sink(self, text, verbose):
+        if (self.sink is not print) or verbose:
```
ottonemo: Why check if the sink is not print? I would only expect a verbosity check here. Also I woild move the reference of net.verbose to this method since it is unlikely to be something different.

taketwo: The check was directly copy/pasted from `PrintLog`. I don't quite understand why, but I assumed there was some clever reason.

Regarding moving the reference, `net` is not part of self, so it's not accessible in `_sink`. Or do I misunderstand your proposal?

BenjaminBossan: The idea here is that if sink is not print, you don't want it to be affected by verbosity.

ottonemo: Ah, my bad. This is probably fine then. But since 2 out of 3 missed the purpose, adding an inline comment might be appropriate.

You are right about `net.verbose` of course.

taketwo:
Ping

ottonemo:
LGTM

PR #254, opened 2018-06-15
DOC: Adds neuralnets

thomasjpfan:
1. Adds `NeuralNetClassifier` and `NeuralNetRegressor` into docs.
2. Fixes formating in docstring:

**Before correction:**
screen shot 2018-06-15 at 1 20 53 am

**After correction:**
screen shot 2018-06-15 at 1 22 44 am

### DIFF
--- a/docs/classifier.rst
+++ b/docs/classifier.rst
@@ -1,0 +1,5 @@
+skorch.classifier
+=================
+
+.. automodule:: skorch.classifier
+    :members:
--- a/docs/neuralnetclassifier.rst
+++ b/docs/neuralnetclassifier.rst
@@ -1,5 +1,0 @@
-skorch.NeuralNetClassifier
-==========================
-
-.. autoclass:: skorch.NeuralNetClassifier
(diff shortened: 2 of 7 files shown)


ottonemo:
I think we should either expose all `*Net` classes on top-level like you propose (including `NeuralNet`) or display them in their respective submodules rather than in global name space (e.g. `regressor.NeuralNetRegressor`). Since we want people to use the global import I would vote for the first one but I'm unsure whether we will have more classes like this in the future (e.g. `NeuralNetBinaryClassifier`) which then clutter the global name space. Therefore it might be best to add `regressor.NeuralNetRegressor` and `classifier.NeuralNetClassifier` to the docs. They are mentioned in the neural net section of the docs anyway, right?

thomasjpfan:
That makes sense. I updated this PR by adding `regressor.NeuralNetRegressor` and `classifier.NeuralNetClassifier`.

ottonemo:
Thanks!

PR #257, opened 2018-06-21
DOC: Uses VERSION

### LINKED ISSUE #256: Make doc version depend on VERSION file
Currently copyright date and version are statically defined in `docs/conf.py`.
We should fill in `version` dynamically from the `/VERSION` file, similar to `setup.py`.

thomasjpfan:
Fixes #256

Assumes `VERSION` will always exists in the root directory.


ottonemo:
Thanks :)

Issue #229, opened 2018-05-22
Contribution guidelines


benjamin-work:
We should add guidelines on how to contribute.

damienlancry:
Hi are there guidelines to contribute somewhere now?

BenjaminBossan:
No, but if you find something nice, you can make a proposal. Probably something about being reasonable and being nice :)

PR #258, opened 2018-06-21
API: Adds sorting_order to Callback

### LINKED ISSUE #54: Callbacks: Generalize `yield_callbacks` into processing and output callbacks

thomasjpfan:
Resolves #54


taketwo:
Essentially, the problem is how to let `NeuralNet` know about callback sorting order. I don't think that communicating such information through inheritance is the most elegant solution. As an alternative, what do you think about introducing a class attribute? It can have a clean, self-explaining name (e.g. `sorting_order`), sorting will be simplified to just calling `sorted()` instead of juggling with class names, and it can support more complicated ordering preferences.

Besides, I believe the name `PostponedCallback` is quite confusing. When I saw the issue title, I was very curious to check the code and see what a callback with such a name will do.

Finally, most of the changes in this PR reformat the code. (Looks like _black_ formatting is being replaced by _yapf_ formatting.) I think this really calls for introducing official formatting style and tool recommendation.

thomasjpfan:
I forgot to turn off my auto formatter before submitting this PR. 😅 

I think having a class attribute is a good idea. For a good balance, I propose letting `sorting_order` be from 0 to 10. For callbacks that do not care about order, they would have `sorting_order=5`. The callback order will be ascending, thus higher `sorting_orders` will be last on the `callbacks_` list. What do you think of this idea?

thomasjpfan:
I updated this PR with the `sorting_order` implementation.

ottonemo:
First of all, thanks for tackling this issue. 

Before continuing to work on this PR, though, I think we should define what the actual problem is. I have the feeling that otherwise this could lead to over-engineering. Especially because issue #54 
in itself is not very well fleshed out.

> and it can support more complicated ordering preferences.

I don't think this is necessarily good. What cases should that be? At the moment there is only one relationship that is problematic, namely that there are output callbacks that depend on access to possibly all history values and processing callbacks that write these values. Therefore, processing callbacks need to be executed before output callbacks. And this is scenario is only problematic because we provide some of those by default (e.g., losses and output) and need to merge the default callbacks with a user-supplied list of callbacks while honoring said dependency. For every other, more complicated case, the user can handle the execution of the callbacks by themselves by ordering the list of callbacks manually. Adding another output callback? Just add it to the end. A custom callback depends on a custom score? Just add it to the list after your custom scoring.

It stands to reason if we need to change the current mechanism at all.

> two classes vs. numeric ordering

One reason for my proposal of the abstract class to discern between both callback categories was that numeric ordering is very hard to implement in a meaningful way. There is no semantic, only numbers which leads to arbitrary assignments of these numbers and chaos (see also: `z-index` in CSS). It is easier to have well defined stages that come with an associated name and are matching the specific use-case we want to solve.

taketwo:
I see two  discussion points here: a) numerical vs. binary ordering; b) classes vs. attributes.

### Numerical vs. binary ordering

> I don't think this is necessarily good. What cases would that be?

`Checkpoint` callback? Needs to go after default processing callbacks (access to `_best` column), but before logging callback (writes to `event_` column, as discussed in #246).

> Numeric ordering is very hard to implement in a meaningful way [...] It is easier to have well defined stages that come with an associated name and are matching the specific use-case we want to solve.

Agree, but we need to come up with such names. How would you classify `Checkpoint` callback?

### Classes vs. attributes

Callbacks already have a hierarchy that is induced by their placement in different source files. I don't think introducing a second (orthogonal) hierarchy via inheritance is a good idea. I still hold that sorting order is better expressed by putting a "tag" on a class.

ottonemo:
> Checkpoint callback? Needs to go after default processing callbacks (access to _best column), but before logging callback

And this is what is already happening in the current implementation. Default callbacks are prepended (with exception of `PrintLog` because output), ordered user callbacks are appended and finally `PrintLog` is added. It is only problematic if you mess up the ordering in the user supplied callback list and add `Checkpoint` before your own scoring callbacks.

> How would you classify Checkpoint callback?

It is not a separate class from my point of view. It is a normal callback as it does not output anything in the sense that `PrintLog` is outputting something.

> I still hold that sorting order is better expressed by putting a "tag" on a class.

I have no strong opinion here, I think tags would work fine. I just don't think that *numerical* tags are a good idea. If there's an attribute that says that this callback is an output callback, it would be fine as well.

thomasjpfan:
I think adding ordering is important to clean up the following:

https://github.com/dnouri/skorch/blob/64b54e458d53f158ed27744806999a9a5656d4e6/skorch/net.py#L346-L349

Right now it only checks `PrintLog` because it is one of the default callbacks. The `Checkpoint` callback should also be called last, but since its not in default callbacks it is not check. From a user point of a view, there are no issues, since `self.callbacks` are appended at the end. From a code structure point of view, having `_yield_callbacks` depend on the default callbacks "has a bad code smell".

I agree that having numeric tags may be too flexible. It seems like @taketwo wants to have three levels:

1. Calculates metrics. (`EpochScoring`)
2. Use metrics from 1. and calculates more metrics. (Future `Checkpoint`)
3. Log all metrics from 1 and 2. (`PrintLog`)

We can restrict the tags to three tags, using `enum.Enum`.

taketwo:
I think an important question here is whether we want to reorder user-supplied callbacks to respect ordering tags (no matter numeric or enum) or not.

thomasjpfan:
Currently, we are already reordering user callbacks, based on class type, `PrintLog`. If a user was to provide an additional `Printlog` callback, it will be placed at the end. Reordering callbacks to make sure everything works, sounds like a benefit to me.

Stepping back for a second, this issue stems from wanting to resolve dependencies between callbacks. A full solution would be to have an actual graph that lays out the dependencies between callbacks, but that feels over engineered for this issue. Using an ordered list, with finite levels, provides a simpler solution.

Another solution is to make no changes, and just document that a user must supply the callbacks in an specific order.

ottonemo:
> +1 this starts to feel most reasonable. As long as the default callbacks have only two types (those that should be strictly first, and those that should be strictly last), ordering of the rest can be done by the user.

Agreed. This is already partially documented in http://skorch.readthedocs.io/en/latest/user/neuralnet.html#callbacks.

benjamin-work:
>>> Another solution is to make no changes, and just document that a user must supply the callbacks in an specific order.
> 
>>     +1 this starts to feel most reasonable. As long as the default callbacks have only two types (those that should be strictly first, and those that should be strictly last), ordering of the rest can be done by the user.
> 
> Agreed. This is already partially documented in http://skorch.readthedocs.io/en/latest/user/neuralnet.html#callbacks.

I also agree. We should postpone further changes to how callbacks are ordered until we notice many users struggling with the current solution.

The only thing the documentation appears to be missing is to mention that all `PrintLog`s are moved to the end, including user-supplied ones, which is so fringe that I don't even see the need for a change here. We could probably mention that user-supplied order is important in the `NeuralNet` docstring.

A little bit more meta: Before implementing features like this, it is probably a good idea to pass us a sketch of the intended changes in an issue. That way discussion can happen before code changes that may never be merged (which is why we need contributing guidelines #229).

Issue #261, opened 2018-06-22
Document how to freeze layers


ottonemo:
See #236.

We should add a minimal example of how to filter parameters with `requires_grad=False` from the optimizer. E.g. as proposed in https://github.com/dnouri/skorch/issues/236#issuecomment-392579800:

```python
def filtered_optimizer(pgroups, **kwargs):
    params = filter(lambda t: t[1].requires_grad, pgroups[0]['params'])
    return torch.optim.SGD(params, **kwargs)

net = Net(module, optimizer=filtered_optimizer)
```

This depends on #260 which would help generalize this example by supporting lists and param group dictionaries alike.

thomasjpfan:
The pytorch `Optimizer` class only checks `require_grad` during initialization. Thus something like this will work:

```python
module = module_cls()
net = net_cls(module)

net.initialize()
# Freeze first layer
for p in module.dense0.parameters():
    p.requires_grad_(False)

X, y = data
net.partial_fit(X, y)

# Unfreeze first layer
for p in module.dense0.parameters():
    p.requires_grad_(True)

net.partial_fit(X, y)
```

During the first `net.initalize()` all of the module's parameters have set `require_grad==True`, thus `net.optimizer_` is created without raising an exception. Then we freeze the first layer, and call `paritial_fit` to avoid initializing again. Next, we unfreeze the layer and call `partial_fit`. This setup will only fail when you freeze all the layers in the module. 

This works because `parameter.grad` is `None` when `requires_grad==False`. For example:

```python
x = torch.ones(2, 2, requires_grad=True)
y = x + 2
z = y * y * 3
out = z.mean()
out.backward()

assert x.grad is not None
assert y.grad is None
```

All optimizers checks for `parameter.grad` first before updating any parameters.

benjamin-work:
> The pytorch Optimizer class only checks require_grad during initialization

I would assume that most of the times when someone wants to freeze layers, they have a pretrained net ready and only want to update parts of it (e.g. imagenet transfer learning). In that case, the module would already be initialized and skorch would not reset the module. So that should work out fine.

In case I have an uninitialized module, I would not to initialize it first, which sucks. However, if we have a solution for #107, it should be possible to make it work with uninitialized modules by setting `net = NeuralNet(..., module__weight0__requires_grad=False)`.

thomasjpfan:
[...]

PR #262, opened 2018-06-24
ENH: Adds filter_requires_grad.

### LINKED ISSUE #260: Add utility function to filter non-grad parameters from optimizer
See #236.

For freezing certain layers it would be helpful to mark the freezed parameters by setting `requires_grad=False` and have a utility function filter those weights before passing all weights to the optimizer.

For example:

```python
def filtered_optimizer(pgroups, **kwargs):
    params = skorch.utils.filter_requires_grad(pgroups)
    return torch.optim.SGD(params, **kwargs)

net = Net(module, optimizer=filtered_optimizer)
```

thomasjpfan:
Resolves #260.

`filter_requires_grad` accepts a list of pgroups dictionaries and yields a list of pgroups with the parameters that have `requires_grad==False` filtered out. All other values in the dictionaries are left alone.


skorch/tests/test_utils.py
```diff
@@ -417,3 +417,72 @@ def type_truth_table():
...
+        assert len(list((filter_pgroups[1]['params']))) == 1
+
+        assert filter_pgroups[0]['lr'] == 0.1
+
+    def test_some_params_requires_graident(
+            self, filter_requires_grad):
+        pgroups = [{
+            'params': [
+                torch.zeros(1, requires_grad=True),
+                torch.zeros(1, requires_grad=True)
```
ottonemo: I think we should test wether the filter function does not simply drop whole groups where only one element is filtered. We could do this by setting requires_grad to False here as well.

skorch/utils.py
```diff
@@ -415,3 +415,22 @@ def store_step(self, step):
...
         """Return the stored step."""
         return self.step
+
+
+def filter_requires_grad(pgroups):
+    """Returns parameter groups where parameters with
+    ``requires_grad==False`` are filtered out.
+    """
+    for pgroup in pgroups:
+        output = {k: v for k, v in pgroup.items() if k != 'params'}
```
ottonemo: Why not use `params = pgroup.pop('params')`?

thomasjpfan: As a utility function, I choose not to mutate elements in `pgroups`.

skorch/tests/test_utils.py
thomasjpfan: I agree. PR was updated with this test.

skorch/tests/test_utils.py
```diff
@@ -417,3 +417,90 @@ def type_truth_table():
...
+
+
+class TestFilterRequiresGrad():
+
+    @pytest.fixture
+    def filter_requires_grad(self):
+        from skorch.utils import filter_requires_grad
+        return filter_requires_grad
+
+    def test_all_parameters_requires_graident(
```
benjamin-work: Typo: `graident` -> `gradient`

skorch/tests/test_utils.py
benjamin-work: Where is that tested?

thomasjpfan:
I'm going to put this on hold till the details on how to use this is laid out in   https://github.com/dnouri/skorch/issues/261.

I would prefer these functions be hidden from the user, and `NeutralNet` be able to filter out `require_grads==False` be default.

benjamin-work:
> I would prefer these functions be hidden from the user, and NeutralNet be able to filter out require_grads==False be default.

In general, I would agree that this the more sensible default. pytorch, however, does explicitely not do it, and according to the principle of least surprise, I lean towards going with pytorch.

Still, it should be easy for the user to change the behavior and a helper function + docs is probably sufficient.

benjamin-work:
Sorry that I haven't thought of it earlier, but could you please add a test that shows that if I initialize a net with additional optimizer arguments, the arguments are correctly passed to the optimizer:

`NeuralNet(..., optimizer=filtered_optimizer(torch.optim.SGD), optimizer__momentum=0.9)`

skorch/tests/test_utils.py
thomasjpfan: Ah I misunderstood the comment, I updated this `test_some_params_requires_gradient` to reflect this.

thomasjpfan:
Update this PR with:

1. Adds test to check that arguments are correctly passed to the optimizer. (`test_passes_kwargs_to_neuralnet_optimizer`)
2. Moves functions to `skorch.helper`.

skorch/helper.py
```diff
@@ -78,3 +79,13 @@ def __repr__(self):
     @property
     def shape(self):
         return (self._len,)
+
+
+def filtered_optimizer(optimizer):
```
taketwo: As discussed in #261, this should probably have a different name. Besides, I think docstring can be more descriptive. Right now without checking the code it's not obvious whether `optimizer` argument should be a class or an instantiated object.

thomasjpfan:
I updated this PR with your suggestions.

If `optimizer__filter_requires_grad` gets accepted as a parameter to `NeutralNet`, these functions would be moved to util, and a user would not need to worry about importing `filter_requires_grad`.

benjamin-work:
All in all, this looks good, but I'm not yet satisfied with the name `filter_requires_grad`. First it sounds as if it were a "filter that requires gradients". Second, it's not really a filter but a wrapper. A more telling name would be `optimizer_with_only_params_that_require_grad`, which obviously is too long. So more along `optimizer_params_requires_grad`?

ottonemo:
I think we can avoid the naming issue by generalising, e.g.:
```
def filtered_optimizer(optimizer, filter_fn):
    der opt(params, **kwargs):
        return optimizer(filter_fn(params), **kwargs)
    return opt
```

benjamin-work:
Thanks, good work.

PR #264, opened 2018-06-30
ENH: Adds more step_size control CyclicLR

thomasjpfan:
This PR includes an updated version of `CyclicLR` that allows more control of the step size. It adds an `step_size_2` to `CyclicLR`:

```python
    step_size : int (default=2000)
      Number of training iterations per for first half of a cycle.
      Authors suggest setting step_size 2-8 x training iterations in epoch.

    step_size_2 : int (default=None)
      Number of training iterations per for second half of a cycle.
      If step_size_2 is None, it is set to step_size.
```

For example with `step_size=3` and `step_size=17` the LR scheduling looks like this for one cycle:

screen shot 2018-06-30 at 3 18 48 pm

There was used in recent paper that uses this LR scheduler for language modeling: https://arxiv.org/abs/1801.06146, where this technique increases the rate of convergence.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -226,8 +226,13 @@ class CyclicLR(object):
       may not actually be reached depending on scaling function.
 
     step_size : int (default=2000)
-      Number of training iterations per half cycle. Authors suggest
-      setting step_size 2-8 x training iterations in epoch.
+      Number of training iterations in the first half of a cycle.
+      Authors suggest setting step_size 2-8 x training iterations in an
+      epoch.
+
+    step_size_2 : int (default=None)
+      Number of training iterations in the second half of a cycle.
+      If step_size_2 is None, it is set to step_size.
 
     mode : str (default='triangular')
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
First let me say that I have only little experience with learning rate schedules.

I wonder how that new parameter interacts with the other existing parameters, e.g. with `mode='exp_range'`. If it doesn't, maybe an error/warning should be given if someone sets both. Or would it maybe be better to split `CyclicLR` to separate incompatible modes?

ottonemo:
What about changing `step_size` and `step_size_2` to `step_size_up` and `step_size_down` respectively? The `_2` is not very descriptive.

thomasjpfan:
I updated this PR with a `DeprecationWarning` for using `step_size` and changed the parameter names to `step_size_up` and `step_size_down`.

benjamin-work:
I believe this is almost good to go, I just saw that you did not rename the tests, which still refer to "step_size_2".

benjamin-work:
Thanks again.

PR #268, opened 2018-07-02
ENH: Scoring uses best score from history

### LINKED ISSUE #267: Scoring callbacks don't respect loaded history
When I resume training from a loaded model/history checkpoint, the scores of the very first epoch are always marked as "best". As a side effect, this always leads to creation of a new checkpoint (which is often actually worse).

Here are two relevant functions from `ScoringBase`:

```python
    def initialize(self):
        self.best_score_ = np.inf if self.lower_is_better else -np.inf
        self.scoring_ = convert_sklearn_metric_function(self.scoring)
        self.name_ = self._get_name()
        return self


    # pylint: disable=attribute-defined-outside-init,arguments-differ
    def on_train_begin(self, net, X, y, **kwargs):
        self.X_indexing_ = check_indexing(X)
        self.y_indexing_ = check_indexing(y)
```

Should we move best score initialization to `on_train_begin()` and make it check history (if not empty)? Or any alternative suggestions?

thomasjpfan:
Fixes #267.

The scoring callbacks queries the history and updates `best_score_`  during `on_train_begin`.

This PR enables the following with proper `best_score_` tracking:

```python
net = NeuralNetClassifier(
    module_cls,
    ...
)
net.fit(X, y)
net.save_params('params.h5')
net.save_history('history.json')

net.initialize()
net.load_params('params.h5')
net.load_history('history.json')
net.partial_fit(X, y)
```

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -101,6 +101,9 @@ class ScoringBase(Callback):
     def on_train_begin(self, net, X, y, **kwargs):
         self.X_indexing_ = check_indexing(X)
         self.y_indexing_ = check_indexing(y)
+
+        # Looks for the right most index where `*_best` is True
+        # That index is used to get the best score in `net.history`
         with suppress(ValueError, IndexError):
             best_name_history = net.history[:, '{}_best'.format(self.name_)]
             idx_best_reverse = best_name_history[::-1].index(True)
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -102,7 +102,8 @@ class ScoringBase(Callback):
         self.X_indexing_ = check_indexing(X)
         self.y_indexing_ = check_indexing(y)
         with suppress(ValueError, IndexError):
-            idx_best = net.history[:, f'{self.name_}_best'].index(True)
+            best_name = '{}_best'.format(self.name_)
+            idx_best = net.history[:, best_name].index(True)
             self.best_score_ = net.history[idx_best, self.name_]
 
     def _scoring(self, net, X_test, y_test):
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -1,6 +1,7 @@
 """ Callbacks for calculating scores."""
 
 from contextlib import contextmanager
+from contextlib import suppress
 from functools import partial
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


skorch/callbacks/scoring.py
```diff
@@ -100,6 +101,9 @@ def initialize(self):
     def on_train_begin(self, net, X, y, **kwargs):
         self.X_indexing_ = check_indexing(X)
         self.y_indexing_ = check_indexing(y)
+        with suppress(ValueError, IndexError):
+            idx_best = net.history[:, f'{self.name_}_best'].index(True)
```
taketwo: Is skorch 3.6-compatible only?

thomasjpfan: `suppress` was introduced in python `3.4`: https://docs.python.org/3/whatsnew/3.4.html#contextlib

`skorch` only supports `3`, and the `3.X` versions `pytorch` supports are `3.5` and `3.6`, so I think `suppress` is okay to use.

taketwo: I was actually referring to the use of f-strings ;)

thomasjpfan: OH I see! I updated this to use the `.format` style string instead.

benjamin-work:
Could you explain in a few words what these lines do and why they solve the problem:

```python
    def on_train_begin(self, net, X, y, **kwargs):
        ...
        with suppress(ValueError, IndexError):
            best_name = '{}_best'.format(self.name_)
            idx_best = net.history[:, best_name].index(True)
            self.best_score_ = net.history[idx_best, self.name_]
```

thomasjpfan:
Oh that is a mistake. I updated the code to:

```python
with suppress(ValueError, IndexError):
    best_name_history = net.history[:, '{}_best'.format(self.name_)]
    idx_best_reverse = best_name_history[::-1].index(True)
    idx_best = len(best_name_history) - idx_best_reverse - 1
    self.best_score_ = net.history[idx_best, self.name_]
```

It looks for `*_best*` in reverse to find the index with the latest true value. I also updated the tests to catch that mistake.

benjamin-work:
Okay, now I believe I get it. Could you please add a comment to this passage to clarify why that is needed.

thomasjpfan:
I added a comment. I hope its clear enough.

benjamin-work:
Thank you very much.

PR #269, opened 2018-07-03
ENH: Adds simulation to lr_scheduler

### LINKED ISSUE #266: Add simulation method to LR scheduler base class
It would be nice to have a method that simulates the learning rate schedule for a given number of epochs which returns `(time, lr)`. These values can then be visualized and verified easily. It probably helps during hyper-parameter tuning as well.

thomasjpfan:
Resolves #266

I have been using code similar to this PR to plot learning rates. Hope you find it useful!

Edit: It won't work with `ReduceLROnPlateau` because of the metrics argument.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -47,6 +47,23 @@ class LRScheduler(Callback):
         self.kwargs = kwargs
 
     def simulate(self, steps, initial_lr):
+        """
+        Simulates the learning rate scheduler.
+
+        Parameters
+        ----------
+        steps: int
+          Number of steps to simulate
+
+        initial_lr: float
+          Initial learning rate
+
+        Returns
+        -------
+        lrs: numpy ndarray
+          Simulated learning rates
+
+        """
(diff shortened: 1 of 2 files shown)


benjamin-work:
I think this can be a very useful addition. After testing it a little bit, I believe we should later add a notebook that plots the learning rate for different policies and parameter settings.

Regarding your implementation, I see no point in returning `np.array(steps)`, since this basically returns what you put in. I guess you did this because @ottonemo said in #266 that the method should return `(time, lr)`, but what I think he means is that `time` should be an array of the same length as `lr`, so that you can do:

`plt.plot(*scheduler.simulate(steps=100, initial_lr=0.01))`

But as I see it, `time` would always just be `np.arange(steps)`, so if it were not returned, we could still do:

`plt.plot(scheduler.simulate(steps=100, initial_lr=0.01))`

with the same outcome. My suggestion is thus to only return the learning rates.

Could you please also add a docstring? We should probably also dedicate a section to lr schedulers in the docs, but that can wait for another PR.

PR #270, opened 2018-07-04
Suppress KeyErrors when trying to load best index from history

taketwo:
(This fixes an issue introduced by #268.)

There might be a situation when history does not contain the column that the scoring callback writes to (and thus checks when loading).

Example: "valid_loss" scoring is added to the callback list by default, however it will not output anything to the history if there is no validation dataset (i.e. if `train_split=None`). Reloading such history will raise `KeyError`.

I updated the unit tests to reproduce exactly this situation. This updated test fails without the patch, and succeeds with.


benjamin-work:
good catch

Issue #107, opened 2017-11-01
Support for recursive parameter assignment on module


ottonemo:
It would be helpful to have the ability to set parameters beyond module level (for sub-components of the module, for example):

```python
class Seq2Seq:
    def __init__(self, encoder, decoder, **kwargs):
        self.encoder = encoder
        self.decoder = decoder

class Encoder:
    def __init__(self, num_hidden=100):
        self.num_hidden = num_hidden
        self.lin = nn.Linear(1, num_hidden)

ef = NeuralNet(
        module=Seq2Seq(encoder=AttentionEncoderRNN, decoder=DecoderRNN),
        module__encoder__num_hidden=23,
    )
```

I would expect `module.encoder.num_hidden` to be set to `23`. This should be robust with respect to the initializtion of the sub-module, for example if the encoder has elements that depend on the initialized value, those elements should be updated as well. In the given example, I would expect not only `module.encoder.num_hidden` to be updated to `23` but also that `module.encoder.lin.out_features` is updated (e.g. by re-initializing the whole module).

ottonemo:
With #117 this can be solved:

For example:

```python
class Seq2Seq:
    def __init__(self, encoder, decoder, **kwargs):
        self.encoder = encoder(skorch.utils.params_for('encoder', kwargs))
        self.decoder = decoder(skorch.utils.params_for('decoder', kwargs))
```

This would allow us to write 

```python
ef = NeuralNet(
        module=Seq2Seq(encoder=AttentionEncoderRNN, decoder=DecoderRNN),
        module__encoder__num_hidden=23,
    )
```

which is exactly what we want.

ottonemo:
It would be also nice to support `nn.Sequential`.

ottonemo:
Still open for debate and there's no clear road. Postponing for r0.3.0.

benjamin-work:
Also up for discussion: Not only allow setting parameters on sub-modules, but also on arbitrary attributes. This would, e.g., allow us to things like:

`net.set_params(module__encoder__embeddings__weight__requires_grad=False)`

thomasjpfan:
I propose the following utility or helper functions:

```python
def set_param_in_module(module, param, value):
    name, key = param.rsplit('__', 1)
    name = name.replace('__', '.')
    for n, p in module.named_parameters():
        if n.startswith(name):
            p.__setattr__(key, value)

def set_params_in_module(module, **kwargs):
    for k, v in kwargs.items():
[...]

PR #273, opened 2018-07-04
Allow to recursively set_params on modules and callbacks

benjamin-work:
Addresses #107 

Until now, parameters with `'__'` in their name were just passed as normal parameters to modules/callbacks. Now those parameters are set recursively on modules/callbacks. This allows to do things like:

`NeuralNet(Seq2Seq, module__encoder__num_units=123)`

where skorch will take care of setting `num_units` on `module.encoder`. Previously, it was the user's job to take care of that.

I have not taken the path suggested by @ottonemo that would require the user to write:

`self.encoder = encoder(skorch.utils.params_for('encoder', kwargs))`

the reason being that the proposed change does not require any special handling on the user's side.

`set_params` has become fairly complex. If any of you, @taketwo @thomasjpfan could take a look at this, that would be great.


skorch/tests/test_net.py
```diff
@@ -1263,6 +1264,60 @@ def test_set_params_with_unknown_key_raises(self, net):
...
+            def __init__(self):
+                super().__init__()
+                self.sub_module = module_cls()
+
+            def forward(self, X):
+                return self.sub_module(X)
+
+        net = net_cls(SuperModule, module__sub_module__num_units=123)
+        net.initialize()
+        assert net.module_.sub_module.num_units == 123
```
thomasjpfan: The following tests would fail:

```python
assert net.module_.sub_module.dense0.out_features == 123
assert net.module_.sub_module.dense1.in_features == 123
```

Setting `num_units` after `__init__` does not update the underlying modules. Is this the intended behavior?

taketwo: Exactly, this was my concern as well. This does not really solve the use-case from #107 (specifically, pay attention to the last sentence).

benjamin-work:
Now I see what you mean. I added a failing test case to cover this (and another failing test).

In general, this is a hard problem. There could be potential solutions, but they are certainly hacky and edge cases (e.g. `nn.Sequential`) make it especially difficult. Therefore, I wouldn't want to go down that way.

There is a reason why sklearn mandates that during `__init__`, the only thing happening should be that attributes are set as they come in, exactly to prevent a mess like this.

With this in mind, here are the solutions I came up with:

1) Don't do it at all; if users want this, they need to deal with `module__foo__bar` themselves.
2) Raise an error when an attribute on a sub-module is tried to be set.
3) @ottonemo's proposal in #107.
4) Only allow setting nested attributes when the user makes sure that they are correctly used.

3 is an okay solution but it doesn't allow to do things like `net.set_params(module__encoder__embeddings__weight__requires_grad=False)`. Also, it requires the user to mix skorch code into their module.

Let me explain 4. It would go in the direction of how sklearn does it but still be similar to the established pytorch way. For recursively setting parameters to work, we would require the module to only set attributes as they are, then to call a special method (e.g. `reset_params`, `init_weights`, whatever is most common):

```python
class MyModule(nn.Module):
    def __init__(self, in_features, out_features=123):
        super().__init__()
        self.in_features = in_features
        self.out_features = out_features
        self.reset_params()  # special method name that skorch expects

    def reset_params(self):
        # all sub-modules must be set here
        self.linear = nn.Linear(self.in_features, self.out_features)
        self.sub_module = MySubmodule(in_features=self.out_features)
```

If we could rely on this, I believe we could make this feature work. However, this is certainly a deviation of how pytorch is used normally, so maybe it's not worth it.

taketwo:
### Option 3

> it requires the user to mix skorch code into their module.

I think this alone is a strong enough reason against this option.

> but it doesn't allow to do things like `set.set_params(module__encoder__embeddings__weight__requires_grad=False)`

I'm not sure I understand. Why would you want to do this way instead of direct:

```python
net.module_.encoder.embeddings.weight.reqires_grad = False
```

### Option 4

Overall I like the idea. The users that don't need nested parameter setting don't need to do anything about their code. The users that do want to benefit from this will need to adhere to the rule (and will get a warning if they don't). That said, I am not sure that we can come up with a "no-surprise" interface. Taking your example, what should happen if the user calls `net.set_param(module__sub_module__in_features)`?

thomasjpfan:
Stepping back for a second, what is this feature trying to solve? For me, I would like to have a nice interface for setting requires_grad. Currently I use the following:

```python
class Encoder(nn.Module):
    def __init__(self):
        super().__init__()
        self.embeddings = nn.Embedding(10, 10)

class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.layer1 = nn.Sequential(OrderedDict([
                ('dense0', nn.Linear(5, 20)),
                ('nonlin', nn.ReLU()),
                ('dense1', nn.Linear(20, 10)),
            ])),
        self.layer2 = Encoder()

def set_requires_grad(m, name, value):
    filter_params = (p for n, p in m.named_parameters()
                    if n.startswith(name))
    for p in filter_params:
        p.requires_grad_(value)

m = MyModule()
set_requires_grad(m, 'layer1', False)
```

This will set the requires_grad of all modules in `layer1` to `False`. To get the same behavior with this PR:

```python
m = MyModule()
net = NeuralNetClassifier(
    m,
    module__layer1__dense0__weight__requires_grad=False,
    module__layer1__dense0__bias__requires_grad=False,
    module__layer1__dense1__weight__requires_grad=False,
    module__layer1__dense1__bias__requires_grad=False)
```

The number of keywords can easily get out of hand. I would go with **option 1** and let the user deal with it.

thomasjpfan:
If we want this to work with grid search, then I would prefer a user to define `set_params` and `get_params` in their modules. This would build on a user's familiarity with `sklearn`.

Since a user would define `set_params` in their modules, all `NeutralNet` will have to do is call the top level `set_params` and the module will handle the reset.

Note, this would not allow `module__encoder__embedding__weight__requires_grad=True`. I think restricting the scope of this feature to user defined parameters such as `num_units` is acceptable.

benjamin-work:
> I would prefer a user to define set_params

This would be the easiest way for skorch, but I believe this would be too much to ask, especially since the user would need to handle nested parameters. Maybe there is a middle ground, like my proposal 4, which is somewhat similar to normal pytorch code, but would certainly still not be trivial to implement.

thomasjpfan:
As a reference, `sklearn` deals with the "estimator in another estimator" issue by having classes like `VotingClassifier` inherit from [_BaseComposition](https://github.com/scikit-learn/scikit-learn/blob/f8adfa27ae8256f708a575f75927bd6f6fff6cd7/sklearn/utils/metaestimators.py#L18). The child classes calls `_get_params` in their `get_params` methods.

If we go with `reset_params` I think updating `recursive_set_params` to this should work:

```python
def recursive_set_params(obj, params):
    """Set parameters on 'deep' attributes, separated by '__'"""
    did_setattr = False
    for key, val in params.items():
        if '__' not in key:
            setattr(obj, key, val)
            did_setattr = True
            continue

        name, _, param = key.partition('__')
        if not param:
            continue

        recursive_set_params(getattr(obj, name), {param: val})

    if (did_setattr and hasattr(obj, 'reset_params')
            and callable(obj.reset_params)):
        obj.reset_params()
```

benjamin-work:
I fiddled with your implementation a bit but couldn't find a satisfying solution. It only worked if the parent module was very explicit about how parameters are delegated to child modules, which more or less puts all the burden on the user, so no real progress.

Maybe I have to look at it from a little distance but right now I don't see a solution. And before introducing a very restricted or complicated solution, we should keep the status quo.

benjamin-work:
This should no longer be necessary with the addition of #343, closing this.

timokau:
Can I use `ParamMapper` to do a grid search over the number of units in a nested module? I'm currently doing something like this

```python
DiscreteChoiceEstimator(
    module=ScoringModule,
    module__pairwise_utility=functools.partial(
        PairwiseDenseUtility, hidden_layers=10, units_per_hidden=30
    ),
)
```

to override the `hidden_layers` and `unit_per_hidden` of the nested "pairwise utility module". That is not very grid search friendly and also not good for parameter discoverability. Does #343 enable a better option? Ideally I would be able to define a parameter space like this

```python
params = [
    {
        "module__pairwise_utility": PairwiseDenseUtility,
        "module__pairwise_utility__hidden__layers": [5, 10, 20],
        "module__pairwise_utility__units_per_hidden": [5, 10, 20, 30, 40, 50, 100],
    }, {
        "module__pairwise_utility": SomeOtherUtility,
        "module__pairwise_utility__some_other_parameter": ["a value", "another"],
    }
]
```

BenjaminBossan:
`ParamMapper` is not built to satisfy this requirement. It is applied only after the parameters have already been initialized, so it's not possible to change the number of layers or units at this point.

So how can your problem be solved? Two possibilities come to mind, depending on your use case:

1. You can change the module's definition

If this is possible, I would pass the parameters of your sub-module(s) as explicit arguments to your parent module, which then takes care of routing them to the sub-module(s). E.g.:

```python
class ParentModule(nn.Module):
    def __init__(self, foo, submodule_bar):
        super().__init__()

        self.foo = foo
        self.submodule = SubModule(bar=submodule_bar)
```

In your grid search, you can then address `module__submodule_bar`.

If you don't want to enumerate all these sub-arguments explicitly, you could use `skorch.utils.params_for` instead:

```python
from skorch.utils import params_for

class ScoringModule(nn.Module):
    def __init__(self, foo, **kwargs):
        super().__init__()

        self.foo = foo
        self.submodule = SubModule(**params_for('submodule', **kwargs)
```

In this case, address the parameter as `module__submodule__bar`.

2. You cannot change the module definition

In this case, you could pass already initialized modules as parameters. As long as you don't `set_params` on any further module parameter, skorch will not re-initialize the module but takes it as it is:

```python
modules = [...]  # initialize modules however you like
param_grid = [
    'lr': [0.001, 0.01, 0.1],
    'module': modules,
    ...
]
```

I hope this helps.

timokau:
Thank you for the explanation. The first option is possible, but not ideal. Option 1.1 reduces flexibility (since the parameters that sub-modules can expect have to be encoded in the super-module). Option 2.2 mixes skorch code with "pure" torch code.

Option 2 sounds close to what I'm currently doing.

I'm not sure how it should work either though. I tried writing up my "intuitive expectations", and it just doesn't work without a lot of assumptions. Either of the options you provided is likely good enough in practice. I think 1.2 seems cleanest for now. Maybe it would be possible to write a small `nn.Module` wrapper / mixin that makes it easy to write "scikit-learn friendly modules". Thanks again for taking the time to write up an explanation.

PR #274, opened 2018-07-06
Add support for 'event_' columns in history

taketwo:
(This implements the ideas discussed in #246.)

`PrintLog` will output such columns on the right, just before 'dur'. When the value stored in history is True, a plus sign will be printed to the table. When the value stored in history is False or None, nothing will be printed, i.e. the cell will be empty.

The 'event_' prefix is stripped off from the key so as to not make the table header unnecessary long.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -93,10 +93,14 @@ class PrintLog(Callback):
 
     def format_row(self, row, key, color):
         """For a given row from the table, format it (i.e. floating
-        points and color if applicable.
+        points and color if applicable).
 
         """
         value = row[key]
+
+        if isinstance(value, bool) or value is None:
+            return '+' if value else ''
+
         if not isinstance(value, Number):
             return value
 
@@ -111,12 +115,14 @@ class PrintLog(Callback):
         return template.format(value)
 
     def _sorted_keys(self, keys):
-        """Sort keys alphabetically, but put 'epoch' first and 'dur'
-        last.
+        """Sort keys, dropping the ones that should be ignored.
 
-        Ignore keys that are in ``self.ignored_keys`` or that end on
-        '_best'.
-
+        The keys that are in ``self.ignored_keys`` or that end on
+        '_best' are dropped. Among the remaining keys:
+          * 'epoch' is put first;
+          * 'dur' is put last;
+          * keys that start with 'event_' are put just before 'dur';
+          * all remaining keys are sorted alphabetically.
         """
         sorted_keys = []
         if ('epoch' in keys) and ('epoch' not in self.keys_ignored_):
@@ -126,10 +132,14 @@ class PrintLog(Callback):
             if not (
                     (key in ('epoch', 'dur')) or
                     (key in self.keys_ignored_) or
-                    key.endswith('_best')
+                    key.endswith('_best') or
+                    key.startswith('event_')
             ):
                 sorted_keys.append(key)
 
+        for key in sorted(keys):
+            if key.startswith('event_') and (key not in self.keys_ignored_):
+                sorted_keys.append(key)
         if ('dur' in keys) and ('dur' not in self.keys_ignored_):
             sorted_keys.append('dur')
         return sorted_keys
@@ -138,6 +148,8 @@ class PrintLog(Callback):
         colors = cycle([color.value for color in Ansi if color != color.ENDC])
         for key, color in zip(self._sorted_keys(row.keys()), colors):
             formatted = self.format_row(row, key, color=color)
+            if key.startswith('event_'):
+                key = key[6:]
             yield key, formatted
 
     def table(self, row):
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -78,6 +78,11 @@ class PrintLog(Callback):
       The number formatting. See the documentation of the ``tabulate``
       package for more details.
 
+    stralign : str (default='right')
+      The alignment of columns with strings. Can be 'left', 'center',
+      'right', or ``None`` (disable alignment). Default is 'right' (to
+      be consistent with numerical columns).
+
     """
     def __init__(
             self,
@@ -85,6 +90,7 @@ class PrintLog(Callback):
             sink=print,
             tablefmt='simple',
             floatfmt='.4f',
+            stralign='right',
     ):
         if isinstance(keys_ignored, str):
             keys_ignored = [keys_ignored]
@@ -92,6 +98,7 @@ class PrintLog(Callback):
         self.sink = sink
         self.tablefmt = tablefmt
         self.floatfmt = floatfmt
+        self.stralign = stralign
 
     def initialize(self):
         self.first_iteration_ = True
@@ -172,6 +179,7 @@ class PrintLog(Callback):
             headers=headers,
             tablefmt=self.tablefmt,
             floatfmt=self.floatfmt,
+            stralign=self.stralign,
         )
 
     def _sink(self, text, verbose):
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -35,7 +35,7 @@ class EpochTimer(Callback):
 
 
 class PrintLog(Callback):
-    """Print out useful information from the model's history.
+    """Print useful information from the model's history as a table.
 
     By default, ``PrintLog`` prints everything from the history except
     for ``'batches'``.
@@ -45,6 +45,14 @@ class PrintLog(Callback):
     ``'train_loss_best'`` will be matched with ``'train_loss'``. The
     ``Scoring`` callback takes care of creating those entries, which is
     why ``PrintLog`` works best in conjunction with that callback.
+
+    ``PrintLog`` treats keys with the ``'event_'`` prefix in a special
+    way. They are assumed to contain information about occasionally
+    occuring events. The ``False`` or ``None`` entries (indicating
+    that an event did not occur) are not printed, resulting in empty
+    cells in the table, and ``True`` entries are printed with ``+``
+    symbol. ``PrintLog`` groups all event columns together and pushes
+    them to the right, just before the ``'dur'`` column.
 
     *Note*: ``PrintLog`` will not result in good outputs if the number
     of columns varies between epochs, e.g. if the valid loss is only
--- a/skorch/tests/callbacks/test_logging.py
+++ b/skorch/tests/callbacks/test_logging.py
@@ -10,7 +10,7 @@ class TestPrintLog:
     @pytest.fixture
     def print_log_cls(self):
         from skorch.callbacks import PrintLog
-        keys_ignored = ['batches', 'dur', 'text']
+        keys_ignored = ['dur', 'event_odd']
         return partial(PrintLog, sink=Mock(), keys_ignored=keys_ignored)
 
     @pytest.fixture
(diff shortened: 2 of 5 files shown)


benjamin-work:
Thanks for the PR.

I haven't checked the code in detail yet, but saw that you haven't documented the new behavior yet. It would be nice if you did this, explaining the special handling of events and maybe give a short example of when this could be useful.

taketwo:
I added a paragraph about 'event_' columns to the `PrintLog` docstring. If you'd like to have more documentation elsewhere, please point out to me.

I further updated with PR with proposed changes to `Checkpoint` callback, disabling message printing by default and writing events to history. Here is an example learning session with `Checkpoint` callback enabled:

image

One thing that bugs me is the left alignment of this new column. I do not understand why this happens.

benjamin-work:
> I added a paragraph about 'event_' columns to the PrintLog docstring

That should be fine for now.

> One thing that bugs me is the left alignment of this new column.

I believe this is the `tabulate` default. From the [docs](https://pypi.org/project/tabulate/):

> ``tabulate`` is smart about column alignment. It detects columns which
contain only numbers, and aligns them by a decimal point (or flushes
them to the right if they appear to be integers). Text columns are
flushed to the left.

> You can override the default alignment with ``numalign`` and
``stralign`` named arguments. Possible column alignments are:
``right``, ``center``, ``left``, ``decimal`` (only for numbers), and
``None`` (to disable alignment).

I'm not sure whether we should change this, since it looks like a global setting that would affect all string columns. A possibly easier solution would be to use a `1` instead of a `+`, though I like `+` better.

I probably can't review this thoroughly before Monday, I hope that's fine.

taketwo:
> I'm not sure whether we should change this, since it looks like a global setting that would affect all string columns.

Why not? Even if someone else uses strings in the history, it must be a similar use-case (short string indicating something, not a message). And right alignment is absolutely reasonable for such use case.

> A possibly easier solution would be to use a 1 instead of a +, though I like + better.

I also dislike this.

> I probably can't review this thoroughly before Monday, I hope that's fine.

No hurry, enjoy your weekend!

skorch/callbacks/logging.py
```diff
@@ -91,10 +99,14 @@ def initialize(self):
 
     def format_row(self, row, key, color):
         """For a given row from the table, format it (i.e. floating
-        points and color if applicable.
+        points and color if applicable).
 
         """
         value = row[key]
+
+        if isinstance(value, bool) or value is None:
```
benjamin-work: If I see it correctly, this also modifies the output of values that are not events. I'm not sure if we really want this.

Furthermore, given the change above, I'd like to suggest changing the implementation in a way that "falsey" values lead to `''` and "truthy" values to `'+'`:

```
event_symbols = {True: '+', False: ''}
value = event_symbols[bool(value)]
```

If we have that, we could even move `event_symbols` to the class level, making it easier for the user the change the symbols if needed.

taketwo:
Your inline comment may evolve into a longer discussion, so I'll answer in the main thread.

> I'd like to suggest changing the implementation in a way that "falsey" values lead to '' and "truthy" values to '+'

I don't think it's a good idea. In our original discussion (#246) we talked about events that are not just True/False, but have an associated value. A concrete example was learning rate change event. (I have not worked with LR schedulers yet, so this PR did not add event recording to scheduler classes. But I keep this use case in mind.) We'd probably want to write new rate value to the event column when rate change occurs, and `None` when it does not. And we don't want `0.0` rate to be treated as "no event" and lead to an empty cell. Of course, `0.0` should not happen in the first place, but then there may be some other user-defined event where zero is a valid state. So the main point is, there is semantic difference between _a_ value, and _no_ value. And `None` is an obvious choice to represent no value.

> If I see it correctly, this also modifies the output of values that are not events. I'm not sure if we really want this.

Right, this affects non-event columns as well. I did not want restrict this to event columns because I believe it's generally useful to output `None` as empty cell. At the moment, none of the callbacks shipped with skorch write `None` to history, but there are use cases for that. For example, I have a very expensive epoch scoring which, to save time, I only want to perform every n-th epoch. I thus created a specialization of `EpochScoring` that skips computation on most epochs. However, on such epochs somethings has to be written to the history, otherwise the table layout will be broken. Of course, `None` is the obvious choice here. And clearly, I want empty cells instead of "None".

For reference, this is how it looks like:
image

> If we have that, we could even move `event_symbols` to the class level, making it easier for the user the change the symbols if needed.

We can add a parameter to specify event symbol other than "+", I don't think it's tied to your other suggestion.

One last thing, you did not answer regarding the alignment. On the screenshot you can also see that I'm aligning to the right now, I think it looks more consistent.

benjamin-work:
> we talked about events that are not just True/False, but have an associated value

Yes, right, I didn't think of that.

> this affects non-event columns as well

I agree that it's sometimes useful but I am a little afraid to fundamentally alter user input in the general case. From the point of view of least surprise, I, as a user, would not expect that writing `False` just leads to my result "vanishing". If I have a script at a later point that works on the history and relies on the values (`[do_x() if h is True else do_y() for h in net.history[:, 'mycol']]`), the changes would break it. If I want my values to be mapped to `'+'` or `''`, I can write them to `event_*`, but what do I do if I don't want this? For that reason, I would not touch other columns.

For your case concerning the expensive scoring call, my solution would probably be to return a number that semantically and visually differs from a real score, in your case e.g. -1. (What would happen if you returned an empty string?)

> We can add a parameter to specify event symbol other than "+"

I wouldn't put in too much effort, I only suggested it because it would be trivial with the other change I proposed. Let's try not to have too many parameters.

> I'm aligning to the right now, I think it looks more consistent.

I believe this completely depends on how people use the history. There could certainly be cases where I would rather want to have my text left-aligned (or center?). If we change this, I would advocate for making this a parameter on `PrintLog` so that it can be reverted if needed. Otherwise, I have no strong opinion.

taketwo:
>  If I have a script at a later point that works on the history and relies on the values (`[do_x() if h is True else do_y() for h in net.history[:, 'mycol']]`), the changes would break it. 

I am not sure I understand your point. What I propose does not alternate history in any way, so your script should work fine.

> For your case concerning the expensive scoring call, my solution would probably be to return a number that semantically and visually differs from a real score, in your case e.g. -1. (What would happen if you returned an empty string?)

That's the thing, how a number can be semantically different from another number? Besides, just compare the output, it looks convoluted and the intent is less clear: 

image image

Further, here is resulting plot in TensorBoard, completely screwed up:

image

The suggestion with empty string (and thus empty cell) will work. Well, almost, because if we don't change default alignment to right, the column header may become left aligned. But I still believe `None` (and thus empty cell) conveys the semantics of "no value" perfectly.

> I believe this completely depends on how people use the history. There could certainly be cases where I would rather want to have my text left-aligned (or center?). 

In all use cases for strings in history that I can imagine, it's not really "text", not even sentences or phrases. Rather a word or a symbol. But I admit somebody can have another use and other preferences, so I'd also be fine with a parameter.

benjamin-work:
> What I propose does not alternate history in any way, so your script should work fine.

Yes, you are right again, so forget what I said.

> I'd also be fine with a parameter.

Okay, so if you'd change that, this PR is fine for my purposes.

benjamin-work:
Let's wait and see what @ottonemo thinks, from my part this looks good.

taketwo:
Our real-valued use-case in mind was LR schedulers, however I haven't had a look into those until now. Actually, I don't see how the learning rate changes can be recorded into the history for two reasons: a) skorch's `LRScheduler` delegates LR changes to the policy classes, and (most of) those come from PyTorch, so we can non change them; b) LR can be different for different parameter groups.

Did I overlook something? Do you have other suggestions for real-valued events?

skorch/callbacks/logging.py
```diff
@@ -70,31 +78,42 @@ class PrintLog(Callback):
...
 
+    stralign : str (default='right')
+      The alignment of columns with strings. Can be 'left', 'center',
+      'right', or ``None`` (disable alignment). Default is 'right' (to
+      be consistent with numerical columns).
+
     """
     def __init__(
             self,
             keys_ignored='batches',
```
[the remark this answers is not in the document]
benjamin-work: We could do this, especially since users will always want this. However, that change doesn't need to be in this PR, as the problem exists independently.

ottonemo:
At the moment LR scheduler do not support param groups (at least not directly, but you can specify a dedicated optimizer for a subgroup of parameters and pass that to the LR scheduler to achieve the same effect). Looking at the way LR schedulers are set up in PyTorch, it seems they are meant to work on global learning rates.

What we could do to log learning rate "changes" is to log the learning rate at every epoch, i.e. `net.history.record('event_lr', self.scheduler.get_lr())`, possibly filtering omitting recording when the value matches the previous one.

skorch/callbacks/logging.py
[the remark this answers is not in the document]
ottonemo: Sure, I'll create an issue that documents this.

taketwo:
Ok, I will extend this PR. Do you want any special handling for the schedulers that change LR after each batch? Or just let it record LR after every batch and thus report whatever the LR is by the end of the epoch?

ottonemo:
Great, thanks!

> Or just let it record LR after every batch and thus report whatever the LR is by the end of the epoch?

I think that reporting whatever `get_lr()` reports at the end of the epoch is fine.

> LR can be different for different parameter groups.

You are right, apparently I was mistaken, LR schedulers *do* support param groups. 
A LR schedulers produces a learning rate for each param group. This poses two problems:

1. there is no consistent way of naming param groups, therefore to report these we need to resort to something like `event_lr_pgroup_0` or report only the LR for the default param group (I prefer the former)
2. there is currently no way to conveniently filter such events from the `PrintLog`, i.e. I can't easily filter `event_lr*` (`fnmatch` might help here, similar to how param group filtering during optimizer initialization works)

taketwo:
> there is no consistent way of naming param groups, therefore to report these we need to resort to something like `event_lr_pgroup_0`

I wonder how much insight a log table with columns `lr_pgroup_0`, `lr_pgroup_1`, `lr_pgroup_2`, etc. can provide.

ottonemo:
The order is deterministic so when the user defines the net they know what the index stands for, it is not completely opaque. But you are right that this is not as useful as it could be. I just don't see a better way (e.g., picking the first name of a parameter in the pgroup is too vague for my taste).

taketwo:
To be honest, both LR schedulers and parameter groups of PyTorch are an unexplored territory for me, I haven't used them in practice (yet). Thus I don't have real-world cases to try out the things we discuss, and without that I won't feel confident about the usefulness of the changes.

Can we perhaps branch this out into a separate discussion/PR and revisit it at a later point in time?

ottonemo:
> Can we perhaps branch this out into a separate discussion/PR and revisit it at a later point in time?

Sure. I'll create a separate issue.

taketwo:
Conflicts resolved.

skorch/tests/callbacks/test_logging.py
```diff
@@ -121,7 +129,7 @@ def test_args_passed_to_tabulate(self, history):
             assert tab.call_args_list[0][1]['floatfmt'] == '.9f'
 
     def test_with_additional_key(self, history, print_log_cls):
-        keys_ignored = ['batches']  # 'dur' no longer ignored
+        keys_ignored = ['batches', 'event_odd']  # 'dur' no longer ignored
```
ottonemo: 'batches' is not necessary anymore

skorch/tests/callbacks/test_logging.py
```diff
@@ -10,7 +10,7 @@ class TestPrintLog:
     @pytest.fixture
     def print_log_cls(self):
         from skorch.callbacks import PrintLog
-        keys_ignored = ['batches', 'dur', 'text']
+        keys_ignored = ['batches', 'dur', 'text', 'event_odd']
```
ottonemo: 'batches' is not necessary anymore

taketwo: Right, I was actually suggesting this in the [other thread](https://github.com/dnouri/skorch/pull/288#issuecomment-405922282).

What is this 'text' by the way? Should I remove it also?

ottonemo: Yeah you can remove it as well, seems to be an artifact from the past.

taketwo:
Done.

PR #277, opened 2018-07-08
DOC: Adds logo in header for mobile

thomasjpfan:
Adds logo in header for mobile. (The header turning blue didn't go with the gray side bar in desktop mode)

## Before

screen shot 2018-07-08 at 1 09 59 am

## After

screen shot 2018-07-08 at 1 08 53 am


benjamin-work:
Thank you.

PR #281, opened 2018-07-13
DOC: Adds transfer learning tutorial

thomasjpfan:
This addresses https://github.com/dnouri/skorch/issues/261 and maybe https://github.com/dnouri/skorch/issues/278.

I covered finetuning and how to use a custom train/valid dataset. I think its pretty common for people coming from the pure `PyTorch` approach to want to provide their own train/validation datasets.

I approached the custom datasets by defining a custom `train_valid` function:

```python
def train_valid_split(train, valid, **kwargs):
    return train, valid

net = NeuralNetClassifier(..., train_split=train_valid_split)
net.fit(train_ds, val_ds);
```

This is a pretty hacky approach to the problem. By defining `train_valid_split ` that just returns its inputs, I treated `X` as the training dataset, and `y` as the validation set.

### DIFF
--- a/notebooks/Transfer_learning.ipynb
+++ b/notebooks/Transfer_learning.ipynb
@@ -1,0 +1,369 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Transfer Learning with skorch\n",
+    "\n",
+    "In this tutorial, you will learn how to train a neutral network using transfer learning with the `skorch` API. Transfer learning uses a pretrained model to initialize a network. This tutorial converts the pure Pytorch approach in [Pytorch's Transfer Learning tutorial](https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html#sphx-glr-beginner-transfer-learning-tutorial-py) into using `skorch`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 71,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "from functools import partial\n",
+    "\n",
+    "import torch\n",
+    "import torch.nn as nn\n",
+    "import torch.optim as optim\n",
+    "import numpy as np\n",
+    "from torchvision import datasets, models, transforms\n",
+    "\n",
+    "from skorch import NeuralNetClassifier\n",
+    "from skorch.callbacks import LRScheduler, Checkpoint\n",
+    "from skorch.helper import filtered_optimizer\n",
+    "from skorch.helper import filter_requires_grad"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Problem"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "\n",
+    "Before we begin, download the data from [here](https://download.pytorch.org/tutorial/hymenoptera_data.zip) and extract it to the current directory.\n",
+    "\n",
+    "We are going to train a neutral network to classify **ants** and **bees**. The dataset consist of 120 training images and 75 validiation images for each class. First we create the training and validiation datasets:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 18,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data_dir = 'hymenoptera_data'\n",
+    "train_transforms = transforms.Compose([\n",
+    "    transforms.RandomResizedCrop(224),\n",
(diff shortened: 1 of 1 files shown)


thomasjpfan:
The accuracy is on par with the original tutorial. It’s the benefit of using a pretrained model and only training the final layers.

I would want this tutorial to only have best practices. I think not having an official way to support custom training and validation datasets, is a high barrier of entry.

benjamin-work:
> It’s the benefit of using a pretrained model and only training the final layers

It works really well.

> I think not having an official way to support custom training and validation datasets, is a high barrier of entry.

Yes, that would be ideal. When we started out with skorch, we put a lot of thought into how to manage train/test splits and even though the current solution does not support this use case, we found it to be the best. At the same time, we thought about implementing a "predefined split" that would support this exact use case, but we left it open. Maybe now is the time to implement it.

I guess that it could look something like this:

```python
class PredefinedSplit:
    def __init__(self, dataset_valid):
        self.dataset_valid = dataset_valid

    def __call__(self, dataset, y=None, groups=None):
        if self.dataset_valid is None:
            raise ValueErrror("some useful message")
        return dataset, self.dataset_valid

    def __getstate__(self):
        state = self.__dict__.copy()
        state['dataset_valid'] = None
        return state
```
But there are certainly some edge cases that still need covering.

taketwo:
> Maybe now is the time to implement it.

:+1: from my side. I work mostly with datasets that have official train/test splits. So far I've resorted to using custom scoring functions that capture test dataset and ignore the data that `EpochScoring` passes to them. But this is again hacky and it would be great to have skorch support for predefined splits.

benjamin-work:
I created a new issue: #282

Regarding this PR: It could either be left as is until the new feature is included, or we wait for it first.

thomasjpfan:
1. I will wait for #282 to be resolved.

2. Regarding `examples` vs `notebooks`, as long as I can keep all the logic in one notebook, I would prefer this tutorial to be in `notebooks`. 

3. I added a `datasets` directory into `notebooks` that can be used by notebooks. The downloading logic is placed in the notebook itself. This allows the notebook to be able to run on all platforms.

benjamin-work:
This looks really good. I have only 2 minor points left:

The download function could be made a little more flexible by allowing to pass the folder. I made this and some minor adjustments with when/how to print things:

```
def download_and_extract_data(dataset_dir='datasets'):
    data_zip = os.path.join(dataset_dir, 'hymenoptera_data.zip')
    data_path = os.path.join(dataset_dir, 'hymenoptera_data')
    url = "https://download.pytorch.org/tutorial/hymenoptera_data.zip"

    if not os.path.exists(data_path):
        if not os.path.exists(data_zip):
            print("Starting to download data...")
            data = request.urlopen(url, timeout=15).read()
            with open(data_path, 'wb') as f:
                f.write(data)

        print("Starting to extract data...")
        with ZipFile(data_zip, 'r') as zip_f:
            zip_f.extractall(dataset_dir)
        
    print("Data has been downloaded and extracted to {}.".format(dataset_dir))
```

Also, on my machine, the final validation accuracy was a little lower than 0.9608, so maybe you could adjust the sentence in the last paragraph to something like `...with a validiation accuracy of roughly 0.96`.

BenjaminBossan:
Looks good, fantastic work. I will merge on Monday.

notebooks/Transfer_Learning.ipynb
```diff
@@ -0,0 +1,400 @@
...
+   "metadata": {},
+   "source": [
+    "# Transfer Learning with skorch"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this tutorial, you will learn how to train a neutral network using transfer learning with the `skorch` API. Transfer learning uses a pretrained model to initialize a network. This tutorial converts the pure Pytorch approach in [Pytorch's Transfer Learning tutorial](https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html#sphx-glr-beginner-transfer-learning-tutorial-py) into using `skorch`.\n",
```
benjamin-work: > This tutorial converts the pure Pytorch approach in ... **into** using `skorch`

There is a word missing after `into`.

benjamin-work:
I was just about to merge when I discovered a small error (see comment). After this, we should really be good.

notebooks/Transfer_Learning.ipynb
taketwo: Nitpick: we typically capitalize PyTorch.

thomasjpfan: I reworded the sentence into the following:

This tutorial converts the pure PyTorch approach described in [PyTorch's Transfer Learning Tutorial](https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html) to `skorch`.

Issue #282, opened 2018-07-13
Better support for predefined splits


benjamin-work:
Addresses concerns raised in #278 and #281 

When users already have data that is split into train and validation, it is currently difficult to make skorch use those without resorting to hacks.

A possible solution is to have a class that can be passed to `train_split` and returns the predefined split. It could look somewhat like this:

```python
class PredefinedSplit:
    def __init__(self, dataset_valid):
        self.dataset_valid = dataset_valid

    def __call__(self, dataset, y=None, groups=None):
        if self.dataset_valid is None:
            raise ValueErrror("some useful message")
        return dataset, self.dataset_valid

    def __getstate__(self):
        state = self.__dict__.copy()
        state['dataset_valid'] = None
        return state
```

Apart from the question whether there could be another, better solution, those are some concerns:

* What to do when my predefined split data is not in a `DataSet`?
* What to do when persisting the split -- just discard the validation data?

thomasjpfan:
Another approach would be:

```python
class TrainValidDataset(torch.utils.data.Dataset):
    def __init__(self, train_ds, valid_ds):
        self.train_ds = train_ds
        self.valid_ds = valid_ds

def train_valid_split(ds, y, **kwargs):
    return ds.train_ds, ds.valid_ds

train_valid_ds = TrainValidDataset(train_ds, val_ds)

net = NeuralNetClassifier(..., train_split=train_valid_split)
net.fit(train_valid_ds, y=None)
```

This keeps all the data in the `train_valid_ds` object. Thus `net` does not need to worry about state.

If `TrainValidDataset` is provided in `skorch`, we can automatically do this in `NeutralNet.get_split_datasets`:

```python
def get_split_datasets(self, X, y=None, **fit_params):
    dataset = self.get_dataset(X, y)
    if isinstance(X, TrainValidDataset):
        dataset_train, dataset_valid = X.train_ds, X.valid_ds
    ...
```

taketwo:
@benjamin-work  This solution feels strange. The user will have to specify the validation dataset as a part of `NeuralNet` instantiation, whereas the train dataset is given to the `fit()` call.

@thomasjpfan This solution feels better. Do we even need `train_valid_split` function?

thomasjpfan:
If we were to update `get_split_datasets`, then `train_valid_split` is not needed. I updated the comment to address this.

benjamin-work:
[...]

PR #283, opened 2018-07-14
WIP: Adds TrainValidDataset

thomasjpfan:
This is PR will be updated based on the discussions in #282.

Currently, `TrainValidDataset` is a namedtuple containing a training and validation dataset. When `get_split_datasets` is called, it first checks for this namedtuple, and returns the corresponding training and validation datasets.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -284,8 +284,9 @@ class CVSplit(object):
 
 
 class TrainValidDataset:
-    """Small wrapper class to support custom training and validation
-    datasets. This class can be passed into ``NeutralNet.fit``,
+    """Small wrapper class to support pre-existing training and
+    validation datasets.
+    This class can be passed into ``NeutralNet.fit``,
     ``NeutralNet.fit_loop``, ``NeutralNet.parital_fit``, or
     ``NeutralNet.get_split_datasets``.
 
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -281,3 +281,23 @@ class CVSplit(object):
     def __repr__(self):
         # pylint: disable=useless-super-delegation
         return super(CVSplit, self).__repr__()
+
+
+class TrainValidDataset:
+    """Small wrapper class to support custom training and validation
+    datasets. This class can be passed into ``NeutralNet.fit``,
+    ``NeutralNet.fit_loop``, ``NeutralNet.parital_fit``, or
+    ``NeutralNet.get_split_datasets``.
+
+    Parameters
+    ----------
+    train_dataset: Dataset
+        Training dataset
+
+    valid_dataset: Dataset
+        Validation dataset
+
+    """
+    def __init__(self, train_dataset, valid_dataset):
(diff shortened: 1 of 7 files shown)


BenjaminBossan:
Continuing the discussion from #282.

If this case is common enough, skorch should make room for it. I have to say, though, that I'm not 100% content with this solution. Sure, it might be necessary, but it clashes with many parts of skorch:

* What to do when you use this but also define a `train_split`? Having both makes no sense.
* When my data is not already in a dataset, I have to wrap it myself, which is normally something skorch does under the hood.
* This is difficult to use in an sklearn pipeline (do we need a helper transformer for that?).
* skorch already supports a similar way to achieve this outcome ([`PredefinedSplit`](http://scikit-learn.org/stable/modules/generated/sklearn.model_selection.PredefinedSplit.html)).

Of course, these kinds of conflicts will necessarily arise when building a bridge between two worlds, but I have the feeling we should encourage one best practice. For me, this is having my data in a more universal format (numpy arrays, dicts, etc.) and not in a custom data format that only works with pytorch and nowhere else.

Long story short: I agree with this PR (when some of the bullet points are fixed) because it facilitates working with pre-existing splits, but would always encourage to take the effort to have the data in a universal format as long as possible.

skorch/dataset.py
```diff
@@ -281,3 +281,23 @@ def __call__(self, dataset, y=None, groups=None):
     def __repr__(self):
         # pylint: disable=useless-super-delegation
         return super(CVSplit, self).__repr__()
+
+
+class TrainValidDataset:
+    """Small wrapper class to support custom training and validation
```
BenjaminBossan: Here and further down, you describe this as "custom datasets" but this is misleading. A custom dataset suggests that I implemented my own type of dataset. I would rather go with "predefined" or "pre-existing split" (maybe the latter to avoid confusion with sklearn's `PredefinedSplit`). Also, I would add a newline after the first sentence.

skorch/net.py
```diff
@@ -607,10 +608,14 @@ def fit_loop(self, X, y=None, epochs=None, **fit_params):
             * a dictionary of the former three
             * a list/tuple of the former three
             * a Dataset
+            * a TrainValidDataset
```
BenjaminBossan: maybe add: a skorch.dataset.TrainValidDataset, since most people won't know where this comes from.

skorch/tests/test_net.py
```diff
@@ -1444,3 +1444,14 @@ def get_step(self):
         expected_losses = list(
             flatten(net.history[:, 'batches', :, 'train_loss']))
         assert np.allclose(side_effect[2::3], expected_losses)
+
+    def test_train_valid_dataset(self, net_cls, module_cls, data, dataset_cls):
```
BenjaminBossan: What is missing somewhat from these tests is to show that `train_ds` is actually used for training and `valid_ds` for validation. This could be achieved, e.g., by mocking `get_iterator` and tracking what it is called with.

PR #285, opened 2018-07-16
BUG: Adds pickle support to filtered_optimizer

thomasjpfan:
Allows filtered_optimizer to be pickled.

### DIFF
--- a/skorch/tests/test_helper.py
+++ b/skorch/tests/test_helper.py
@@ -282,10 +282,7 @@ class TestOptimizerParamsRequiresGrad:
         assert len(net.optimizer_.param_groups) == 1
         assert net.optimizer_.param_groups[0]['momentum'] == 0.9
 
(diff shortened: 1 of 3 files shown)


skorch/tests/test_helper.py
```diff
@@ -280,3 +281,11 @@ def forward(self, X):
...
         assert len(net.optimizer_.param_groups) == 1
         assert net.optimizer_.param_groups[0]['momentum'] == 0.9
+
+    def test_pickle(self, filtered_optimizer, filter_requires_grad, tmpdir):
+        opt_fn = tmpdir.mkdir('filtered_optiizer').join('opt.pkl')
+        opt = filtered_optimizer(torch.optim.SGD, filter_requires_grad)
+
+        # Does not raise
+        with opt_fn.open('wb') as f:
+            pickle.dump(opt, f)
```
ottonemo: Maybe use `pickle.dumps` instead to avoid temporary files.

PR #286, opened 2018-07-16
ENH: Adds predefined_split

thomasjpfan:
Fixes https://github.com/dnouri/skorch/issues/282

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -115,7 +115,7 @@ dictionary. Below, there is example code on how to achieve this:
 I already split my data into training and validation sets, how can I use them?
 ------------------------------------------------------------------------------
 
-If you have already defined a training and validation datasets that are
+If you have predefined training and validation datasets that are
 subclasses of PyTorch :class:`~torch.utils.data.Dataset`, you can use
 :func:`~skorch.helper.predefined_split` to wrap your validation dataset and
 pass it to :class:`~skorch.net.NeuralNet`'s ``train_split`` parameter:
@@ -128,8 +128,7 @@ pass it to :class:`~skorch.net.NeuralNet`'s ``train_split`` parameter:
         ...,
         train_split=predefined_split(valid_ds)
     )
-    # None is passed because train_ds already contains the target
-    net.fit(train_ds, None)
+    net.fit(train_ds)
 
 If you split your data by using :func:`~sklearn.model_selection.train_test_split`,
 you can create your own skorch :class:`~skorch.dataset.Dataset`, and then pass
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -116,8 +116,9 @@ I already split my data into training and validation sets, how can I use them?
 ------------------------------------------------------------------------------
 
 If you have already defined a training and validation datasets that are
-subclasses of :class:`~torch.utils.data.Dataset`, you can use
-:func:`~skorch.helpers.predefined_split` to train your network:
+subclasses of PyTorch :class:`~torch.utils.data.Dataset`, you can use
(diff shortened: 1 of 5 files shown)


docs/user/FAQ.rst
```diff
@@ -110,3 +110,43 @@ dictionary. Below, there is example code on how to achieve this:
...
     net = MyNet(MyModule, ..., criterion__reduce=False)
     net.fit(X, y)
+
+
+I already split my data into training and validation sets, how can I use them?
+------------------------------------------------------------------------------
+
+If you have already defined a training and validation datasets that are
+subclasses of :class:`~torch.utils.data.Dataset`, you can use
+:func:`~skorch.helpers.predefined_split` to train your network:
```
taketwo: Somehow the sentence "you can use `predefined_split()` to train your network" does not sound right to me. We are not using this function to train the network. We are using this function to attach to the network a fixed validation dataset that is independent on the training data passed to the `fit()` method.

thomasjpfan: I see what you mean. I rephrased the sentence to try to make it more clear.

docs/user/FAQ.rst
```diff
@@ -110,3 +110,44 @@ dictionary. Below, there is example code on how to achieve this:
     # for each sample so that it can be weighted
     net = MyNet(MyModule, ..., criterion__reduce=False)
     net.fit(X, y)
+
+
+I already split my data into training and validation sets, how can I use them?
+------------------------------------------------------------------------------
+
+If you have already defined a training and validation datasets that are
```
ottonemo: > [...] defined *a* training [...]

typo

docs/user/FAQ.rst
```diff
@@ -110,3 +110,44 @@ dictionary. Below, there is example code on how to achieve this:
...
+.. code:: python
+
+    from skorch.helpers import predefined_split
+
+    net = NeuralNet(
+        ...,
+        train_split=predefined_split(valid_ds)
+    )
+    # None is passed because train_ds already contains the target
+    net.fit(train_ds, None)
```
ottonemo: for `NeuralNet` you don't need to pass `y=None`, this is only necessary for classifiers and regressors.

skorch/tests/test_helper.py
```diff
@@ -280,3 +281,23 @@ def forward(self, X):
...
+        from skorch.dataset import Dataset
+
+        valid_dataset = Dataset(*data)
+        train_split = predefined_split(valid_dataset)
+
+        tmp_fn = tmpdir.mkdir('predefined_split').join('test.pkl')
+
+        # does not raise
+        with tmp_fn.open('wb') as f:
+            pickle.dump(train_split, f)
```
ottonemo: Maybe just dump the pickle into a string using `pickle.dumps`? Saves the temporary file hassle.

benjamin-work:
Just generally: Do we want to have `make_split` a public function? What other purpose does it serve? Also: Do we want to have one in `utils.py` and the other in `helper.py` or not both in the same module (which)?

skorch/helper.py
```diff
@@ -113,3 +114,21 @@ def filtered_optimizer(optimizer, filter_fn):
...
+    >>> valid_ds = skorch.Dataset(X, y)
+    >>> net = NeutralNet(..., train_split=predefined_split(valid_ds))
+
+    Parameters
+    ----------
+    dataset: torch Dataset
+       Validiation dataset
+
+    """
+    from skorch.utils import make_split
```
benjamin-work: Could the import not be at top level?

thomasjpfan:
`predefined_split` is in `helper.py` since it is meant to accessed by a user.`make_split` is in `utils.py` since it only used by `skorch` internally. 

This PR was updated to convert `make_split` to a private function.

PR #288, opened 2018-07-17
ENH: Always adds batches to keys_ignored

thomasjpfan:
Fixes https://github.com/dnouri/skorch/issues/284.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -52,7 +52,7 @@ class PrintLog(Callback):
 
     Parameters
     ----------
-    keys_ignored : str or list of str (default='batches')
+    keys_ignored : str or list of str (default=None)
       Key or list of keys that should not be part of the printed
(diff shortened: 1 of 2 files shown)


benjamin-work:
Thanks

taketwo:
Perhaps this should have removed 'batches' from lines 13 and 124.

benjamin-work:
@taketwo what do you mean?

taketwo:
I meant these functions in `test_logging.py`:

```python
    def print_log_cls(self):
        from skorch.callbacks import PrintLog
        keys_ignored = ['batches', 'dur', 'text']
        return partial(PrintLog, sink=Mock(), keys_ignored=keys_ignored)
```

and

```python
    def test_with_additional_key(self, history, print_log_cls):
        keys_ignored = ['batches']  # 'dur' no longer ignored
```

No need to have `'batches'` in the ignored keys anymore.

Issue #250, opened 2018-06-13
Remove old deprecations for 0.3 release


benjamin-work:
Should only concern the `use_cuda` parameter in `NeuralNet` and `Dataset`.

PR #292, opened 2018-07-20
Clean ups for release 0.3

benjamin-work:
* Remove deprecations related to use_cuda/device
* Add a missing space in error message.
* Improve docstring of to_tensor.
* Cosmetic: Add missing space after colon.

Addresses #250

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -2,6 +2,7 @@
 
 from functools import partial
 from numbers import Number
+import warnings
 
 import numpy as np
 from sklearn.model_selection import ShuffleSplit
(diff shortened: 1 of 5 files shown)


taketwo:
Does skorch have any deprecation policy? The `device` parameter was removed after 0.2 release, i.e. the deprecation warning was never a part of any release. Thus for those who stick with released versions you are effectively breaking interface without a warning.

benjamin-work:
Technically, you are correct. The general changes (which were related to PyTorch 0.4) that resulted in the deprecation were in place in 0.2 but we forgot to adapt this part of the code (meaning it should have been in 0.2).

So we could extend the deprecation period. But perhaps this feature was used so little that we can get away with removing it nonetheless?

taketwo:
This is a philosophical question; a clear deprecation policy would help to answer this.

Given that skorch is still in its 0.Xes... it's maybe kind of okay. Still, with 1.5K stars I suppose there are thousands of users, and there is a chance some of them still have this `device` around. IMHO it would be non-decent maintainership to break their code without warning and a grace period. Especially when it costs nothing to have this warning around.

benjamin-work:
Our unofficial deprecation policy is to wait for 1 release, I believe at this stage moving fast is still okay.

I restored the warning in `Dataset` for the reasons you gave.

Issue #279, opened 2018-07-12
Documentation for train-split doesn't match with code


spott:
In:

https://github.com/dnouri/skorch/blob/a4eae21d7409c478a05a571ac67e5f18ec996f19/skorch/net.py#L139-L143

The documentation says that the `train-split` method should return a `(X_train, X_val, y_train, y_val)` tuple, however, it appears that it should instead pass a `(X_dataset, y_dataset)` tuple, as seen in:

https://github.com/dnouri/skorch/blob/a4eae21d7409c478a05a571ac67e5f18ec996f19/skorch/net.py#L1068-L1069

BenjaminBossan:
Your are right, we forgot to update the docstring when we made a change in `CVSplit`. Thanks.

spott:
I can create a pull request for this, if needed.  Just let me know.

BenjaminBossan:
> I can create a pull request for this

Yes, that would be great.

benjamin-work:
@spott Are you working on the PR?

spott:
@benjamin-work:

Thanks for the reminder.  I had forgotten about this.

ottonemo:
Resolved by #293.

PR #293, opened 2018-07-20
Update documentation to reflect correct return of `train_split`

spott:


spott:
Fixes #279

BenjaminBossan:
Thank you very much.

Issue #289, opened 2018-07-18
Should Checkpoint save history as well?


ottonemo:
Continuation of the discussion in #246.

I think the points can be summarized to

1. should there be a `save_history` parameter
2. should we support pickling the whole model instead of using `save_*` methods (which consequently includes the history)
3. should we introduce a second `target` parameter for the history, e.g. `history_target` and `params_target`
4. should we find a better name for `target`

I think that saving the history alongside the model is a good idea, so yes to (1). Having the additional ability to use pickle for checkpointing via (2) sounds like a good idea but I'm not yet convinced where this feature makes life easier for the user, it would be nice to have an usage example that demonstrates the benefit of this feature. For (3) I would argue that we can re-use the existing parameter, e.g. strip the folder (if any) and append a static file name or append a `.history` suffix to the whatever is stored in `target`. Regarding (4) I think that something akin to `path` is a better choice but I don't have a strong opinion here.

benjamin-work:
1) Then it's decided
2) First it can be a matter of convenience. Second, the code that loads the model may not care for whether it's a skorch NeuralNet or a completely different sklearn model. In that case, pickle would work but `save/load_params` would be useless.
3) I agree we could do it like that but this is a little bit less flexible.
4) We could try to be consistent with our existing code, i.e. use `f`. That name is not great in itself, but conforms to PyTorch.

thomasjpfan:
1, 3, 4. `history_target` can be used define a path and configure `Checkpoint` to save the history. `params_target` is a good name in the context of `Checkpoint` since it is saving model parameters.

2. I think pickling the whole `NeuralNet ` object would make the model less portable. Current, after a user is done training a model in `skorch`, they easily move their module and parameters to another environment.

benjamin-work:
I'm not too keen on the `*_target` names, is there a precedent for this?

ottonemo:
It is not necessarily a path but possibly a file-like object. It denotes in every case the thing data is written to, i.e. the target destination of the checkpoint callback.

benjamin-work:
Yes, it's a path or file-like, but PyTorch (and, as an extension, `skorch.NeuralNet.save_params`) calls this variable `f`. Granted, `f` is not the best variable name (though there is the convention of `with open(...) as f:`), but neither is target. When in doubt, I would opt for consistency.

thomasjpfan:
Consider the following API:

```python
class Checkpoint(Callback):
    ...
    def __init__(
            self,
            f='',
[...]

PR #296, opened 2018-07-24
Support saving params, history, and pickling in Checkpoint

taketwo:
This PR is a work in progress to add the features discussed in #289. I would like to get feedback before updating unit tests and documentation.

Some notes/discussion points:
 * I dropped sink output for a bunch of reasons:
   * Too verbose if all three targets are used
   * Meaningless if target is a file-like object
   * Would unnecessary complicate `save_model()`
   * Now that we have events in the log I don't see much value in having it anyway
 * This change may break existing code that did not explicitly specify argument names in constructor call. For example, `Checkpoint('model.pt', 'train_loss_best')` will now behave differently. Not sure how to deal with this other than introducing [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/).

### DIFF
--- a/skorch/tests/callbacks/test_training.py
+++ b/skorch/tests/callbacks/test_training.py
@@ -20,6 +20,16 @@ class TestCheckpoint:
             mock.side_effect = lambda x: x
             yield mock
 
+    @pytest.yield_fixture
+    def save_history_mock(self):
+        with patch('skorch.NeuralNet.save_history') as mock:
+            yield mock
+
+    @pytest.yield_fixture
+    def pickle_dump_mock(self):
+        with patch('pickle.dump') as mock:
+            yield mock
+
     @pytest.fixture
     def net_cls(self):
         """very simple network that trains for 10 epochs"""
@@ -93,7 +103,7 @@ class TestCheckpoint:
             ('my_score', scoring),
             checkpoint_cls(
                 monitor='epoch_3_scorer',
-                target='model_{last_epoch[epoch]}_{net.max_epochs}.pt',
+                f_params='model_{last_epoch[epoch]}_{net.max_epochs}.pt',
                 sink=sink),
         ])
         net.fit(*data)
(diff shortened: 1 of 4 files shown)


benjamin-work:
Thanks for taking this task.

Regarding dropping `sink`: I wouldn't do this because it can still be useful for things like logging (where you usually don't want to use `history`). However, I see that the message itself can hardly reflect every combination of parameters. So maybe a simple message like "A checkpoint was triggered in epoch {epoch}." is enough?

Regarding backwards compatibility: I would like to see your example above still working. For this, I would change the arguments to be:

```
     def __init__(
            self,
            target=None,
            monitor='valid_loss_best',
            f_params='params.pt',
            f_history=None,
            f_pickle=None,
    ):
        if `target` is not None:
            # deprecation warning
            f_params = target
```

Then you can deprecate the use of `target` and the code above still does what it is supposed to do.

taketwo:
Good ideas, I updated the code.

One thing that I noticed is that `isinstance(f, file)` fails with "name file not defined". Should I check for being `io.BytesIO` instead, or rather existence of `write` attribute (that's what [`pickle.dump`](https://docs.python.org/3/library/pickle.html#pickle.dump) requires)?

benjamin-work:
> One thing that I noticed is that isinstance(f, file) fails

Can you not get around this issue by using `skorch.utils.open_file_like`?

thomasjpfan:
You can call `_format_target` on `self.f_pickle`, then use contextmanager , `skorch.utils.open_file_like`, which will yield a file object to be consumed by `pickle.dump`.

taketwo:
Thanks for the hint, very useful. Unless you have other comments, implementation-wise this is now done.  I'll add tests and documentation soon.

taketwo:
I've updated the documentation and added a unit test.

skorch/tests/callbacks/test_training.py
```diff
@@ -105,6 +115,22 @@ def epoch_3_scorer(net, *_):
         assert net.history[2, 'event_cp'] is True
         assert all((x is False) for x in net.history[3:, 'event_cp'])
 
+    def test_save_targets(
+            self, save_params_mock, save_history_mock, pickle_dump_mock,
+            net_cls, checkpoint_cls, data):
+        sink = Mock()
```
benjamin-work: You are assigning a sink but don't really use it anywhere. Why not stay with the default?

taketwo: Just a copy-paste left-over

taketwo:
> For the warning, could you add a sentence that the user should best use keyword arguments? 

Sure. But maybe we should rather _force_ the user to use keyword arguments?

benjamin-work:
> But maybe we should rather force the user to use keyword arguments?

Nah, I'm not a fan of using the force :)

taketwo:
Should be good to go.

taketwo:
By the way, in the current implementation the whole `net` object is pickled. Is this desired? Maybe only `net.module_` should be pickled? What's the best practice?

Some background: I haven't used pickling so far and tried it just now. My `net` can not be pickled because one of the callbacks uses `_thread.lock` objects. `net.module_` on the other hand can be pickled just fine.

ottonemo:
The optimizer’s state and possibly the state of the criterion may be of value to the user as well and can’t be restoRed from init parameters. The same goes for callbacks such as lr schedulers. But we should definitely take another look at what needs to be pickled and what can be avoided. For callbacks we might provide a default implementation of getstate that avoids parameters with a specific pattern.

benjamin-work:
> By the way, in the current implementation the whole net object is pickled. Is this desired?

Yes, it is very much intended. In addition to the advantages mentioned above, this allows a user to do things like:

```
with open('mymodel.pkl', 'rb') as f:
    model = pickle.load(f)

y_pred = model.predict(X)
```

which is less line of code and it works with *every sklearn model*. If we only save `module_` or use `save_params`, the user would need to know the model specifics to load it. Pickling the whole model shows the major advantage of using a standard API and a standard persistence protocol.

benjamin-work:
Thanks for implementing my change requests, good work.

taketwo:
Oh okay, makes sense.

>  For callbacks we might provide a default implementation of getstate that avoids parameters with a specific pattern.

Could you elaborate on this idea? Maybe branch out a new issue if you think it will be worth implementing indeed.

Issue #295, opened 2018-07-23
A callback for modifying the loss before the optimizer obtains it.


spott:
There is a class of things that don't appear to be able to be possible within the current callback framework of Skorch.

The first one that comes to mind is in adversarial training.  In order to do efficiently, you would want to run all samples through the network once, create new samples by adding some noise that is dependent on the gradient to the old samples, then run those through again:

```
Xi.requires_grad = True
output = net(Xi)
loss = criterion(output, yi)
loss.backward()

sign = torch.ge(0, Xi.grad)
sign = (sign - 0.5)/.5
Xi_at = torch.add(Xi, -epsilon, sign)

at_output = net(Xi_at)
loss_at = criterion(at_output, yi)
loss_at.backward()

optimizer.step()
```

Unfortunately, this isn't actually possible using the current framework (without creating a new `NeuralNet` class).

This could probably be fixed if `Xi` and `yi` were passed to the `on_grad_computed` callback, but it isn't currently.

benjamin-work:
In general, I see no obstacle to passing `Xi` and `yi` to `on_grad_computed`. If you'd like, you can open a PR for this.

For your particular problem, however, I believe that overriding `train_step_single` (if you work on the current master branch) to look similar to what you proposed could be the better approach. The reason is that I would normally not expect another training step to be made within a callback.

One of our goals with skorch was to make it easy to subclass `NeuralNet` and tinker with its methods. There are too many NN architectures to be able to cover them all with one class. We do that all the time for our own projects. If you find subclassing difficult, tell us where we could improve further.

ottonemo:
The question is if things like adversarial training can be modularised and reused for different architectures. It would be nice to have a pluggable virtual adversarial training callback but I'm not sure if this is feasible.

In general I'm with Benjamin on this, you should resort to using the object framework of python to make such fundamental changes to a net.

spott:
> In general I'm with Benjamin on this, you should resort to using the object framework of python to make such fundamental changes to a net.

I feel like this kind of thing -- data augmentation or regularization -- shouldn't really need a whole new `NeuralNet` class to make work.  That just makes it harder to play around with adding and removing these kinds of regularizers... ideally you should be able to do a grid search with and without adversarial training!

> If you'd like, you can open a PR for this.

Will do.

benjamin-work:
> I feel like this kind of thing -- data augmentation or regularization -- shouldn't really need a whole new NeuralNet class to make work

[...]

PR #297, opened 2018-07-26
Feature/grad add data to callback 'on_grad_computed'

spott:
Work towards #295


taketwo:
Please update documentation [here](https://skorch.readthedocs.io/en/latest/user/callbacks.html#on-grad-computed). Also probably makes sense to update the signature in all classes that have this callback, even though they don't use the new arguments (`NeuralNet`, `GradientNormClipping`, `Callback`).

spott:
This is something that I'm a little bit unclear about:

The `Callback` base class only has `net` and `**kwargs` for all callback functions.  I actually had to dig into the `NeuralNet` class in order to figure out what is passed, and what the names of the parameters are.

I figure this is to allow the callbacks to be flexible when the `NeuralNet` class is subclassed, but maybe it should be documented somewhere other than the source code?

I'm happy to add the calling signature to `Callback`, `NeuralNet` and `GradientNormClipping` if we think that they should be there?

BenjaminBossan:
> I figure this is to allow the callbacks to be flexible when the NeuralNet class is subclassed, but maybe it should be documented somewhere other than the source code?

Yes, this is the reason why we did it this way. It is somewhat documented [here](https://skorch.readthedocs.io/en/latest/user/callbacks.html#callback-base-class) but if you think more documentation is needed, we should add it.

> I'm happy to add the calling signature to Callback, NeuralNet and GradientNormClipping

So far, we haven't added arguments that were not needed in the methods but technically, it would be better if the methods have the same signature, so now might be a good time to start it.

docs/user/callbacks.rst
```diff
@@ -48,44 +48,51 @@ on_train_begin
 ^^^^^^^^^^^^^^
 
 Called once at the start of the training process (e.g. when calling
-fit).
+fit).  Signature: `on_train_begin(net, X, y)`.
```
taketwo: Maybe paste signatures directly into the headers? There are precedents for doing this already: https://skorch.readthedocs.io/en/latest/user/neuralnet.html#fit-x-y

benjamin-work:
@spott give us a ping when this is ready for review

spott:
@benjamin-work. I think this is ready for review.

I've changed the generic callback signatures, but not the signatures for each of the callbacks (for example, `GradientNormClipping`'s `on_grad_computed` callback doesn't accept `Xi` and `yi`, it just accepts `**kwargs`:  I think it is cleaner that way (it is only accepting that which it is using).  If you disagree, I can change it.

`net` and `callbacks/base.py` have been changed to reflect the accepted callback signature.  I have left `**kwargs` in there to allow for flexibility.

benjamin-work:
LGTM, thanks.

PR #298, opened 2018-07-27
WIP: Update to 0.4.1

thomasjpfan:
This PR updates skorch to support PyTorch 0.4.1. There is one small optimization to `skorch.utils. to_tensor` by using [`torch.as_tensor`](https://pytorch.org/docs/stable/torch.html#torch.as_tensor), which never copies unless necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,6 @@ install:
   - source activate skorch-env
   - conda install --file=requirements-dev.txt
   - python setup.py install
-  - conda install -c pytorch 'pytorch-cpu>=0.4.0'
+  - conda install -c pytorch 'pytorch-cpu>=0.4.1'
 script:
   - pytest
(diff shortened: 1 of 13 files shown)


taketwo:
Maybe also update import/docs for `Subset` since it was moved (https://github.com/pytorch/pytorch/pull/7816).

BenjaminBossan:
Nice and quick work. I will have a deeper look at this on Monday. Too bad that this came shortly after our release ;) Are the changes breaking or only deprecated?

taketwo:
Subset is not breaking.

>  Too bad that this came shortly after our release ;)

What about skorch 0.3.1? :)

thomasjpfan:
There aren't any breaking changes:

1. `torch.nn.functional.sigmoid` was deprecated in favor of  `torch.sigmoid`
2. `torch.nn.functional.tanh` was deprecated in favor of  `torch.tanh`
3. `torch.utils.data.dataset.Subset` can now be imported up a level, `torch.utils.data.Subset`.

thomasjpfan:
Since the PyTorch version on pypi is built on cuda-9.0, this PR also updates the Dockerfile to use cuda-9.0.

benjamin-work:
LGTM, @ottonemo do you see anything missing?

PR #300, opened 2018-07-29
ENH: Prints header on train start

thomasjpfan:
When `partial_fit` is called a second time, the header is not printed. This PR adds header printing before every train event.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -106,6 +106,9 @@ class PrintLog(Callback):
         self.keys_ignored_.add('batches')
         return self
 
+    def on_train_begin(self, net, **kwargs):
+        self.first_iteration_ = True
+
     def format_row(self, row, key, color):
         """For a given row from the table, format it (i.e. floating
         points and color if applicable).
--- a/skorch/tests/callbacks/test_logging.py
+++ b/skorch/tests/callbacks/test_logging.py
@@ -74,6 +74,13 @@ class TestPrintLog:
         expected = ['epoch', 'nmse', 'train_loss', 'valid_loss']
         assert columns == expected
 
+    def test_header_partial_fit(self, net, data, sink):
+        net.partial_fit(*data)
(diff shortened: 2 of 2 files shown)


thomasjpfan:
CI is failing because it installs PyTorch 0.4.1

benjamin-work:
I believe we had a reason for not doing this, tough I can't remember for sure. Perhaps @ottonemo has a better memory network than me.

One reason could have been to better support outside fit loop:

```
net = NeuralNet(..., max_epochs=1).initialize()
for _ in range(max_epochs):
    for X, y in my_data_generator():
        net.partial_fit(X, y)
```

But I'm not sure whether this is a realistic use case. Even if it were, one could get around it by using `fit_loop` after the first call.

thomasjpfan:
This PR is not urgent. It makes working interactively in a notebook look slightly nicer when calling `partial_fit` to continue training.

taketwo:
Actually I do use an outside fit loop. It's similar to what you have posted:

```python
    while len(net.history) < net.max_epochs:
        net.partial_fit(dataset, epochs=(net.max_epochs - len(net.history)))
        if visualize():
                break
```

With this I can _ctrl+c_ while training and visualize the outputs of the network. When I close the visualizer training process resumes. Without this PR the log table stays nearly intact (besides from ^C line), whereas with this PR I get a header after each time I break out for visualization.

Do you think my use-case is legitimate or is there a better way?

benjamin-work:
> Do you think my use-case is legitimate

Yes, as mentioned, this may have been the reason why we chose not to reset the attribute.

> is there a better way

The only way I see is to use `(partial_)fit` on the first call and `fit_loop` in subsequent calls.

thomasjpfan:
@taketwo Good point! I have not thought of that use-case.

benjamin-work:
I don't believe we should have different behaviors for this depending on the environment. tqdm is a special case.

Issue #294, opened 2018-07-23
Callback name inference


taketwo:
At the moment callback names are either supplied explicitly as first element of a tuple, or are inferred from the callback class name. If two or more callbacks of the same class are used, explicit names have to be used to avoid collision. This leads to the following code (snippet from skorch itself):

```python
    def _default_callbacks(self):
        return [
            ('epoch_timer', EpochTimer()),
            ('train_loss', BatchScoring(
                train_loss_score,
                name='train_loss',
                on_train=True,
                target_extractor=noop,
            )),
            ('valid_loss', BatchScoring(
                valid_loss_score,
                name='valid_loss',
                target_extractor=noop,
            )),
            ('print_log', PrintLog()),
        ]
```

I see some room for improvement here. Note that the strings `train_loss` and `valid_loss` have to be specified twice. First time to define a callback name, second to define a history column name. While technically it's possible to have different names, I doubt it's a realistic use-case. I, for one, always find myself copy-pasting such a callback and then adjusting the name two times, which is annoying. I think it would be very useful if the callback name could be inferred automatically from the history column name.

Second thing to notice is that instead of using automatic inference from the class name (i.e. `EpochTimer` or `PrintLog`), skorch itself opts to explicitly `snake_case` the names. I'd argue that this can be made the default behavior.

To summarize, I propose to update `NeuralNet._yield_callbacks()` such that:
* When callback is an instance and has `name` attribute, it is used.
* When callback name is inferred from class name, it is snake cased.

BenjaminBossan:
You're making good points. It's always a preferable to avoid repetition and be more pythonic. There are, however, some counter-arguments to your points:

Auto-naming: When a user writes their own callback which so happens to have a `name` attribute, this could lead to confusion when their existing code suddenly stops working. Moreover, you could create name conflicts were there currently are none, e.g. when two different callbacks have a `name` attribute with the same value.

Snake-casing: We thought about this one too. But again you have to consider custom callbacks by users. You cannot guarantee that they use camel-case. What do you do with `My_Callback`? You would need a function that unambiguously identifies names that should be snake-cased. And again, the change could break existing code.

In sum, that doesn't mean your proposals are rejected, just that they need some reflection. The current solutions may be ugly at times, but they are also simple and predictable.

[...]

PR #301, opened 2018-07-31
On name conflict assign unique callback names

ottonemo:
Callback names will be appended with a count suffix to
make the name unique if it appears more than once.
The suffix is of format `_{i}` with `i` starting at 1
with underscore as a separator to be addressable using
keywords arguments.

Example:

    e1 = EpochScoring(name='a')
    e2 = EpochScoring(name='b')
    net = Net(callbacks=[e1, e2]).initialize()
    cbs = dict(net.callbacks_)

Results in:

    cbs['EpochScoring_1'] == e1
    cbs['EpochScoring_2'] == e2

This addresses #294.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -363,7 +363,7 @@ class NeuralNet(object):
         not unique, a ValueError is raised.
 
         """
-        names_seen = set()
+        names_seen = {}
         callbacks_ = []
 
         class Dummy:
@@ -372,11 +372,19 @@ class NeuralNet(object):
             pass
 
         for name, cb in self._yield_callbacks():
-            if name in names_seen:
-                raise ValueError("The callback name '{}' appears more than "
-                                 "once.".format(name))
-            names_seen.add(name)
+            names_seen[name] = names_seen.get(name, []) + [cb]
 
+        named_callbacks = []
+
+        for name, cbs in names_seen.items():
+            for i, cb in enumerate(cbs):
+                if len(cbs) > 1:
+                    unique_name = '{}_{}'.format(name, i+1)
+                else:
+                    unique_name = name
+                named_callbacks.append((unique_name, cb))
+
+        for name, cb in named_callbacks:
             # check if callback itself is changed
             param_callback = getattr(self, 'callbacks__' + name, Dummy)
             if param_callback is not Dummy:  # callback itself was set
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -324,7 +324,8 @@ class NeuralNet(object):
         pass
 
     def _yield_callbacks(self):
-        """Yield all callbacks set on this instance.
+        """Yield all callbacks set on this instance including
+        a set whether its name was set by the user.
 
         Handles these cases:
           * default and user callbacks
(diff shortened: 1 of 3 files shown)


benjamin-work:
I agree with the general gist of this feature, namely making it easy to add callbacks when I don't care about their names.

A small problem I see with this implementation is that it may silently rename callbacks that the user named explicitely. So modifying your example:

```
net = Net(callbacks=[('scoring', e1), <many-other-callbacks>, ('scoring', e2)]).initialize()
net.callbacks_  # returns [..., ('scoring_1', ...), ..., ('scoring_2', ...)]
```

In my opinion, this should either still raise an error or at least trigger a warning. For the other case, not having specified names, I'm fine with the automated counting.

Another issue right now is that this does not prevent duplicate names:

```
net = Net(callbacks=[('scoring', e1), ('scoring', e2), ('scoring_1', e3)]).initialize()
net.callbacks_  # returns [..., ('scoring_1', ...), ('scoring_2', ...), ('scoring_1', ...)]
```

I know this is a little bit contrived but it should still not happen, since `set_params` is not well behaved in this situation.

skorch/tests/test_net.py
```diff
@@ -571,10 +571,13 @@ def test_callback_same_inferred_name_twice(self, net_cls, module_cls):
...
         net = net_cls(module_cls, callbacks=[cb0, cb1])
 
-        with pytest.raises(ValueError) as exc:
-            net.initialize()
-        expected = "The callback name 'some-name' appears more than once."
-        assert str(exc.value) == expected
+        net.initialize()
+
+        cbs = dict(net.callbacks_)
+        assert 'some-name_1' in cbs
```
taketwo: Unrelated to this PR, but shouldn't dashes in the custom names be replaced with underscores?

benjamin-work: Why?

taketwo: Names with dashes can not be used to construct keyword arguments, so `net.set_params(callbacks__some-name_1__foo=10)` is not possible. Or is the assumption that if the user has chosen a name with a dash, then he does not plan to use it with `set_params()`?

benjamin-work: > Or is the assumption that if the user has chosen a name with a dash, then he does not plan to use it with set_params()?

In general, when a user wants to use commodities like `set_params`, they should make sure that their parts cooperate. We can only give them guidance.

For your example, I believe the dash is still no problem:

`net.set_params(**{'callbacks__some-name_1__foo':10})`

ottonemo:
@benjamin-work pls review again

skorch/net.py
```diff
@@ -336,19 +337,54 @@ def _yield_callbacks(self):
...
+            if len(cbs) > 1 and name in names_set_by_user:
+                raise ValueError("Found duplicate user-set callback name "
+                                 "'{}'. Use unique names to correct this."
+                                 .format(name))
+
+            for i, cb in enumerate(cbs):
+                if len(cbs) > 1:
+                    unique_name = '{}_{}'.format(name, i+1)
+                    if unique_name in grouped_cbs:
+                        raise ValueError("Setting unique name failed since "
```
benjamin-work: As in the error message above, I would mention that we deal with *callback names* to make it extra clear.

benjamin-work:
Almost good to go. 

Could you also adjust the docstring [here](https://github.com/dnouri/skorch/blob/2ac11debf6ebd2d3610e199d12e3c119722643c5/skorch/net.py#L145) to reflect the new changes?

(And while you're at it, could you update the code example at the end of the description: `net.set_params(callbacks__print_log__keys=['epoch', 'train_loss']` should read `keys_ignored` instead)

ottonemo:
@benjamin-work addressed your comments.

benjamin-work:
LGTM, thanks.

PR #303, opened 2018-08-03
MRG: Adds image segmentation example

### LINKED ISSUE #160: Add example of image segmentation task
There should be an example (notebook) that shows how to deal with this kind of problem. Additionally, it could show how to load images and segmentation masks on the fly.

thomasjpfan:
Resolves #160 

After reading the comments, here is a list of things to be updated:

* [x] Add requirements.txt file to directory.
* [x] Add docstrings to custom classes/functions.
* [x] Add link to UNet paper.
* [x] Add more details about using `skorch` features.
* [x] Remove abstraction for freezing layers.
* [x] Improve readability of unqualified names (`filtered_optimizer`).
* [x] Return axes when plotting
* [x] Add IOU to metrics
* [x] Add nbviewer link to readme


BenjaminBossan:
Thanks, this looks fantastic, I will review this on Monday.

examples/nuclei_image_segmentation/README.md
```diff
@@ -0,0 +1,13 @@
+# Nuclei Image Segmentation With UNets
+
+## Setup
+
+1. Install dependencies: `pip install -r requirements.txt`
+1. Follow [Kaggle's installation and configuration documentation](https://github.com/Kaggle/kaggle-api#installation) to install and configure the kaggle cli
+1. Go to Kaggle's [2018 Data Science Bowl Competition](https://www.kaggle.com/c/data-science-bowl-2018), click on "Late Submission" and accept the terms and conditions to get access to the data.
+1. Run `./dl_extract_prepare.sh` to download, extract and prepare the data.
```
taketwo: For me this fails on `from PIL import Image`. This again comes to the requirements installation. Neither Pillow nor torchvision is in `requirements.txt`, so novice user may be left puzzled what's going on and why something is missing after the dependencies installation step has succeeded.

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
```diff
@@ -0,0 +1,521 @@
...
+   "metadata": {},
+   "source": [
+    "# Nuclei Image Segmentation Tutorial"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this tutorial, we will implement a UNet to solve Kaggle's [2018 Data Science Bowl Competition](https://www.kaggle.com/c/data-science-bowl-2018). The challenge asks participants to find the location of nuclei from images of cells. In additional to `skorch`, this tutorial uses `matplotlib` for plotting, `torchvision` for image augmentation and Kaggle's offical cli to download data. Please follow [Kaggle's installation and configuration documentation](https://github.com/Kaggle/kaggle-api#installation) to install the kaggle cli. Next, navigate to the [competition page](https://www.kaggle.com/c/data-science-bowl-2018), click on \"Late Submission\" and accept the terms and conditions to get access to the data."
```
taketwo: This duplicates the instructions given in README, but are different (e.g. here `torchvision` is explicitly requested). I think there should be a unified introduction, same inside the notebook and in the README.

Also, I think the introduction should outline the approach that will be taken to solve the problem. In its current state, it only mentions that this is going to be a UNet model (by the way, a link to the UNet paper would be useful). Would be nice to explain in a nutshell what UNet is, which backbone network will be used for feature extraction, where the weights come from, how we will freeze them, etc.

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
```diff
@@ -0,0 +1,521 @@
...
...allJZiOMxJbYNcW/89fa60uWCuATLOBZsYaZx0tQXsiQhdtgQGWnVvjkhzy3D3ApuVrqpu7HCTTzSydszcW7HLpbaTYMxw81BubRNbA33hvCzjF6qDBNhp1u7v71mtVuOXTgQE05X+qXIK0Hp9h44/1NZTDNup8w9d+xAj+1TdsR1PMb5P1RvrOgS0TzGoxxYAx86J75069iUlMntPsbr+e3TDiGDGH4B+TGQI/WHrda1WqxEQOFPkoCyySzGIbNre2L7pa19ycWZ1s9mMnrvAXiKFOJbA3kJiuVyOW7YOTAAuLy/He9L3PW/evBkXHH68X9/BhYMnXwTHextBnN+bqLF1WUQEXrG9Dga9ftdsRtDk4BYYg4mi5tLrjiDWZQ6RUYxzIO4UeBuAPUDuY+H99uDF6LoQ56YXP88BsIhwf3/P58+fGYaBz58/77kTTAOjnMH1eqLm2cfav6O9vT52cTwcHHubvI9epy82SylcXl6O/fBxd2Y4SkC8v15nXPTERa33P7qROHiFnX2YA+HIhEev8yj3eEl5MZjNNRn4S0YSigpJE7WFDyFizKcKFuxlACY7oBQZvUcR2651ptW/chLFsoU5I4saKJYIjExl2SK9qDU1thdjW6VaOlJRam1AW63tWcYq9mBOAmqTSeQZdFfCfK683sJqraTznvsfC6uV0HXC8qqj6xJUuPl+S/4gyCbzV38j9K1C6YRuadDMgRJ1TG+A8XyZbiHITEEKmjJlI6xSQdaCbpWyBhVhVkAolI1w/wFqzSxmlcVXJlVI/iDPwvKrwn/1Tz3/6T/2/P4/3dJ1/RjUVUk83qy5mAtpMUO6Ytv31e5bQUm9eb5qy9ylAt1CuLwQPp7Dw43JQ5Y5UebGjG+2IA9QrjCHh2oWbAWzLsvYakaKgtYGlJUq2SQnRUdPV3q7t0nt9Zp2C6Isia63YL+6UcrW7MMszkzpUoLO7v+2Cn0DxxZYyJjNTXZTiNKSeUhDuMnlCamOADug6i/emsu/YJx18C9EZyf89+mXSDz/GBA4BPqmrOZzAdUxdvTYccfKU6BlyrRO2/0c4PZSYHSozkNg9RCTO2WQj7Xr2Dl/Tpm25dBCIQKBeH3VXQR7nEvx+MikRtDjrJiDBmcf45yNmdFiLvpji6+/JMP9/1e5vr7eGzcvbrbvEfCR0XJQ43ICl2s44L+6umI2m3
```
taketwo: I think it's more educational to not abstract away how the layers are frozen. After all, isn't that one of the purposes of the tutorial to demonstrate fine-tuning?

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
```diff
@@ -0,0 +1,521 @@
...
+    "from pathlib import Path\n",
+    "\n",
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "import torch\n",
+    "from torch.optim import SGD\n",
+    "import torch.nn as nn\n",
+    "from torch.nn.functional import binary_cross_entropy_with_logits\n",
+    "from sklearn.model_selection import train_test_split\n",
+    "from skorch.helper import (\n",
```
taketwo: I am not sure if it's good to import unqualified names. This is supposed to be educational notebook, but if everything is imported like this and is used without prefixes, then while reading the notebook it's impossible to tell whether a function is a skorch function or a custom utility. E.g., is `filtered_optimizer()` a skorch function? What about `freeze_layer()`?

examples/nuclei_image_segmentation/README.md
```diff
@@ -0,0 +1,13 @@
+# Nuclei Image Segmentation With UNets
+
+## Setup
+
+1. Install dependencies: `pip install -r requirements.txt`
```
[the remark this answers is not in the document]
benjamin-work: > I think it could be beneficial if we formulate what are the points that this notebook is supposed to teach.

From my reading, this notebook is intended for users who are familiar with the general topic (image segmentation) and pytorch but want to get into skorch. I agree that this could be made explicit at the start. However, I find the amount of code shown in the notebook to be sufficient. The reader may or may not look at the particular code if they like. Since the code is sufficiently clear, this should be no problem (a *short* docstring for some of the classes and functions would be nice, though).

> This is step is somewhat strange since it will only get the dependencies of skorch, not skorch itself or Jupyter.

The install instruction in the notebook is better, true (but it also lacks the instruction to install jupyter notebook). Either improve on this instruction or leave it out since it's already documented. What I would like to see in the README is a link to the notebook using nbviewer. Then the user can look at the installation instruction in the notebook before needing to isntall jupyter.

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
benjamin-work: > by the way, a link to the UNet paper would be useful

:+1:

Then I don't believe that an explanation of UNet is still necessary.

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
benjamin-work: I agree that making it a little more explicit will be helpful.

benjamin-work:
This is a great notebook, thank you for putting in so much effort.

I only have some minor points left:

It can sometimes be nice if plotting functions return the (fig and) ax object so that a user may add or modify things (disadvantage: have to put `;` at the end of the function to suppress `__repr__()`).

As far as I can tell, the source code is all yours, but if you re-used some code, make sure to attribute it correctly.

It would be really nice to have an additional validation metric during training, ideally IoU, as used to score the competition.

You duplicate the code for `BCEWithLogitsLossPadding` in the notebook. Is there a specific reason for this? I seem to remember that there was a jupyter magic command to just show code, which would avoid the duplication. However, I can't find it anymore :/

I also found that the lazy data loading takes a lot of time, for me half of the training time:

```
%time xx = [train_ds[i] for i in range(1852)]
CPU times: user 5min 31s, sys: 6.09 s, total: 5min 37s
Wall time: 44.2 s
```

Maybe we should consider just loading everything to memory? From my calculation, the amount of data is not that large:

`(1852 * (3 + 1) * 288 * 288 * 4 / (2 ** 20) -> ~2344MB`

I also ran the model without pre-training, I show the results just for completeness:

```
  epoch    train_loss    valid_loss    cp      dur
-------  ------------  ------------  ----  -------
      1        0.3617        0.2183     +  94.4275
      2        0.2251        0.2020     +  88.3065
      3        0.2234        0.2295        88.4043
      4        0.1918        0.1807     +  88.5496
      5        0.1713        0.1685     +  88.4643
      6        0.1727        0.1142     +  88.3118
      7        0.1309        0.1144        88.3109
      8        0.0999        0.0926     +  88.7039
      9        0.0928        0.1002        88.1258
     10        0.0812        0.0920     +  88.1639
     11        0.0808        0.0836     +  88.0856
     12        0.0708        0.0925        88.0929
     13        0.0695        0.0903        88.2675
     14        0.0654        0.0810     +  88.1655
     15        0.0641        0.0806     +  88.1365
     16        0.0635        0.0793     +  88.0718
     17        0.0634        0.0788     +  88.1757
     18        0.0633        0.0811        88.1508
     19        0.0739        0.0830        88.3374
     20        0.0648        0.0781     +  88.4009
```

It's interesting that without pre-training, the model is better at the start but worse at the end.

examples/nuclei_image_segmentation/README.md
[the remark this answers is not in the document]
taketwo: > From my reading, this notebook is intended for users who are familiar with the general topic (image segmentation) and pytorch but want to get into skorch

That's a legit target. But then perhaps more emphasis can be put where skorch features are used. Right now it feels more like a "how to solve that dataset" example, but should be more like "how skorch makes it easy to solve that dataset".

thomasjpfan:
> You duplicate the code for BCEWithLogitsLossPadding in the notebook. 

@benjamin-work Can you expand on this statement?

thomasjpfan:
Regarding lazy loading, I would hope that using four workers in the `Dataloader` would be enough to keep the GPU fed. I will implement keeping everything in memory to see it changes training time.

examples/nuclei_image_segmentation/Nuclei_Image_Segmentation.ipynb
```diff
@@ -0,0 +1,570 @@
...
...izKcKFuxlACY7oBQZvUcR2651ptW/chLFsoU5I4saKJYIjExl2SK9qDU1thdjW6VaOlJRam1AW63tWcYq9mBOAmqTSeQZdFfCfK683sJqraTznvsfC6uV0HXC8qqj6xJUuPl+S/4gyCbzV38j9K1C6YRuadDMgRJ1TG+A8XyZbiHITEEKmjJlI6xSQdaCbpWyBhVhVkAolI1w/wFqzSxmlcVXJlVI/iDPwvKrwn/1Tz3/6T/2/P4/3dJ1/RjUVUk83qy5mAtpMUO6Ytv31e5bQUm9eb5qy9ylAt1CuLwQPp7Dw43JQ5Y5UebGjG+2IA9QrjCHh2oWbAWzLsvYakaKgtYGlJUq2SQnRUdPV3q7t0nt9Zp2C6Isia63YL+6UcrW7MMszkzpUoLO7v+2Cn0DxxZYyJjNTXZTiNKSeUhDuMnlCamOADug6i/emsu/YJx18C9EZyf89+mXSDz/GBA4BPqmrOZzAdUxdvTYccfKU6BlyrRO2/0c4PZSYHSozkNg9RCTO2WQj7Xr2Dl/Tpm25dBCIQKBeH3VXQR7nEvx+MikRtDjrJiDBmcf45yNmdFiLvpji6+/JMP9/1e5vr7eGzcvbrbvEfCR0XJQ43ICl2s44L+6umI2m3F9fU1KifPz83G8YX+x4dd2iYCDRX/dwZgHji6Xyz/RdsagHgdUMUI+AnMHrT5HnHWNcgMHw96OyOJ5XyOomzKkXl8MMowJDaKOM0oSHFD63z6m7onqINTBnvfNfVR///vf8/j4yN3dHSlZUhC3Q4syBAfxvs3u8gsH6DEI0vvtRIXXEbM8ul54Pp/vsbgPDw/j/V4sFuM8iAFtHszmOmO/z4f0s16XL0xc8uKA1hcVs9mM7777jvv7e96/fz8y476Qur+/H+/zdDHykvJCMNsYKn+mO7PVGFFFW1COs2JmcJ8wsJD8i08ilLPizKiItPAoB3xGxe48Uxn3wkdGLz6Y1B86XoM2raZ/6YI0j9wxeMzbkryXLYCpEwYVFktleWXv8NBZ8JNYFqycjNlNtVJWwupzYvMtkCp96kznm0rTtYKn7XVlr0pt2+MmS80zYb4VBlGGYnZYuU+kuaWHnW1hcyesC2x/Nm/afpm4fJXpqzLLjs4s69j5eeLVb3r+8J83ZJ0xyBY6OL/smXeVLlnEfkoJLSBiRl2qttBIvfnuIj7YQp+FxSzxKC3
```
benjamin-work: "compeititons" -> "competition's"

benjamin-work:
There are some typos left but they should be easy to fix.

Regarding caching: True, it doesn't seem to help.

Regarding IoU: The validation score is better than the best score on Kaggle, is that because of the approximation?

thomasjpfan:
> Regarding IoU: The validation score is better than the best score on Kaggle, is that because of the approximation?

My validation set probably does not represent their test sets distribution.

benjamin-work:
> My validation set probably does not represent their test sets distribution.

Can this be improved by using a different train/valid split than random? E.g., if samples should be divided by subject, a `GroupKFold` could be used.

thomasjpfan:
Since there isn't a group variable to group by, we would need a method to divide the dataset into groups. There are several ways to accomplish this:

1. Since there are three visual distinct types of images, we can use traditional computer vision techniques to group the images.
2. Train a unsupervised model to group the images.

We would need to verify that the group distribution in the entire dataset is consistant with the group distribution in the test set. Then we can use these groups with `GroupKFold`.

I considered doing this, but thought it would stray too far from using `skorch`.

benjamin-work:
> Since there isn't a group variable to group by, we would need a method to divide the dataset into groups

Never mind, I thought there might be a patient ID or something.

So this is a great addition, thank you very much for putting so much effort into it.

PR #307, opened 2018-08-07
Fix imports in the basic README example

taketwo:
* Remove `torch` import (unused)
* Import `NeuralNetClassifier` from `skorch` (moved away from `skorch.net`)


benjamin-work:
Thank you very much.

PR #310, opened 2018-08-09
MRG: Removes _missingno and refactors error handing in history

### LINKED ISSUE #306: ProgressBar (maybe) prevents garbage collection
I started to use the `ProgressBar` callback today and noticed that it affects GPU memory consumption of my program.

Specifically, without this callback, `nvidia-smi` consistently reports `4247MiB` usage by my program. After adding the progress bar, consumption starts from the same number, but then steadily increases throughout the first epoch, eventually peaking around `7500MiB`, and does not change anymore in subsequent epochs. I could reproduce this behavior in every run. I also tried to run with different batch sizes. The absolute numbers change, e.g. for half the batch it's `2502MiB` and `4182MiB`, but it's still a huge overhead.

Looking at the callback code, I don't see any reason for this behavior. Perhaps the console IO somehow prevents garbage collection? No idea.

Most likely this is not a skorch issue. Nevertheless, I think it would be great if we can confirm this, and maybe add a warning to the callback documentation.

thomasjpfan:
Fixes #306

### DIFF
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -17,17 +17,19 @@ def partial_index(l, idx):
 
     if needs_unrolling or needs_indirection:
         output = []
-        error_n = None
+        error = None
         for n in l:
             try:
                 output.append(partial_index(n, idx))
-            except KeyError:
+            except KeyError as e:
                 output.append(None)
-                error_n = n
+                if error is None:
+                    error = e
 
-        if all(o is None for o in output) and error_n is not None:
-            # Raises KeyError
-            return partial_index(error_n, idx)
+        if all(o is None for o in output) and error is not None:
+            raise error
+        if error is not None:
+            error.__traceback__ = None
         return output
 
     # join results of multiple indices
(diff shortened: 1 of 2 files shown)


ottonemo:
While this is indeed a proper fix we are losing traceback information which could hide a bug in history. Additionally, it might be a good idea to re-think the way history processes indices, maybe we can do without recursion and exception propagation and avoid this situation altogether.

thomasjpfan:
I updated this PR by refactoring how `history` handles `KeyError`s. Instead of storing the exception, it uses `None` as a placeholder.

benjamin-work:
Just an update, I looked at this PR and also verified that it solves the initial issue. It will, however, take some more time to understand the solution and its implications, as well as thinking about whether a more fundamental solution is necessary.

thomasjpfan:
I reverted this PR to a commit where the exception is not stored. There is a little more overhead with calling `partial_index` again just to raise the exception. I think this is acceptable, since it removes the need to keep the exception in memory.

thomasjpfan:
Closed in favor of https://github.com/dnouri/skorch/pull/312

Issue #305, opened 2018-08-06
Recommended way to implement a custom dataset with data augmentation?


taketwo:
So far I haven't used data augmentation in my experiments with skorch, but now I need to. Therefore, I am wondering what is the canonical way to do so.

With pure PyTorch, I'd subclass `Dataset` to implement parsing/loading of data, and then create a `DataLoader` with needed transforms.

skorch hides the data loader from me, so I have no choice but to couple augmentation with the dataset parsing/loading logic. The two options I see are: 1) subclass PyTorch `Dataset` and implement everything inside `__getitem__` 2) subclass skorch `Dataset` and keep augmentation separated in `transform`.

I suppose the second option is the intended way to go? If so, then there seem to be a bug in skorch, [this](https://github.com/dnouri/skorch/blob/master/skorch/dataset.py#L58) function:

```python
def uses_placeholder_y(ds):
    if isinstance(ds, torch.utils.data.Subset):
        return uses_placeholder_y(ds.dataset)
    return isinstance(ds, Dataset) and ds.y is None
```

should not take existence of the `y` field for granted since the custom dataset may store target in a different field.

Stepping back a bit, did you consider somehow providing a possibility to set transform function for the data loader that skorch creates internally? (Such that the augmentation logic is not mixed with the dataset logic).

benjamin-work:
Well you can change the `Dataset` and `Dataloader` used by `NeuralNet` as long as some conditions hold (notably `__getitem__` returning `X, y`). As far as I can tell, this is compatible with the standard way of doing image augmentation in torchvision.

For the image transformation steps themselves, torchvision provides the `transform` parameter that you can couple with `transforms.Compose` to chain multiple transformations together.

Personally, I would only use these facilities to apply random image transformations. Everything that is applied only once per image I would move to an sklearn `Pipeline` with the `NeuralNet` only as the last step.

Why do I prefer this? First of all it's quicker because the transformation steps are not repeated for each epoch. But more importantly, I try to stay in "numpy world" as much as possible and use general purpose tools for image processing (skimage, opencv2). Those are usually more feature complete and more general. When later I want to test the same model in another framework, I can re-use most of my image processing pipeline.

If, on the other hand, you are certain to only ever use pytorch and/or your data is too big for memory, you should go full torchvision.

> there seem to be a bug in skorch, this function:

True, this should deal with the case of `ds.y` not existing. Would you be so kind as to open a new issue for this?

taketwo:
[...]

PR #311, opened 2018-08-10
uses_placeholder_y should not require existence of "y" field

taketwo:
Fixes a problem mentioned #305.


benjamin-work:
Thank you.

PR #312, opened 2018-08-15
Rewrite History to not use any recursion.

benjamin-work:
Instead unroll the successive indexing steps and perform them backwards, i.e. starting with batches and following with epochs.

This PR solves #306 and is an alternative to #310

### DIFF
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -2,61 +2,58 @@
 
 
 # pylint: disable=invalid-name
-class _missingno:
-    def __init__(self, e):
-        self.e = e
-
-    def __repr__(self):
-        return 'missingno'
+class _none:
+    """Special placeholder since ``None`` is a valid value."""
 
 
-def _incomplete_mapper(x):
-    for xs in x:
-        # pylint: disable=unidiomatic-typecheck
-        if type(xs) is _missingno:
-            return xs
-    return x
+def _not_none(items):
+    """Whether the item is a placeholder or contains a placeholder."""
+    if not isinstance(items, (tuple, list)):
+        items = (items,)
+    return all(item is not _none for item in items)
 
 
-# pylint: disable=missing-docstring
-def partial_index(l, idx):
-    needs_unrolling = (
-        isinstance(l, list) and len(l) > 0 and isinstance(l[0], list))
-    types = int, tuple, list, slice
-    needs_indirection = isinstance(l, list) and not isinstance(idx, types)
-
-    if needs_unrolling or needs_indirection:
-        return [partial_index(n, idx) for n in l]
-
-    # join results of multiple indices
-    if isinstance(idx, (tuple, list)):
-        zz = [partial_index(l, n) for n in idx]
-        if isinstance(l, list):
-            total_join = zip(*zz)
-            inner_join = list(map(_incomplete_mapper, total_join))
-        else:
-            total_join = tuple(zz)
-            inner_join = _incomplete_mapper(total_join)
-        return inner_join
-
-    try:
-        return l[idx]
-    except KeyError as e:
-        return _missingno(e)
+def _filter_none(items):
+    """Filter special placeholder value, preserves sequence type."""
+    type_ = list if isinstance(items, list) else tuple
+    return type_(filter(_not_none, items))
 
 
-# pylint: disable=missing-docstring
-def filter_missing(x):
-    if isinstance(x, list):
-        children = [filter_missing(n) for n in x]
-        # pylint: disable=unidiomatic-typecheck
-        filtered = list(filter(lambda x: type(x) != _missingno, children))
+def _getitem(item, i):
+    """Extract value or values from dicts.
 
-        if children and not filtered:
-            # pylint: disable=unidiomatic-typecheck
-            return next(filter(lambda x: type(x) == _missingno, children))
-        return filtered
-    return x
+    Covers the case of a single key or multiple keys. If not found,
+    return placeholders instead.
+
+    """
+    if not isinstance(i, (tuple, list)):
+        return item.get(i, _none)
+    type_ = list if isinstance(item, list) else tuple
+    return type_(item.get(j, _none) for j in i)
+
+
+def _unpack_index(i):
+    """Unpack index and return exactly four elements.
+
+    If index is more shallow than 4, return None for trailing
+    dimensions.
+
+    """
+    # fill trailing indices with None
+    i_e, k_e, i_b, k_b = i + tuple([None] * (4 - len(i)))
+
+    # handle special case of
+    # history[j, 'batches', somekey]
+    # which should really be
+    # history[j, 'batches', :, somekey]
+    if i_b is not None and not isinstance(i_b, (int, slice)):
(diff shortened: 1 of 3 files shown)


benjamin-work:
As discussed in #310 I put a little more thought into a more fundamental solution to rewriting `history.__getitem__` and came up with this solution.

My goal was to avoid recursion and instead "unroll" the consecutive indexing arguments, since we do know what each level of indexing should be. With this flat implementation, there is no more problem with circular references in tracebacks.

I introduces the assumption that if someone indexes history beyond the second level, the second level must be `'batches'`, i.e. this is now forbidden:

`history[:, 'my-list', 3]`

During this work, I discovered that in several places in the code, we call history like this:

`history[:, 'batches', 'some-key']`

The problem is with the 3rd level, `some-key`; the third level should be controlling the batch-level index, with the 4th level being responsible for the batch-level key. I made my implementation work with this call but would argue that it is wrong and we should not allow this.

Please discuss whether you like this solution or rather #310 or something different altogether.

skorch/history.py
```diff
@@ -2,61 +2,58 @@
...
+
+
+def _unpack_index(i):
+    """Unpack index and return exactly four elements.
+
+    If index is more shallow than 4, return None for trailing
+    dimensions.
+
+    """
+    # fill trailing indices with None
```
taketwo: Maybe assert that `len(i) <= 4`?

skorch/history.py
```diff
@@ -145,24 +143,68 @@ def record_batch(self, attr, value):
...
+        i_e, k_e, i_b, k_b = _unpack_index(i)
+        keyerror_msg = "Key '{}' was not found in history."
+        batches_unpacked = False
+
+        if i_b is not None and k_e != 'batches':
+            raise KeyError("History indexing beyond the 2nd level is "
+                           "only possible if key 'batches' is used, "
+                           "found key '{}'.".format(k_e))
+
+        items = self.to_list()
```
taketwo: This operation is virtually for free, right?

skorch/history.py
benjamin-work: yes, maybe we can check the length here, but I would rather raise a KeyError.

skorch/history.py
benjamin-work: This should create a shallow copy so almost no overhead.

However, we could maybe make sure that this is not slower than the previous implementation. Running `examples/benchmarks/mnist.py` does not seem to reveal a noticeable increase in skorch overhead.

skorch/history.py
taketwo: Yeah, `KeyError` would be cleaner than an assertion I suppose. And definitely better than `ValueError: too many values to unpack` as it is right now.

benjamin-work:
I performed a small benchmark on history: https://gist.github.com/benjamin-work/9a4a897fa1d8ccdcbc1b167e7d9e6668

The results for 5 runs each

* current master: 2.99069, 3.00728, 3.63035, 3.56770, 3.95994
* for #310: 3.12510, 2.92873, 2.65750, 2.58816, 2.89876
* for this PR: 2.14163, 2.12535, 1.96851, 2.03332, 1.75482

So it seems that this implementation is the fastest (maybe less function call overhead). But overall, each implementation is probably fast enough for this never to matter.

skorch/history.py
benjamin-work: This now raises a `KeyError`.

thomasjpfan:
Overall, I prefer this non-recursion approach over that previous implementation or my #310 fix.

I think `history[:, 'batches', 'some-key']` should not be supported at all. It's not documented in `History`, so it should be safe to remove. There are only a few places where it is used in tests:

https://github.com/dnouri/skorch/blob/16d7e52c7e2832f286e8db1d4aa6458d31f3491a/skorch/tests/test_dataset.py#L359

https://github.com/dnouri/skorch/blob/16d7e52c7e2832f286e8db1d4aa6458d31f3491a/skorch/tests/test_history.py#L133

https://github.com/dnouri/skorch/blob/16d7e52c7e2832f286e8db1d4aa6458d31f3491a/skorch/tests/test_net.py#L1393-L1394

benjamin-work:
> I think history[:, 'batches', 'some-key'] should not be supported at all. It's not documented in History, so it should be safe to remove. There are only a few places where it is used in tests

It may be used elsewhere, we can't be sure, so I would rather deprecate it instead of removing it completely.

@ottonemo what is your opinion on the rewriting `History`?

taketwo:
So are there any objections to merging this?

benjamin-work:
We were a little more busy than usual the last days and @ottonemo wanted to have a thorough look at this to really get it right.

skorch/history.py
```diff
@@ -145,24 +148,68 @@ def record_batch(self, attr, value):
...
+            ]
+            # get rid of empty batches
+            items = [b for b in items if b not in (_none, [], ())]
+            if not _filter_none(items):
+                # all rows contained _none or were empty
+                raise KeyError(keyerror_msg.format(k_b))
+
+        # extract the batches, but only if not already done
+        if k_e is not None:
+            if not batches_unpacked:
```
ottonemo: I'm wondering if it wouldn't be clearer to re-formulate the condition to

```python
if k_e is not None and i_b is None:
```

and get rid of `batches_unpacked` to indicate that we are handling the case 
`history[i_e, k_e]` and not `history[:, 'batches', i_b, k_b]`.

skorch/history.py
```diff
@@ -145,24 +148,68 @@ def record_batch(self, attr, value):
...
+                else _getitem(batches, k_b)
+                for batches in items
+            ]
+            # get rid of empty batches
+            items = [b for b in items if b not in (_none, [], ())]
+            if not _filter_none(items):
+                # all rows contained _none or were empty
+                raise KeyError(keyerror_msg.format(k_b))
+
+        # extract the batches, but only if not already done
```
ottonemo: Not sure about the statement 'extract batches' since we are collecting on epoch level, e.g. `history[:, 'train_loss']` which is not batch-level at all. Maybe better: extract epoch values.

skorch/history.py
```diff
@@ -145,24 +148,68 @@ def record_batch(self, attr, value):
...
+        batches_unpacked = False
+
+        if i_b is not None and k_e != 'batches':
+            raise KeyError("History indexing beyond the 2nd level is "
+                           "only possible if key 'batches' is used, "
+                           "found key '{}'.".format(k_e))
+
+        items = self.to_list()
+
+        # extract indices of batches
```
ottonemo: I would like to have a hint of what portion is handled in each section, e.g. a comment like

```python
# extract indices of batches
# handles: items = history[..., k_e, i_b]

[...]

# extract keys of batches
# handles: items = history[..., k_e, i_b][k_b]

[...]

# extract the batches, but only if not already done 
# handles: items = history[..., k_e]

[...]

# extract the epochs 
# handles: items = history[i_e, ..., ..., ...]
```

benjamin-work:
@ottonemo I addressed the comments and added a benchmark script. Please review again.

PR #314, opened 2018-08-18
RFC: Creates batch_idx on first run

thomasjpfan:
Refactors `_get_batch_idx` to keep track of the batch index.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -81,6 +81,7 @@ class LRScheduler(Callback):
     def initialize(self):
         self.policy_ = self._get_policy_cls()
         self.lr_scheduler_ = None
+        self.batch_idx_ = 0
         return self
 
     def _get_policy_cls(self):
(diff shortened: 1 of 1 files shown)


benjamin-work:
I agree that the existing method to count the batch index is flaky since it assumes that all epochs have the same number of batches.

With your proposed change, you would increase the count and return the index. I would prefer such a method not to have a side effect, since it assumes that `_get_batch_idx` is only called once per batch. I thus have two alternative proposals:

* Set a `batch_idx_` attribute during `initialize`, increment it during `on_batch_end`, get rid of `_get_batch_idx`.
* Count the number of batches (e.g. `len(list(flatten(net.history[:, 'batches'])))`).

I can think of one way how those two proposals differ, though I'm not sure if it ever happens: When a user de-activates the LR scheduler for some epochs during training (which is possible with `set_params`), the first proposal would only count the batches during which the scheduler was active, the second would count all batches.

thomasjpfan:
I went with the first approach. This was to avoid the runtime cost of calling `len` after every batch.

benjamin-work:
Thanks.

PR #315, opened 2018-08-18
Fix a number issues with ProgressBar

taketwo:
This PR fixes several issues that I discovered while working with `ProgressBar`. See individual commit messages for details.

I would also propose to change the default `batches_per_epoch` mode in `ProgressBar` to "auto". Is there any reason against this?

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -281,7 +281,7 @@ class ProgressBar(Callback):
 
     # pylint: disable=attribute-defined-outside-init
     def on_batch_end(self, net, **kwargs):
-        self.pbar.set_postfix(self._get_postfix_dict(net))
+        self.pbar.set_postfix(self._get_postfix_dict(net), refresh=False)
         self.pbar.update()
 
     # pylint: disable=attribute-defined-outside-init, arguments-differ
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -204,24 +204,37 @@ class PrintLog(Callback):
 
 
 class ProgressBar(Callback):
-    """Display a progress bar for each epoch including duration, estimated
-    remaining time and user-defined metrics.
+    """Display a progress bar for each epoch.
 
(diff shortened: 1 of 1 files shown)


benjamin-work:
I haven't worked much with tqdm or `ProgressBar`, so I will trust you here that the changes are sensible. I guess that showing the impact of the changes through a unit test is difficult.

If you believe that `'auto'` is the better choice here, you can go ahead and change it. I assume that `'count'` is a little bit safer when non-conventional datasets are used?

taketwo:
> I guess that showing the impact of the changes through a unit test is difficult.

For the first two commits, yes. But for the third one it would be nice to have a test, because the "auto" mode has been broken for unknown period of time by now--and no one noticed. If you have any suggestions, I can try to implement them.

> If you believe that 'auto' is the better choice here, you can go ahead and change it. 

The "auto" mode requires that the dataset length and the training batch size are known and stay fixed throughout an epoch. I think this is a very mild and realistic assumption. In fact, if it does not hold, I don't see any point in using a progress bar in the first place.

>  I assume that 'count' is a little bit safer when non-conventional datasets are used?

Well, at least in one way "auto" is even safer than "count". It does not require that the length of the dataset and/or batch size stay the same throughout the training, because the number of batches is estimated anew before every epoch. Contrary to that, in the "count" mode the number of batches is empirically computed once and is assumed to not change anymore.

benjamin-work:
I put a little bit more thought into this and came to the conclusion that `ProgressBar` probably needs some minor refactoring. I opened [an issue](https://github.com/dnouri/skorch/issues/319) for this. For this PR, I would say we're good.

What I meant with `'count'` being more robust is if you somehow get a `DataLoader` whose batch sizes differs from `net.batch_size` or `net.iterator_{train,valid}__batch_size` but if anyone does this, they shouldn't be surprised when the progress bar is confused.

It's your call if you'd like to change the default.

taketwo:
I switched the default to "auto" and extended the docstring.

benjamin-work:
Great, thank you for this.

PR #318, opened 2018-08-21
MAINT: make sure copy links weights and optimizer

### LINKED ISSUE #317: module weights don't update after copy.deepcopy
When a model is copied with `copy.deepcopy`, the weights are frozen and don't update when `fit` is called.

Here's an example:

``` python
import torch.nn as nn
import torch
from skorch import NeuralNetRegressor
import copy
import toolz

def get_data(n, d):
    X, y = make_regression(n_features=d, n_samples=n)
    y -= y.min()
    y /= y.max()
    y = y.reshape(-1, 1).astype('float32')

    X = X.astype('float32')
    return X, y

X, y = get_data(1000, 28 * 28)

class Simple(nn.Module):
    def __init__(self):
        super().__init__()
        self.simple = nn.Sequential(
            nn.Linear(28 * 28, 1),
            nn.Sigmoid(),
        )
        
    def forward(self, x):
        return self.simple(x)
n1 = NeuralNetRegressor(
    module=Simple,
).initialize()

def check(net, name=''):
    p = toolz.first(net.module_.parameters()).detach().numpy()
    print(name + ' first params =', p.flat[:3])

def assert_all_params_equal(net1, net2):
    for p1, p2 in zip(net1.module_.parameters(), net2.module_.parameters()):
        assert p1.requires_grad
        assert p2.requires_grad
        assert torch.all(p1 == p2)
    
check(n1, name="n1")
n1.partial_fit(X, y)
check(n1, name="n1")

n2 = copy.deepcopy(n1)
check(n2, name="n2")
assert_all_params_equal(n1, n2)

n3 = n2.partial_fit(X, y[::-1])

check(n1, name="n1")
check(n2, name="n2")
check(n3, name="n3")

assert_all_params_equal(n2, n3)
assert_all_params_equal(n1, n2)
```

which, when run, outputs this:

```
n1 first params = [-0.03383204 -0.00945246  0.03146857]
  epoch    train_loss    valid_loss     dur
-------  ------------  ------------  ------
      1        0.0415        0.0529  0.0245
      2        0.0404        0.0523  0.0251
      3        0.0394        0.0517  0.0252
n1 first params = [-0.03349511 -0.00889289  0.02929294]
n2 first params = [-0.03349511 -0.00889289  0.02929294]
      4        0.0431        0.0487  0.0250
      5        0.0431        0.0487  0.0246
      6        0.0431        0.0487  0.0247
n1 first params = [-0.03349511 -0.00889289  0.02929294]
n2 first params = [-0.03349511 -0.00889289  0.02929294]
n3 first params = [-0.03349511 -0.00889289  0.02929294]
```

stsievert:
Closes #317.

This needs review, hence the WIP label. I don't use `state_dict` as mentioned in https://github.com/dnouri/skorch/issues/317#issuecomment-414629936, and I coded to the test.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1325,8 +1325,11 @@ class NeuralNet(object):
                 device = device.type
             return device.startswith('cuda')
 
-        disable_cuda = uses_cuda(state['device']) and not torch.cuda.is_available()
-        load_kwargs = {} if not disable_cuda else {'map_location': lambda store, loc: store}
+        disable_cuda = (uses_cuda(state['device']) and
+                        not torch.cuda.is_available())
+        load_kwargs = {}
+        if disable_cuda:
+            load_kwargs = {'map_location': lambda store, loc: store}
 
         with tempfile.SpooledTemporaryFile() as f:
             f.write(state['cuda_dependent_attributes_'])
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1307,13 +1307,15 @@ class NeuralNet(object):
 
     def __getstate__(self):
         state = self.__dict__.copy()
+        cuda_attrs = {}
         for key in self.cuda_dependent_attributes_:
             if key in state:
                 val = state.pop(key)
-                with tempfile.SpooledTemporaryFile() as f:
-                    torch.save(val, f)
-                    f.seek(0)
-                    state[key] = f.read()
+                cuda_attrs[key] = val
+        with tempfile.SpooledTemporaryFile() as f:
+            torch.save(cuda_attrs, f)
+            f.seek(0)
+            state['cuda_dependent_attributes_'] = f.read()
 
         return state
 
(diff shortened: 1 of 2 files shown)


benjamin-work:
Thank you for taking on this issue.

The fix looks sound. The use of `state_dict` could help with portability but that is another issue.

There are some adjustments required for the unit test:

* As it is right now, it succeeds even without the fix. The reason is that you call `fit` the second time instead of `partial_fit` (always first check whether the test fails before writing the bug fix).
* The test name should be more explanatory.
* The unit test definitely needs more explanations in the form of comments, e.g. it is not restricted to deep copying but would also occur with pickle. You can also link to the issue in a comment.
* Finally, it is not quite clear whether all of the asserts are needed. Something like this directly after the deepcopy already reveals the bug:

```
        # check that the optimizer references the same parameters as used by 
        # the module
        for p0, p1 in zip(
            n2.module_.parameters(), n2.optimizer_.param_groups[0]['params']
        ):
            assert p0 is p1
```
But I also see the value in comparing the parameters after the fit call, so leave that in as well. The comparison of `train_loss`, on the other hand, seems pointless -- if the module parameters are updated, can the loss ever the same (and if it were, the issue would certainly lie somewhere else)?

stsievert:
Comments addressed. I've added a lot more comments and this seems like a better test.

> The comparison of train_loss, on the other hand, seems pointless -- if the module parameters are updated, can the loss ever the same (and if it were, the issue would certainly lie somewhere else)?

This is a user-facing test, and the reason I filed this issue and PR. I agree that it's redundant, but if that `assert` fails users will complain, and it's a good sanity check.

skorch/tests/test_net.py
```diff
@@ -125,6 +126,51 @@ def net_pickleable(self, net_fit):
...
+            raise ValueError
+
+        # Test to make sure the parameters got copied correctly
+        close = [torch.allclose(p1, p2)
+                 for p1, p2 in zip(n1.module_.parameters(),
+                                   n2.module_.parameters())]
+        assert all(close)
+
+        # make sure the parameters change
+        # cal this twice to make sure history updates below
```
benjamin-work: Typo "cal"; also, why are 2 calls needed? When I comment one out, the test still passes.

skorch/net.py
```diff
@@ -1323,24 +1325,26 @@ def uses_cuda(device):
                 device = device.type
             return device.startswith('cuda')
 
-        disable_cuda = False
+        disable_cuda = uses_cuda(state['device']) and not torch.cuda.is_available()
+        load_kwargs = {} if not disable_cuda else {'map_location': lambda store, loc: store}
```
benjamin-work: This line is too long, could you please break it?

skorch/tests/test_net.py
stsievert: Thanks for this. I can also pass `epochs=2` to `partial_fit`, and I explained more why in the comments.

benjamin-work:
Thank you very much for this contribution.

Issue #316, opened 2018-08-19
Refactor notebooks to be able to run completely in Google's Colaboratory


thomasjpfan:
It could be beneficial to allow notebooks to  run completely in Colab. Sometimes the notebooks would have to be updated to be able to run in Colab. For example, here is an updated version of Transfer Learning that runs on Colab:

https://colab.research.google.com/github/thomasjpfan/skorch/blob/colab-basic-usage/notebooks/Transfer_Learning.ipynb

To run on a GPU, a user would need to navigate: `Runtime -> Change runtime type -> Hardware accelerator -> GPU -> Save`.

taketwo:
Speaking of skorch tutorials/notebooks, I wanted o propose moving them to Sphinx-Gallery, the same as PyTorch. What do you think?

benjamin-work:
In general, we should put more effort into facilitating learning and discovering more advanced features, so these ideas sound great.

> Sometimes the notebooks would have to be updated to be able to run in Colab

> Speaking of skorch tutorials/notebooks, I wanted o propose moving them to Sphinx-Gallery

Could you both quickly summarize the changes required for that?

thomasjpfan:
For colab to work we need an additional cell to:

1. Install pytorch, skorch, and other dependencies.
2. Download data if needed.

In my example above, I needed to add the following cell at the top:

```bash
!pip install torch torchvision skorch pillow==4.1.1
!mkdir -p datasets
!wget -nc --no-check-certificate https://download.pytorch.org/tutorial/hymenoptera_data.zip -P datasets
!unzip -u datasets/hymenoptera_data.zip -d datasets
```

Tutorials that do not need to download data, just needs to `pip install` dependencies.

taketwo:
Regarding Sphinx-Gallery, the tutorial code will need to be converted from Notebook format to normal Python scripts with prose written as rst markup in comments.

For example, here is the source code for one of PyTorch tutorials: [code](https://github.com/pytorch/tutorials/blob/master/beginner_source/blitz/cifar10_tutorial.py), and this is how it appears on their website: [html](https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html).

[...]

PR #321, opened 2018-08-23
DOC: Update to run in colab

thomasjpfan:
Addresses some of #316

You can see how this works here: https://colab.research.google.com/github/thomasjpfan/skorch/blob/transfer-learning-colab/notebooks/Transfer_Learning.ipynb

### DIFF
--- a/docs/user/tutorials.rst
+++ b/docs/user/tutorials.rst
@@ -11,7 +11,7 @@ The following are examples and notebooks on how to use skorch.
 
 * `Benchmarks skorch vs pure PyTorch <https://github.com/dnouri/skorch/blob/master/examples/benchmarks/mnist.py>`_ - Compares the performance of skorch and using pure PyTorch on MNIST.
 
-* `Transfer Learning with skorch <https://github.com/dnouri/skorch/blob/master/notebooks/Transfer_Learning.ipynb>`_ - Train a neutral network using transfer learning with skorch.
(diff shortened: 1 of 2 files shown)


benjamin-work:
As a colab newbie, do I have to do anything special before running the notebook? The imports fail for me because torch is not installed. The data download fails because there is

> No such file or directory: 'datasets/hymenoptera_data'

thomasjpfan:
I forgot to add the download cell. I updated the notebook to install the dependencies and `wget` the data files.

benjamin-work:
Now I could progress further but got this cuda error:

> RuntimeError: cuda runtime error (35) : CUDA driver version is insufficient for CUDA runtime version at /pytorch/aten/src/THC/THCGeneral.cpp:74

Also could you make the download cell optional? Ideally, it should execute depending on the environment but I didn't see any env variable that signals that colab is used. Perhaps `DATALAB_ENV=GCE` can be used as a proxy?

thomasjpfan:
I updated the PR by using `DATALAB_ENV` as a flag on the download cell.

I am unable to run into that error. Which cell do you run into the issue?

thomasjpfan:
Checking the cuda version tensorflow is built on (in colab):

```bash
!cat /usr/local/lib/python3.6/dist-packages/tensorflow/include/external/local_config_cuda/cuda/cuda/cuda_config.h | grep TF_CUDA_VERSION
#define TF_CUDA_VERSION "8.0"
```

Does changing the `pip install...` line to the following fix your issue?

```
! [ "$DATALAB_ENV" = "GCE" ] && pip install http://download.pytorch.org/whl/cu80/torch-0.4.1-cp36-cp36m-linux_x86_64.whl torchvision skorch pillow==4.1.1
```

benjamin-work:
I did not enable GPU, that may have caused the error.

Regarding the colab cell, could you add a comment to explain what this is and why it is needed?

benjamin-work:
Thank you, nice addition.

PR #323, opened 2018-08-23
Replace yield_fixture with fixture

taketwo:
Standard `fixture` now supports yielding. `yield_fixture` has been deprecated since pytest 3.

See: https://docs.pytest.org/en/latest/yieldfixture.html


benjamin-work:
Interesting, I didn't know about this. I would have expected `DeprecationWarning`s but oh well.

Thanks for the PR.

PR #326, opened 2018-08-26
ProgressBar refactoring

### LINKED ISSUE #319: Refactor ProgressBar
This class could benefit from some small refactorings.

For instance, it overrides the `batches_per_epoch` attribute when it is `'count'`, which contradicts the policy to never change init params. It is also surprising that `'count'` triggers the estimation only once per fit call, whereas `'auto'` does it once per epoch.

This refactoring should also include more rigorous unit tests that control the inferred batch size in `'count'` and `'auto'` mode. To facilitate this unit test, there could be a public function that derives the batch size from a dataset:

```
def estimate_batches_per_epoch(dataset, batches_per_epoch='auto'):
    if isinstance(batches_per_epoch, int):
        return batches_per_epoch
    if batches_per_epoch == 'count':
        return None
    ...
```

taketwo:
Fixes #319.

I decided to not follow the proposed approach. For testing I patch `tqdm` and check that the expected `total` (i.e. number of epochs) is passed to its constructor.

As for the implementation of "count" mode, I switched to checking the length of the "batches" list in the history. This will have and added benefit that "count" will work as expected when resuming training from saved model/history.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -216,8 +216,8 @@ class ProgressBar(Callback):
     this number is determined automatically using the dataset length
     and the batch size. If this heuristic does not work for some
     reason, you may either specify the number of batches explicitly
-    or let the ``ProgressBar`` determine it empirically by counting
-    the number of batches in the very first epoch.
+    or let the ``ProgressBar`` count the actual number of batches in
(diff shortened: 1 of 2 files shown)


benjamin-work:
Thank you very much.

PR #327, opened 2018-08-27
Add license to manifest

dhirschfeld:


benjamin-work:
Thanks for the PR, I will gladly add this. Just for my understanding, could you tell me why this is needed?

dhirschfeld:
It ensures that the tarball that you upload to pypi contains the license file. conda-forge insists on having a license file packaged as part of the build (see linked issue above). Plus, I think it's generally considered good practice to distribute the license with the source.

benjamin-work:
Okay, that is what I wanted to know. Thank you.

PR #329, opened 2018-08-31
Remove superfluous call to is_dataset()

taketwo:
This test is performed inside `get_dataset()` anyway.


ottonemo:
Thanks, seems reasonable :)

PR #331, opened 2018-09-04
BUG: Uses OrderedDict for callbacks

### LINKED ISSUE #330: Bug in callback order when using Python < 3.6
The offending line is here:

https://github.com/dnouri/skorch/blob/1177be7404ece996d0e6b63982f39b65fe1811e3/skorch/net.py#L366

When dictionary order is not preserved, order of the callbacks may change.

thomasjpfan:
Fixes #330

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -2,6 +2,7 @@
 
 import fnmatch
 from itertools import chain
+from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


ottonemo:
Thank you, looks good :)

Issue #325, opened 2018-08-26
Working with PackedSequence issues


Naffer11:
Firstly I'd like to thank you for the great job you have done with skorch - I've found it very useful!
However, I have 2 observations.

Lately, while I've been trying to work with sequence data with a varying length I decided to use PackedSequence data from torch.
The first thing that is easy to fix is the to_tensor function from skorch.utility - although it's docstring states it supports PackedSequence, as this class is a namedtuple under the hood, it is also an instance of a tuple, and as so the PackedSequence is being handled as a tuple.

```
import torch

seq = [torch.rand(4,1) for _ in range(5)]
pack_seq = torch.nn.utils.rnn.pack_sequence(seq)

print(isinstance(pack_seq, tuple))
print(isinstance(pack_seq, torch.nn.utils.rnn.PackedSequence)
```
This can be easily fixed moving more specific condition of being a PackedSequence before the condition of being a tuple.

The second thing is more of a thought on handling a varying length data by skorch - the most convenient way I found to handle it by Pytorch is by using a list of Tensors and then converting it to mentioned earlier PackedSequence. The problem arises with the all the internal skorch lenght checking functions - they rise error when the data have inconsistent lenght, what is an expected situation sometimes like when dealing with varying length sequences. I managed to overwrite all the places where it is done ( in Dataset __init__(), in CVSplit __call__() and NeuralNet in fit_loop()) because they all call skorch.utils.get_len() the function which unpacks the data to check the lengths - as this function is defined in the package directory there is no other way than overriding all the functions that call it to change its default behaviour.
I was wondering do you think it would be possible to:
a) give an option to pass on checking the length in all listed structures by one parameter (something like check_len =False)
b) enable the user to pass a function responsible for checking the lenghts of passed structures
I think the second option is much more elegant, but I am not sure how it could be best done.

Hope I described clearly my points if anything is unclear I can provide more details if needed.

BenjaminBossan:
Thank you for using skorch.

Regarding your first point, I will look at it on Monday. If you think you know how to fix that, I also invite you to make a PR.

Regarding your second point, do I get it right that you pass your proper data and your array of lengths as a list and then get problems because the list does not have the same length as your `y` data? If so, I would recommend you pass your data as a dictionary:

```
class MyModule(nn.Module):
    ...
    def forward(self, X, lengths):  # dictionary keys are passed as keyword arguments
        ....

X = {'X': data, 'lenghts': lengths}
```

[...]

PR #335, opened 2018-09-07
Make to_tensor work correctly with PackedSequence

benjamin-work:
Also, add a bunch of tests for to_tensor.

Partly addresses #325.

### DIFF
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 import torch
 from torch.nn.utils.rnn import PackedSequence
+from torch.nn.utils.rnn import pack_padded_sequence
 
 from skorch.tests.conftest import pandas_installed
 
@@ -29,14 +30,10 @@ class TestToTensor:
         t = to_tensor(t, device='cpu')
         assert t.device.type == 'cpu'
 
-    x = torch.zeros((5, 3)).float()
-    y = torch.as_tensor([2, 2, 1])
-    z = np.arange(15).reshape(5, 3)
-
     def tensors_equal(self, x, y):
         """"Test that tensors in diverse containers are equal."""
         if isinstance(x, PackedSequence):
-            return x == y
+            return self.tensors_equal(x[0], y[0]) and self.tensors_equal(x[1], y[1])
 
         if isinstance(x, dict):
             return (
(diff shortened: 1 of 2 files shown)


skorch/tests/test_utils.py
```diff
@@ -28,6 +29,39 @@ def test_device_setting_cuda(self, to_tensor):
...
+    @pytest.mark.parametrize('X, expected', [
+        (x, x),
+        (y, y),
+        (z, torch.as_tensor(z)),
+        ({'a': x, 'b': y, 'c': z}, {'a': x, 'b': y, 'c': torch.as_tensor(z)}),
+        (torch.as_tensor(55), torch.as_tensor(55)),
+        (PackedSequence(x, y), PackedSequence(x, y)),
+    ])
+    def test_tensor_conversion(self, to_tensor, X, expected):
+        result = to_tensor(X, 'cpu')
```
githubnemo: it might be worthwhile to make this an optional GPU test as well (if available), e.g.

```python
result = to_tensor(X, 'cpu')
# ... asserts ...
if not torch.cuda.is_available():
    return
result = to_tensor(X, 'cuda')
# ... asserts ...
```

benjamin-work:
@githubnemo Could you pls review again?

skorch/tests/test_utils.py
```diff
@@ -28,6 +30,72 @@ def test_device_setting_cuda(self, to_tensor):
...
+                    {'a': x, 'b': y, 'c': z},
+                    {'a': x, 'b': y, 'c': torch.as_tensor(z).to('cuda')}
+                ),
+                (torch.as_tensor(55), torch.as_tensor(55).to('cuda')),
+                (
+                    pack_padded_sequence(x, y),
+                    pack_padded_sequence(x, y).to('cuda')
+                ),
+        ]:
+            yield X, expected, device
```
githubnemo: can't you avoid code duplication by re-using the parameters? something like

```python
devices = ['cpu']
if torch.cuda.is_available():
    devices.append('cuda')
for dev in devices:
   for p in params:
      yield (p[0], p[1].to(dev), dev)
```

benjamin-work: This unfortunately doesn't work because some of the parameters don't support `.to(dev)` (list, tuple, dict).

I pondered quite a bit about this parameter thing but finally settled on something that is a little bit repetitive but I hope at least easy to understand.

PR #337, opened 2018-09-07
[MRG] Adds SetRequiresGradParams

### LINKED ISSUE #261: Document how to freeze layers

thomasjpfan:
Resolves #261.

This PR restricts the matching to only using `fnmatch`. I made this choice because `optimizer__param_groups` in `NeutralNet` works with wildcards the same way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ python:
 cache:
   apt: true
   timeout: 1000
-  directories:
-    - $HOME/miniconda/pkgs/
-before_cache:
-  - rm $HOME/miniconda/pkgs/*.tar.bz2
-  - rm $HOME/miniconda/pkgs/urls.txt
+#   directories:
+#     - $HOME/miniconda/pkgs/
+# before_cache:
+#   - rm $HOME/miniconda/pkgs/*.tar.bz2
+#   - rm $HOME/miniconda/pkgs/urls.txt
 install:
   - sudo apt-get update
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,23 @@ python:
 cache:
   apt: true
   timeout: 1000
-#   directories:
-#     - $HOME/miniconda/pkgs/
-# before_cache:
-#   - rm $HOME/miniconda/pkgs/*.tar.bz2
-#   - rm $HOME/miniconda/pkgs/urls.txt
+  directories:
+    - $HOME/cache
+before_cache:
+  - find ${HOME}/conda/pkgs/ -mindepth 1 -maxdepth 1 -type d -exec rm -r {} \;
 install:
   - sudo apt-get update
+  - mkdir -p ${HOME}/cache/pkgs
+  - "[ ! -f ${HOME}/cache/miniconda.sh ] && wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ${HOME}/cache/miniconda.sh || :"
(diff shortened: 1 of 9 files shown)


taketwo:
Perhaps I'm missing something important, but why does this need to be a callback? Why not just a helper function `set_requires_grad(net, pattern, requires_grad)`? I would understand the need if the callback also reverted its actions on train end, but this is not the case.

benjamin-work:
> why does this need to be a callback?

I agree that having this functionality in a callback is a little bit unintuitive. Personally, I would really prefer `NeuralNet(module__embeddings__weight__requires_grad=False)` to work, but oh well. The difference between your proposal and a callback is that the callback can be parametrized, meaning that I can for instance test in a grid search whether freezing some parameters is better or not.

Nonetheless, a function as you proposed it could be useful, so I wonder whether that functionality should be factored out and `SetRequiresGradParams` changed to use that function.

> I would understand the need if the callback also reverted its actions on train end

Do you have a use case for this? Should that be an option for the callback?

taketwo:
> Do you have a use case for this? Should that be an option for the callback?

Not really, this was just a speculative idea. But, now that you have mentioned grid search, maybe this is **the** use case? After trying one combination of frozen parameters you'd probably want to revert back before checking another one?

benjamin-work:
> After trying one combination of frozen parameters you'd probably want to revert back before checking another one?

This should not be necessary because `GridSearchCV` clones the model and only then calls `set_params`, meaning that only the clones are modified.

thomasjpfan:
`NeuralNet(module__embeddings__weight__requires_grad=False)` could work as long as its restricted to modifying parameters associated with `torch.nn.Parameter`. Parameters such as `cuda` or `requires_grad` will work as expected. Consider the following example:

```python
class MyModule(nn.Module):
    def __init__(self, in_features, out_features=123):
        self.linear = nn.Linear(in_features, out_features)
        self.sub_module = MySubmodule(in_features=out_features)
```

`in_features` and `out_features` would not be supported. The following would be supported:

1. `NeuralNet(module__sub_module__requires_grad=False)`
1. `NeuralNet(module__sub_module__cuda=False)`
1. `NeuralNet(module__linear__requires_grad=False)`
1. `NeuralNet(module__linear__weight__requires_grad=False)`

What do you think?

benjamin-work:
Yes, I thought about that possibility, mainly having a list of parameter names that are "okay" to be set without needing to re-initialize modules. However, from a user perspective, it would be quite hard to tell what works and what doesn't, and keeping that list up-to-date is certainly daunting.

Moreover, we may get into trouble because the user can name their parameters whatever they like:

```
class MyModule(nn.Module):
    def __init__(self, requires_grad=10):
        self.linear = MySubmodule(in_features=requires_grad)

NeuralNet(module__linear__requires_grad=20)
```

The last line would look totally fine for skorch but would actually not work as intended. Sure, that example is contrived, but who knows what other parameter names there are that are less clear-cut. In the end, we just cannot know if setting a parameter that way would work or not, and putting that burden on the user feels like too much a burden.

thomasjpfan:
I am placing this PR on hold till a decision is made in https://github.com/dnouri/skorch/issues/342.

Issue #342, opened 2018-09-11
Introduce a parameter initialization stage


ottonemo:
Since there is still discussion around how to properly and conveniently freeze parameters (#261, #337) I'd like to present a generic way of approaching this problem while avoiding callbacks (which may seem an unnatural choice for such an action, but that might also be personal preference) and opening up a number of modifications that were not easily possible.

The general idea is to introduce a parameter initialization stage that applies user-specified functions to certain parameters, e.g.:

```python
net = Net(
	MyModule,
	parameter_map={
		'linear[0-9].weight': skorch.helper.freeze_param,
		'linear[0-9].bias': skorch.helper.freeze_param,
		'rnn*.h2h': torch.init.xavier_normal_,
	}
	parameter_map__linear0__weight=torch.init.xavier_normal_,
)
```

this approach has the following benefits:

- works with both uninitialized and initialized modules
- compatible with grid search (`parameter_map__linear0__weight` notation)
- easily extensible to multiple optimizers (basically no change needed under the assumption of non-overlapping parameter sets between optimizers)
- allows easy test of different initialization schemes
- compatible with torch utility functions

while there are a few benefits it has the drawback of introducing a new specialized parameter that needs documentation and introduces a new workflow besides callbacks.

I'd like to hear your feedback.

Idea sketch:
```python
class _untrainable_marker: pass

def freeze_parameter(param):
	param.requires_grad = False
	return _untrainable_marker

def apply_parameter_map(self, map, named_parameters):
	for name, param in named_parameters:
		for map_name, map_fn in map.items():
			if not fnmatch.fnmatch(name, map_name):
				continue
			if map_fn(param) is _untrainable_marker:
				break
		else:
			continue
		yield name, param

class Net:
	def __init__(self, parameter_map=None, **kwargs):
		self.parameter_map = self._collect_parameter_map(parameter_map, kwargs)

	def collect_parameter_map(self, pmap, kwargs):
		for k, fn in self.get_params_for('parameter_map', kwargs).items():
			pmap[k.replace('__', '.')] = fn
		return pmap

	def initialize_parameters(self):
		self.trainable_parameters_ = apply_parameter_map(self.parameter_map, self.module_.named_parameters())

	def initialize_optimizer(self):
		self.optimizer_ = self.optimizer(self.trainable_parameters_)

	def initialize(self):
		self.init_callbacks()
		self.init_module()
		self.init_parameters()
		self.init_optimizer()

net = Net(
	MyModule,
[...]

PR #343, opened 2018-09-14
Implement PoC of ParamMapper from #342

ottonemo:
See #342 for a detailed discussion. Essentially, this provides a way to map arbitrary functions over parameters at a user-defined point in time.


githubnemo:
If this works properly I would suggest to deprecate `skorch.helper.filtered_optimizer` or at least `skorch.helper.filter_requires_grad`.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,131 @@ def _sink(self, text, verbose):
...
+    ...    if len(net.history) % 2 == 0:
+    ...        return skorch.utils.freeze_parameter
+    ...    else:
+    ...        return skorch.utils.unfreeze_parameter
+    >>> cb = ParamMapper('embedding*.weight', schedule=my_sched)
+    >>> net = Net(myModule, callbacks=[cb])
+
+    Parameters
+    ----------
+    patterns : str or callable or list
```
taketwo: It may be useful to clarify that these are matched in the Unix filetype style, not a proper regex.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,131 @@ def _sink(self, text, verbose):
...
+            map_fn(p)
+
+
+
+class Freezer(ParamMapper):
+    """Freeze matching parameters in the first epoch. You may
+    specify a specific point in time (either by epoch number or using a callable)
+    when the parameters are frozen using the ``at`` parameter.
+
+    See ``ParamMapper`` for details.
```
taketwo: Would be more convenient for user if this was clickable, i.e.  use 

```
:class:`.ParamMapper`
```

examples/benchmarks/freezing.py
```diff
@@ -0,0 +1,136 @@
...
+    print()
+
+    print(f'test_a: µ_rss = {dur_a[0]}, µ_valid_loss = {dur_a[1]}, µ_dur={dur_a[2]}')
+    print(f'test_a: µ_rss = {dur_b[0]}, µ_valid_loss = {dur_b[1]}, µ_dur={dur_b[2]}')
+
+    # valid losses must be identical
+    assert np.allclose(dur_a[1], dur_b[1])
+
+    # memory usage should be nearly identical (within 4MiB)
+    assert abs(dur_a[0] - dur_b[0]) < (4*1024**2)
```
benjamin-work: You could use `np.isclose` here is well with the `atol` parameter.

examples/benchmarks/freezing.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # duration should be nearly identical
+    assert abs(dur_a[2] - dur_b[2]) < 1e-3
```
benjamin-work: Again, `np.isclose` can help here. Also, the restriction is a little narrow, for me this test failed.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
         #  We do not want to be affected by verbosity if sink is not print
         if (self.sink is not print) or verbose:
             self.sink(text)
+
+
+
+class ParamMapper(Callback):
+    """Map arbitrary functions over module parameters filtered by pattern
+    matching. In the simplest case the function is only applied once at
```
benjamin-work: Perhaps add a paragraph after the first sentence.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+    --------
+    Initialize a layer on first epoch before the first training step:
+
+    >>> init = partial(torch.nn.init.uniform_, a=0, b=1)
+    >>> cb = ParamMapper('linear*.weight', at=1, fn=init)
+    >>> net = Net(myModule, callbacks=[cb])
+
+    Reset layer initialization if train loss reaches a certain value:
+
+    >>> at = lambda net: net.history[-1, 'train_loss'] > 1
```
benjamin-work: Does this fit the description that the function is triggered `if train loss reaches a certain value`?

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+    >>> net = Net(myModule, callbacks=[cb])
+
+    Parameters
+    ----------
+    patterns : str or callable or list
+      The pattern(s) to match parameter names against. Patterns are
+      UNIX globbing patterns as understood by :func:`~fnmatch.fnmatch`.
+      Patterns can also be callables which will get called with the
+      parameter name and are regarded as a match when the callable
+      returns a truthy value.
```
benjamin-work: You don't mention what `list` would do.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+    schedule : callable or None
+      If specified it determines how ``fn`` and ``at`` are used.
+      This function receives ``net``, ``at`` and ``fn`` and returns
+      the function to be applied on the matched parameters.
+
+    """
+    def __init__(self, patterns, fn=noop, at=1, schedule=None):
+        self.at = at
+        self.fn = fn
+        self.schedule = schedule or self._default_schedule
```
benjamin-work: The logic of ` or self._default_schedule` should also be moved to `initialize`.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+    at : int or callable
+      In case you specify an integer it represents the epoch number the
+      function ``fn`` is applied to the parameters, in case ``at`` is
+      a function it will receive ``net`` as parameter and the function
+      is applied to the parameter once ``at`` returns ``True``.
+
+    schedule : callable or None
+      If specified it determines how ``fn`` and ``at`` are used.
+      This function receives ``net``, ``at`` and ``fn`` and returns
+      the function to be applied on the matched parameters.
```
benjamin-work: I have to admit I don't see what use case you have in mind for `schedule`, especially since the example above ("periodically freeze and unfreeze ...") could also be achieved by `at`.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+        params = self.filter_parameters(self.patterns, params)
+        map_fn = self.schedule(net, self.at, self.fn)
+
+        for _, p in params:
+            map_fn(p)
+
+
+
+class Freezer(ParamMapper):
+    """Freeze matching parameters in the first epoch. You may
```
benjamin-work: "in" -> "at the start of"

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+    See :class:`.ParamMapper` for details.
+    """
+    def __init__(self, *args, **kwargs):
+        kwargs['at'] = kwargs.get('at', 1)
+        kwargs['fn'] = kwargs.get('fn', freeze_parameter)
+        super().__init__(*args, **kwargs)
+
+
+class Unfreezer(ParamMapper):
+    """Inverse operation of ``Freezer``."""
```
benjamin-work: I would rather see a non-referential description (especially since the `Freezer` doc already refers to `ParamMapper`).

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+        self.schedule = schedule or self._default_schedule
+        self.patterns = patterns
+
+    def initialize(self):
+        if not isinstance(self.patterns, (list, tuple)):
+            self.patterns = [self.patterns]
+
+        if isinstance(self.at, int):
+            self.at = partial(self._epoch_at, epoch=self.at)
+
```
benjamin-work: `initialize` should return `self`.

Also, I would like to see a check on `at` if it is `int`, because I can easily see someone setting `at=0`, which will just silently fail.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,145 @@ def _sink(self, text, verbose):
...
+
+class Initializer(ParamMapper):
+    """Apply any function on matching parameters in the first epoch.
+
+    Examples
+    --------
+
+    Use ``Initializer`` to initialize all dense layer weights with
+    values sampled from an uniform distribution on epoch begin of
```
benjamin-work: "to initialize ... values ... on epoch begin of the first epoch" -> "... on the beginning of ..."

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+        from skorch.utils import to_numpy
+
+        mod = classifier_module()
+
+        net = net_cls(
+            mod,
+            lr=0,
+            max_epochs=1,
+            callbacks=[
+                initializer('dense*.weight', partial(constant_, val=5)),
```
benjamin-work: I believe this pattern will fail now with the refactored tests because `classifier_module` now uses `nn.Sequential` internally.

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+            classifier_module,
+            max_epochs=2,
+            callbacks=[
+                freezer('dense*.weight'),
+                freezer('dense1.bias'),
+            ])
+
+        net.initialize()
+
+        assert net.module_.dense0.weight.requires_grad
```
benjamin-work: Same issue: `classifier_module` now uses `nn.Sequential`, hence param names changed.

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+            lr=5,
+            module__nonlin=tanh,
+            callbacks=[
+                freezer('dense*.weight'),
+                freezer('dense1.bias'),
+            ])
+
+        net.initialize()
+
+        assert net.module_.dense0.weight.requires_grad
```
benjamin-work: Same issue: `classifier_module` now uses `nn.Sequential`, hence param names changed.

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+                unfreezer('dense*.weight', at=2),
+                unfreezer('dense1.bias', at=2),
+            ])
+
+        net.initialize()
+
+        # epoch 1, freezing parameters
+        net.partial_fit(*classifier_data)
+
+        assert not net.module_.dense0.weight.requires_grad
```
benjamin-work: Same issue: `classifier_module` now uses `nn.Sequential`, hence param names changed.

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+                    schedule=schedule,
+                ),
+            ])
+
+        net.initialize()
+
+        # epoch 1, freezing parameters
+        net.partial_fit(*classifier_data)
+
+        assert not net.module_.dense0.weight.requires_grad
```
benjamin-work: Same issue: `classifier_module` now uses `nn.Sequential`, hence param names changed.

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+                                       classifier_data, freezer):
+        from skorch.utils import to_numpy
+        from torch import tanh
+
+        mod = classifier_module()
+        net = net_cls(
+            mod,
+            max_epochs=2,
+            lr=5,
+            module__nonlin=tanh,
```
benjamin-work: What is the point of `tanh` and generally, the difference to the previous test?

skorch/tests/callbacks/test_training.py
```diff
@@ -317,3 +317,276 @@ def test_wrong_threshold_mode(
...
+            ])
+
+        net.fit(*classifier_data)
+
+        assert np.allclose(to_numpy(net.module_.dense0.weight), 5)
+        assert np.allclose(to_numpy(net.module_.dense1.weight), 5)
+        assert np.allclose(to_numpy(net.module_.dense1.bias), 10)
+
+    def test_freezing_is_effective(self, net_cls, classifier_module,
+                                   classifier_data, freezer):
```
benjamin-work: This test has a lot of similarities with the following tests, can it be refactored to re-use more code?

skorch/callbacks/training.py
```diff
@@ -273,3 +278,155 @@ def _sink(self, text, verbose):
...
+
+    def initialize(self):
+        if not isinstance(self.patterns, (list, tuple)):
+            self.patterns = [self.patterns]
+
+        if isinstance(self.at, int):
+            self.at = partial(self._epoch_at, epoch=self.at)
+
+    def named_parameters(self, net):
+        return list(net.module_.named_parameters())
```
thomasjpfan: The following should work without needing to convert the generator to a list:
```python
def named_parameters(self, net):
    yield from net.module_.named_parameters()
```

benjamin-work: If `named_parameters` is only consumed once (which seems to be the case), a simple `return net.module_.parameters()` should be enough (or no special method at all).

skorch/tests/callbacks/test_training.py
ottonemo: The point is to test also how re-initialization of the module affects the behavior. There is a test that evaluates the functionality. I added a comment explaining just that.

skorch/callbacks/training.py
ottonemo: Pretty sure that it is necessary since we evaluate it more than once (for each pattern)

examples/benchmarks/freezing.py
```diff
@@ -0,0 +1,136 @@
...
+    for i in range(len(mod.seq) - 1):
+        assert not mod.seq[i].weight.requires_grad
+
+    net.fit(X, y)
+
+    for i in range(len(mod.seq) - 1):
+        assert not mod.seq[i].weight.requires_grad
+        assert not net.module_.seq[i].weight.requires_grad
+
+    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
```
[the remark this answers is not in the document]
ottonemo: I'm using fork now for each test so that every process has its own resource pool.

skorch/callbacks/training.py
ottonemo: Yes. The `at` function returns `True` as soon as `train_loss` reaches a certain value (i.e. `1`).

skorch/callbacks/training.py
ottonemo: I don't see how it can be achieved with `at`. You'd need to create two `ParamMapper` to solve this with `at`, e.g.

    ParamMapper(fn=freeze, at=lambda net: len(net.history) % 2 == 0)
    ParamMapper(fn=unfreeze, at=lambda net: len(net.history) % 2 == 1)

which is not very grid search-friendly.

skorch/callbacks/training.py
benjamin-work: Ah I see.

skorch/callbacks/training.py
benjamin-work: I would assume that `train_loss` decreases over time, so at the start, the condition is always true and later false. The description makes it sound like the condition is false at the start and becomes true over time.

skorch/callbacks/training.py
ottonemo: I'm not a big fan of repeating documentation and I don't see anything there is to add (it really is just the inverse operation of `Freeze`). I added a link to the implementation of `Freeze` so that navigating there is easier but I don't see what's there to describe.

skorch/callbacks/training.py
benjamin-work: My point is that I would like to see an example that uses `at` and `fn`, because otherwise, `schedule` seems to be pretty orthogonal to those two. If there is no conceivable case for that, we have to wonder why those parameters are passed. Or maybe the schedule functionality needs to be in another class, so that you have either `at` and `fn`, or `schedule`, but not both.

skorch/callbacks/training.py
benjamin-work: And I'm not a big fan of having to import `Freezer` to look up what `Unfreezer` does (when working in jupyter, for instance).

A simple 

```Unfreeze matching parameters at the start of the first epoch.

This performs the inverse operation of :class:`.Freezer`. For more information, look there.

```

should do.

skorch/callbacks/training.py
ottonemo: Since `schedule` is the generalisation of the static `at`/`fn` case I think it makes sense to have it here. While I can see that it is possible to keep the implementation but separate the current implementation in `StaticParamMapper` (supports `at`/`fn`) and `PeriodicParamMapper` (supports `schedule`), I don't think it adds much clarity.

The parameters on the other hand are, indeed, quite useless at it seems. I suggest to remove them. We could possibly provide the callback instance in addition to `net` but at least for the default implementation of the schedule this is not necessary (as it is a method and can access `self.at` anyway). The new signature would then be `schedule(net: NeuralNet) -> callable`.

benjamin-work: Okay, this looks like a good compromise. Make sure to mention in the `schedule` docstring that depending on the input, `schedule` may supersede `at` and `fn`.

ottonemo:
@benjamin-work pls review again

skorch/callbacks/training.py
thomasjpfan: I was thinking something like this for `filter_parameters`:

```python
from collections import product

def filter_parameters(self, patterns, params):
    patten_fns = [
        pattern if callable(pattern) else lambda n: fnmatch(n, pattern)
    ]
    for pattern_fn, (name, param) in product(patten_fns, params):
        if pattern_fn(name):
            yield name, param
```

This would allow `named_parameters` to remain a generator.

skorch/callbacks/training.py
```diff
@@ -273,3 +278,173 @@ def _sink(self, text, verbose):
...
+
+        return self
+
+    def named_parameters(self, net):
+        return list(net.module_.named_parameters())
+
+    def filter_parameters(self, patterns, params):
+        for pattern in patterns:
+            pattern_fn = lambda n: fnmatch(n, pattern)
+            pattern_fn = pattern if callable(pattern) else pattern_fn
```
benjamin-work: Maybe `pattern_fn = partial(fnmatch, pat=pattern) if callable(pattern) else pattern` is more elegant?

skorch/tests/callbacks/test_training.py
```diff
@@ -314,3 +314,232 @@ def test_wrong_threshold_mode(
...
+
+    @pytest.mark.parametrize('mod_init, mod_kwargs', [
+        (False, {}),
+        (True, {}),
+        # Supply a module__ parameter so the model is forced
+        # to re-initialize. Even then parameters should be
+        # frozen correctly.
+        (False, {'module__hidden_units': 5}),
+        (True, {'module__hidden_units': 5}),
+    ])
```
benjamin-work: This `parametrize` could be split into 2 like in the test above, but this is only a minor point.

examples/benchmarks/freezing.py
```diff
@@ -0,0 +1,149 @@
...
+
+    for i in linear_idcs[:-1]:
+        assert not mod.sequential[i].weight.requires_grad
+        assert not net.module_.sequential[i].weight.requires_grad
+
+    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+    return rss, net.history[-1, 'valid_loss'], np.mean(net.history[:, 'dur'])
+
+
+def test_b(verbose=False):
```
benjamin-work: `verbose` unused

skorch/callbacks/training.py
benjamin-work: Also, assigning lambda can be dangerous (though here it should be fine):
```
>>> x = 3
>>> f = lambda y: x + y
>>> g = partial(add, x)
>>> x = 10
>>> f(5)
15
>>> g(5)
8
```

ottonemo:
@benjamin-work addressed your comments

benjamin-work:
Looks good, nice solution overall. Would you open an issue detailing other parts of skorch that need updating after this (docs, notebooks, deprecating old way come to mind).

PR #345, opened 2018-09-14
BUG: Uses rm -f when removing files in CI when caching

thomasjpfan:


thomasjpfan:
In https://travis-ci.org/dnouri/skorch/jobs/428672872, I saw the following:

> rm: remove write-protected regular file ‘/home/travis/conda/pkgs/mkl-2019.0-117/info/LICENSE.txt’

When there is no cache, the files can be removed without error because they are writable. When the files were restored from cache, I suspect that travis makes them not writable. In this case, all `rm` commands would need user confirmation to proceed.. Adding `rm -f` forces the removal without prompting the user.

As a side note, the `find ...` command, removes all directories from the `conda/pkg` directory, this leaves behind all the compressed packages. In the previous version of caching, all the files in every directory were being restored from cache, which took a long time.

benjamin-work:
Thank you.

PR #346, opened 2018-09-18
deprecate filtered optimizer

ottonemo:
Work in progress and open for discussion.

Updates notebooks that freeze parameters and deprecates `filtered_optimizer` et al.

### DIFF
--- a/notebooks/Transfer_Learning.ipynb
+++ b/notebooks/Transfer_Learning.ipynb
@@ -47,7 +47,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


thomasjpfan:
I tested the tutorial notebook locally, and everything works as expected.

thomasjpfan:
@ottonemo 

1. Yes `%matplotlib inline` is missing, it should be included.
2. This is strange, `plot_masks` should work with both numpy arrays and the pillow image, it calls `plt.imshow` on the mask.
3. This could be a bug.

Can you post a github gist of the segmentation notebook you currently working on?

ottonemo:
@thomasjpfan 
I think by carefully setting up the environment I could eliminate most errors but the images still don't look like the images you showed (did the split or the dataset change?). Also the cropping seems wrong and does not look like the cropping in your version. It would be nice of you to check the current state of the notebook: https://gist.github.com/ottonemo/e52610bef80ee23df36007b2184e379d

I just changed the layer freezing and removed the callback names since they are not necessary anymore.

benjamin-work:
> It looks like the `train_test_split` is not deterministic

If that turns out not to be the reason, it could also be possible that the underlying data has changed.

ottonemo:
@benjamin-work with the nuclei notebook merged, please review the changes

PR #347, opened 2018-09-18
MRG: Refactors Paramapper.filter_parameters to use iterator

thomasjpfan:
This is a small refactor to `filter_parameters` to allow `named_parameters` to return an iterator. This reduces the memory usage by not having to allocate memory for a list.

### DIFF
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -4,6 +4,7 @@ import pickle
 import warnings
 from fnmatch import fnmatch
 from functools import partial
+from itertools import product
 
 import numpy as np
 from skorch.callbacks import Callback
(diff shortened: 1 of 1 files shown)


ottonemo:
Does it *really* reduce memory usage? I'm a bit skeptical since, once you evaluate `named_parameters` as a list, there is no generator in the background to generate new results, so everything in [torch.nn.Module._named_parameters](https://github.com/pytorch/pytorch/blob/4e8d9a4a5838e99082c6866032a39efd78d341ac/torch/nn/modules/module.py#L725), for example, can be garbage collected.

benjamin-work:
My impression is that since everything is consumed immediately after it is produced, either approach should not require significantly more or less memory. That said, I like the use of `product` for these kinds of nested loops.

thomasjpfan:
In this case, the memory savings is measurably small. When a API returns an iterator such as `named_parameters`, I prefer leaving it as a iterator, unless there is no other way. In this case, `named_parameters` can be left as an iterator would any compromises.

PR #352, opened 2018-10-04
TST: Less flakey tests

### LINKED ISSUE #350: TestNeuralNet.test_net_learns is flakey
The `TestNeuralNet.test_net_learns` can sometimes fail, because the accuracy_score is sometimes lower than 0.65.

thomasjpfan:
Fixes #350 by using flaky

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,3 +8,4 @@ pytest
 pytest-cov
 sphinx
 sphinx_rtd_theme
+flaky
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -10,6 +10,7 @@ import numpy as np
 import pytest
 import torch
 from torch import nn
+from flaky import flaky
 import torch.nn.functional as F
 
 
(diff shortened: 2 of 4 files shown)


BenjaminBossan:
I'm not sure whether this is the right way. The main reason for that kind of test is to prevent a major bug that prevents training. With these changes, it is too easy that if that bug were to happen, the test would still pass by accident (random fluctuations in score). Therefore, a reasonable margin should be set to make it very unlikely that the test passes accidentally.

Ideally, we would like to have a very high confidence that the model improved but that would probably require many runs/more data/longer training, all of which slow down testing. Using seeds everywhere is not enough because the results would still [depend on the platform](https://pytorch.org/docs/master/notes/randomness.html#reproducibility).

So maybe we should seriously consider libraries like [flaky](https://pypi.org/project/flaky/) that would allow us to only rerun the test if it fails (though that would lower confidence in the results).

benjamin-work:
We are looking at this, will probably merge tomorrow.

benjamin-work:
Thank you.

Issue #354, opened 2018-10-07
NeuralNet.load_params() attempts to load on GPU when no GPU is available


vykhand:
[Code starting at this line](https://github.com/dnouri/skorch/blob/4230d5d5b2290732df4b9a6009dd71ae7b67f9f4/skorch/net.py#L1410) is not working correctly in my opinion. I am trying to debug the restarting from checkpoint on my mac when there is no GPU, and it fails because it attempts to load into GPU.

Here is the code:

```
        use_cuda = self.device.startswith('cuda')
        cuda_req_not_met = (use_cuda and not torch.cuda.is_available())
        if use_cuda or cuda_req_not_met:
            # Eiher we want to load the model to the CPU in which case
            # we are loading in a way where it doesn't matter if the data
            # was on the GPU or not or the model was on the GPU but there
            # is no CUDA device available.
            if cuda_req_not_met:
                warnings.warn(
                    "Model configured to use CUDA but no CUDA devices "
                    "available. Loading on CPU instead.",
                    ResourceWarning)
                self.device = 'cpu'
            model = torch.load(f, lambda storage, loc: storage)
        else:
            model = torch.load(f)
```

I ran this in the debugger, and in my case both use_cuda and cuda_req_not_met are false, leading us to the cuda version of torch.load and failing.

PR #356, opened 2018-10-07
Fixes problem attempt to load on GPU when no GPU is available (#354)

vykhand:


benjamin-work:
This is probably in reference to #354. I believe this solution could create another problem, namely when a user indicates they want to use cuda but cuda is not available.

Do I understand correctly that your case is that you train with a GPU and later want to load on a machine without a GPU? Just for a test, what happens when you set `device='cuda'` on the net?

benjamin-work:
> I believe that statement cuda_req_not_met = (use_cuda and not torch.cuda.is_available()) is very confusing.

I think this is fine, it literally means "cuda requirements are not met when we want to use cuda and cuda is not available".

The bug in the implementation seems to be that if we run into the `else` clause, loading fails. I could replicate this. Apparently, we need the `map_location` argument, so this could be a more correct fix:

```
        ...
        if cuda_req_not_met:
            # Eiher we want to load the model to the CPU in which case
            # we are loading in a way where it doesn't matter if the data
            # was on the GPU or not or the model was on the GPU but there
            # is no CUDA device available.
            warnings.warn(
                "Model configured to use CUDA but no CUDA devices "
                "available. Loading on CPU instead.",
                ResourceWarning)
            self.device = 'cpu'

        model = torch.load(f, map_location=lambda storage, loc: storage)
```

A more flexible solution could be that we need to allow users to pass the `map_location` argument to `load_params` so that they can control it themselves.

benjamin-work:
FYI we have a plan to fix this problem and to also improve test coverage of this whole topic. PR should be coming soon.

vykhand:
>         if cuda_req_not_met:
>             # Eiher we want to load the model to the CPU in which case
>             # we are loading in a way where it doesn't matter if the data
>             # was on the GPU or not or the model was on the GPU but there
>             # is no CUDA device available.
>             warnings.warn(
>                 "Model configured to use CUDA but no CUDA devices "
>                 "available. Loading on CPU instead.",
>                 ResourceWarning)
>             self.device = 'cpu'
> 
>         model = torch.load(f, map_location=lambda storage, loc: storage)
> ```
That would still prevent the model from loading on my Mac because cuda_requ_not_met is false. Even if I set self.device = "cuda". So I'll have to stick to my local fork in such cases :(

benjamin-work:
Did you look at my PR #358? Ideally, you could test it to before merging. This is the kind of problem that is hard to test except in practice.

benjamin-work:
Since #358 is merged, I will close this. @vykhand please re-open in case #358 did not solve the original issue.

PR #359, opened 2018-10-09
DOC: Updates seg tutorial to use freezer

thomasjpfan:
Continues the fix in #346

This PR saves the training and validation ids into a json file to allow the notebook to be reproducible.


benjamin-work:
You uploaded a file ` Nuclei_Image_Segmentation.ipynb.ipynb`. Is this intended?

PR #360, opened 2018-10-11
[MRG] Adds optimizer and history to save/load_params

### LINKED ISSUE #357: Add an easy way to save and load the optimizer
Some optimizers are stateful. When just using `save_params`, this state is lost. We should either add a possibility to save the parameters of the optimizer in a specific method, or adjust `save_params` so that optimizer state is included.

Note that this will also affect `Checkpoint`.

thomasjpfan:
Fixes #357

1. Adds optimizer and history to `save_params` and `load_params`. Deprecates using `f`.
2. Adds the optimizer option to `Checkpoint`.
3. Deprecates `save_history` and `load_history` in favor of `save_params` and `load_params` to save history. This simplifies the skorch api to only needing `save_params` and `load_params` to save/load state.

This PR can merge easily with #358 by replacing `_get_state_dict` with the implementation in #358 and returning the state dict.

### DIFF
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -34,6 +34,7 @@ class Checkpoint(Callback):
     Some or all of the following can be saved:
 
       - model parameters (see ``f_params`` parameter);
+      - optimizer state (see ``f_optimizer`` parameter);
       - training history (see ``f_history`` parameter);
       - entire model object (see ``f_pickle`` parameter).
 
@@ -84,6 +85,18 @@ class Checkpoint(Callback):
 
       >>> cb = Checkpoint(f_params="params_{last_epoch[epoch]}.pt")
 
+    f_optimizer : file-like object, str, None (default=None)
+      File path to the file or file-like object where the optimizer
+      state should be saved. Pass ``None`` to disable saving
+      model parameters.
+
+      If the value is a string you can also use format specifiers
+      to, for example, indicate the current epoch. Accessible format
+      values are ``net``, ``last_epoch`` and ``last_batch``.
+      Example to include last epoch number in file name:
+
+      >>> cb = Checkpoint(f_optimizer="optimizer_{last_epoch[epoch]}.pt")
+
     f_history : file-like object, str, None (default=None)
       File path to the file or file-like object where the model
       training history should be saved. Pass ``None`` to disable
@@ -108,6 +121,7 @@ class Checkpoint(Callback):
             target=None,
             monitor='valid_loss_best',
             f_params='params.pt',
+            f_optimizer=None,
             f_history=None,
             f_pickle=None,
             sink=noop,
@@ -122,6 +136,7 @@ class Checkpoint(Callback):
             f_params = target
         self.monitor = monitor
         self.f_params = f_params
+        self.f_optimizer = f_optimizer
         self.f_history = f_history
         self.f_pickle = f_pickle
         self.sink = sink
@@ -157,10 +172,11 @@ class Checkpoint(Callback):
           - training history;
           - entire model object.
         """
-        if self.f_params:
-            net.save_params(self._format_target(net, self.f_params))
-        if self.f_history:
-            net.save_history(self._format_target(net, self.f_history))
+        if self.f_params or self.f_history:
+            net.save_params(
+                self._format_target(net, self.f_params),
+                f_optimizer=self._format_target(net, self.f_optimizer),
+                f_history=self.f_history)
         if self.f_pickle:
             f_pickle = self._format_target(net, self.f_pickle)
             with open_file_like(f_pickle, 'wb') as f:
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -12,6 +12,7 @@ import pickle
 from unittest.mock import Mock
 from unittest.mock import patch
 import sys
+from contextlib import ExitStack
 
 import numpy as np
 import pytest
@@ -327,6 +328,126 @@ class TestNeuralNet:
         score_after = accuracy_score(y, net.predict(X))
         assert np.isclose(score_after, score_before)
 
+    def test_save_load_state_dict_file_with_history_optimizer(
+            self, net_cls, module_cls, data, tmpdir):
+        net = net_cls(
+            module_cls, max_epochs=10, lr=0.1,
+            optimizer=torch.optim.Adam)
+        X, y = data
+        net.fit(X, y)
+
+        skorch_tmpdir = tmpdir.mkdir('skorch')
+        p = skorch_tmpdir.join('testmodel.pkl')
+        o = skorch_tmpdir.join('optimizer.pkl')
+        h = skorch_tmpdir.join('history.json')
+
+        with ExitStack() as stack:
+            p_fp = stack.enter_context(open(str(p), 'wb'))
+            o_fp = stack.enter_context(open(str(o), 'wb'))
+            h_fp = stack.enter_context(open(str(h), 'w'))
+            net.save_params(p_fp, f_optimizer=o_fp, f_history=h_fp)
+
+            # 'step' is state from the Adam optimizer
+            orig_steps = [v['step'] for v in
+                          net.optimizer_.state_dict()['state'].values()]
+            orig_loss = np.array(net.history[:, 'train_loss'])
+            del net
+
+        with ExitStack() as stack:
+            p_fp = stack.enter_context(open(str(p), 'rb'))
+            o_fp = stack.enter_context(open(str(o), 'rb'))
+            h_fp = stack.enter_context(open(str(h), 'r'))
+            new_net = net_cls(
+                module_cls, max_epochs=10, lr=0.1,
+                optimizer=torch.optim.Adam).initialize()
+            new_net.load_params(p_fp, f_optimizer=o_fp, f_history=h_fp)
+
+            new_steps = [v['step'] for v in
+                         new_net.optimizer_.state_dict()['state'].values()]
+            new_loss = np.array(new_net.history[:, 'train_loss'])
+
+            assert np.allclose(orig_loss, new_loss)
+            assert orig_steps == new_steps
+
+    def test_save_load_state_dict_str_with_history_optimizer(
+            self, net_cls, module_cls, data, tmpdir):
+        net = net_cls(
+            module_cls, max_epochs=10, lr=0.1,
+            optimizer=torch.optim.Adam)
+        X, y = data
+        net.fit(X, y)
+
+        skorch_tmpdir = tmpdir.mkdir('skorch')
+        p = str(skorch_tmpdir.join('testmodel.pkl'))
+        o = str(skorch_tmpdir.join('optimizer.pkl'))
+        h = str(skorch_tmpdir.join('history.json'))
+
+        net.save_params(p, f_optimizer=o, f_history=h)
+
+        # 'step' is state from the Adam optimizer
+        orig_steps = [v['step'] for v in
+                      net.optimizer_.state_dict()['state'].values()]
+        orig_loss = np.array(net.history[:, 'train_loss'])
+        del net
+
+        new_net = net_cls(
+            module_cls, max_epochs=10, lr=0.1,
+            optimizer=torch.optim.Adam).initialize()
+        new_net.load_params(p, f_optimizer=o, f_history=h)
+
+        new_steps = [v['step'] for v in
+                     new_net.optimizer_.state_dict()['state'].values()]
+        new_loss = np.array(new_net.history[:, 'train_loss'])
+
+        assert np.allclose(orig_loss, new_loss)
+        assert orig_steps == new_steps
+
+    def test_save_params_f_keyword_deprecation(
+            self, net_cls, module_cls, tmpdir):
+        p = tmpdir.mkdir('skorch').join('testmodel.pkl')
+        net = net_cls(module_cls).initialize()
+
+        # TODO: remove this test when the target argument is removed
+        # after its deprecation grace period is over.
+        with pytest.warns(DeprecationWarning):
+            net.save_params(f=str(p))
+
+        assert p.exists()
+
+    def test_save_params_not_init_optimizer(
+            self, net_cls, module_cls, tmpdir):
+        from skorch.exceptions import NotInitializedError
+
+        net = net_cls(module_cls).initialize_module()
+        skorch_tmpdir = tmpdir.mkdir('skorch')
+        p = skorch_tmpdir.join('testmodel.pkl')
+        o = skorch_tmpdir.join('optimizer.pkl')
+
+        with pytest.raises(NotInitializedError) as exc:
+            net.save_params(str(p), f_optimizer=o)
+        expected = ("Cannot save state of an un-initialized optimizer. "
+                    "Please initialize first by calling .initialize() "
+                    "or by fitting the model with .fit(...).")
+        assert exc.value.args[0] == expected
+
+    def test_load_params_not_init_optimizer(
+            self, net_cls, module_cls, tmpdir):
+        from skorch.exceptions import NotInitializedError
+
+        net = net_cls(module_cls).initialize_module()
+        skorch_tmpdir = tmpdir.mkdir('skorch')
+        p = skorch_tmpdir.join('testmodel.pkl')
+        o = skorch_tmpdir.join('optimizer.pkl')
+
+        net.save_params(str(p))
+
+        with pytest.raises(NotInitializedError) as exc:
+            net.load_params(str(p), f_optimizer=o)
+        expected = ("Cannot load state of an un-initialized optimizer. "
+                    "Please initialize first by calling .initialize() "
+                    "or by fitting the model with .fit(...).")
+        assert exc.value.args[0] == expected
+
     def test_save_state_dict_not_init(
             self, net_cls, module_cls, tmpdir):
         from skorch.exceptions import NotInitializedError
@@ -371,7 +492,7 @@ class TestNeuralNet:
             'Model configured to use CUDA but no CUDA '
             'devices available. Loading on CPU instead.')
 
-    def test_save_load_history_file_obj(
+    def test_save_params_with_history_file_obj(
             self, net_cls, module_cls, net_fit, tmpdir):
         net = net_cls(module_cls).initialize()
 
@@ -379,15 +500,17 @@ class TestNeuralNet:
 
         p = tmpdir.mkdir('skorch').join('history.json')
         with open(str(p), 'w') as f:
-            net_fit.save_history(f)
+            net_fit.save_params(f_history=f)
         del net_fit
         with open(str(p), 'r') as f:
-            net.load_history(f)
+            net.load_params(f_history=f)
 
         assert net.history == history_before
 
+    # TODO: remove this test when the target argument is removed
+    # after its deprecation grace period is over.
     @pytest.mark.parametrize('converter', [str, Path])
-    def test_save_load_history_file_path(
+    def test_save_history_file_path(
             self, net_cls, module_cls, net_fit, tmpdir, converter):
         # Test loading/saving with different kinds of path representations.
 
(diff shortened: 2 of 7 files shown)


benjamin-work:
Just FYI, I'm looking at this but am somewhat busy.

> What if we provide an (optional) common path prefix option in `Checkpoint` constructor?

I believe if we did this, it should be consistent across all methods (i.e. also `load/save_params`). Just to get a perspective on this, is this a common pattern, can you name other libs that do something similar?

thomasjpfan:
`ignite` Checkpoint object uses a common prefix with a dirname in their parameters: https://github.com/pytorch/ignite/blob/83fe17dd513a4f0ed7fe1cfea6ea00541557465b/ignite/handlers/checkpoint.py#L7

The common path prefix option would add some complexity for better ergonomics:

```python
## Option 1: no common prefix
net.save_params(f_params="run1_model.pkl", 
                f_optimizer="run1_opt.pkl", f_training="run1_training.json")

## Option 2: common prefix
net.save_params(f_params="model.pkl", f_optimizer="opt.pkl", 
                f_training="training.json", common_prefix="run1")
```

I tend to like **option 1: no common prefix**, since there is less "magic".

benjamin-work:
The thing with "magic" is that I'm okay with it as long as it's intuitive and straightforward. That can often be derived be asking whether that pattern is common or not, which is why I asked for other libraries. ignite is relevant because it covers similar ground as skorch, but it is a rather niche library (as is skorch). If this pattern was used in a more mainstream library (say pytorch, sklearn, pandas), that would be a strong argument IMO.

thomasjpfan:
Keras does not run into this issue, since it only saves the model parameters. sklearn uses pickle to save the the whole model. Tensorboard saves many files into the folder. Tensorflow uses a `SavedModel` that saves many files into a directory. (Tensorflow has many appraoches to saving models.)

If we are okay with `common_prefix`, we can provide a pretty good api for starting and stopping a training session:

```python
f_names = {"f_params": "model.pkl", "f_optimizer": "opt.pkl", "f_history": "history.json"}

best_checkpoint = Checkpoint(common_prefix="exp1", **f_names)

latest_checkpoint = Checkpoint(common_prefix="exp1_latest", 
                               monitor=None, load_init_state=True, **f_names)

net = NeuralNet(..., callbacks = [best_checkpoint, latest_checkpoint])
```

This adds `load_init_state` to Checkpoint which will load the parameters optimizer and history in `initialize`. This api could be even simpler by having default values for `f_optimizer` and `f_history` so they are checkpointed by default. This means the user would only need to supply `common_prefix` to run a new experiment:

```python
best_checkpoint = Checkpoint(common_prefix="exp2")

latest_checkpoint = Checkpoint(common_prefix="exp2_latest", 
                               monitor=None, load_init_state=True)

net = NeuralNet(..., callbacks = [best_checkpoint, latest_checkpoint])
```

thomasjpfan:
I am unsure if `NeuralNet.load_params` and `NeuralNet.save_params` should have the concept of `filename_prefix`. Currently, in this PR, the API is pretty simple:

```python
def load_params(
    self, f_params=None, f_optimizer=None, f_history=None, f=None):
    ....

def save_params(
    self, f_params=None, f_optimizer=None, f_history=None, f=None):
    ....
```
Adding `filename_prefix` would complicate this API. I think only `Checkpoint` should have a `filename_prefix` parameter. Additionally, if we have default values for `f_optimizer` and `f_history`, and add `load_saved_state` to `Checkpoint`, a user would not need to call `load_params` or `save_params` anymore:

```python
cp = Checkpoint(filename_prefix="experiment1", load_saved_state=True, monitor=None)
net = NeuralNet(..., callback=[cp])
net.fit(X, y)
```

`Checkpoint` will load the saved state if it exists. After training, the user can run:

```python
cp = Checkpoint(filename_prefix="experiment1", load_saved_state=True, monitor=None)
net = NeuralNet(..., callback=[cp]).initialize()
net.predict(X)
```

The call to `initialize` will trigger `load_saved_state` with the default values of `f_*` prefixed with `filename_prefix`. This `Checkpoint` feature can be done in another PR.

If we agree on the API of `NeuralNet.load_params` and `NeuralNet.save_params`, I think this PR is ready for review/merge. I would prefer not to add more to this PR making it harder to review.

benjamin-work:
> If we are okay with `filename_prefix`, we can provide a pretty good api for starting and stopping a training session ... This api could be even simpler by having default values for f_optimizer and f_history

That is a pretty good argument.

> I am unsure if `NeuralNet.load_params` and `NeuralNet.save_params` should have the concept of `filename_prefix` ... I think only Checkpoint should have a filename_prefix parameter.

Could you explain why the same API would not work as well for `load/save_params`? I would really prefer to have the same API for this and `Checkpoint`, with as little code duplication as possible.

> `Checkpoint` will load the saved state if it exists.

I'm not convinced that we should have `Checkpoint` do the saving and loading of models, in my mind this is going at it the wrong way.

benjamin-work:
Something else that came to my mind: Should we gracefully intercept possible JSON encoding errors when saving history? Say I have a very long running model and start if overnight with a `Checkpoint`. When history is saved by default but I accidentally put a numpy structure into history, I come back the next day only to find that model training has stopped. Of course it would be "my fault" but at the same time very inconvenient (and as of now the error would not occur since history is not saved by default).

thomasjpfan:
> Could you explain why the same API would not work as well for load/save_params? I would really prefer to have the same API for this and Checkpoint, with as little code duplication as possible.

My major concern has to do with the default values and how `Checkpoint` has its own custom filename formatting. I think the default values for the `f_*` arguments in `load/save_params` should be `None`. This way the user could save and load specific parts of the state. Additionally, I think `load/save_params` should not handle any custom formatting (including prefixes). I was imaging the API to `Checkpoint` and `load_params` to look like:

```python
cb = Checkpoint(fn_prefix="exp2", monitor=None)
net = NeuralNet(..., callbacks=[cb])
net.fit(X, y)

# load and continue training
net = NeuralNet(..., callbacks=[cb]).initialize()
net.load_params(f_params=cb.f_params_, f_optimizer=cb.f_optimizer_, f_history=cb.f_history_)
net.fit(X, y)
```

where `load_params` uses the formatted values of `f_*` to load the parameters. The formatted values would include the `fn_prefix` and other custom formatting. This way `Checkpoint` manages all the filenames.

> I'm not convinced that we should have Checkpoint do the saving and loading of models, in my mind this is going at it the wrong way.

I agree. Maybe a `LoadInitState` callback would be better to load models. This way the api would look like:

```python
cb = Checkpoint(fn_prefix="exp2", monitor=None)
load_init_state = LoadInitState(checkpoint=cb)
net = NeuralNet(..., callbacks=[cb, load_init_state])
net.fit(X, y)
```

A user can run this over and over again, and training will continue where it left off. Like my previous example, `Checkpoint` handles all filenames, internally `LoadInitState` will use `cb.f_*_`, to get the filenames to load during `initialize`.

> Should we gracefully intercept possible JSON encoding errors when saving history?

I agree. `load/save_params` should still raise the exception and `Checkpoint` will have to catch it. Then `Checkpoint` can print the error to sink, and save everything else that can be saved. This would lead to a saved model and optimizer without the correct history, but it is better than nothing.

BenjaminBossan:
Okay, I think I largely agree with the above. The only thing that I would like to discuss is whether we need  this:

`net.load_params(f_params=cb.f_params_, f_optimizer=cb.f_optimizer_, f_history=cb.f_history_)`

Instead it would be nice if we could just have:

`net.load_params(cb)`

Maybe a little bit surprising but certainly a time saver. `LoadInitState` would then just call this code.

thomasjpfan:
I agree that `net.load_params` with all those keywords does not look nice. Another way to make it nicer is:

```python
net.load_params(**cb.formatted_files)
```

where `formatted_files` is a dictionary with the formatted file names.

thomasjpfan:
With the addition of `LoadInitState` the api looks like this:

```python
cp = Checkpoint(monitor='valid_loss_best')
load_state = LoadInitState(cp)
net = NeuralNet(..., callbacks=[cp, load_state])
net.fit(X, y)
```

The above example can run multiple times, and will always start off from where the last checkpoint. In this case, the checkpoint happens when `valid_loss_best` is true. This means if `max_epoch=10` and the `valid_loss_best` happens at epoch 5, the second run will start at epoch 5 and continue for another 10 epochs.

If a user wants to save at the end of every checkpoint:

```python
lastest_cp = Checkpoint(monitor=None)
load_state = LoadInitState(cp)
net = NeuralNet(..., callbacks=[latest_cp, load_state])
net.fit(X, y)
```

Repeated runs of this will continue training from where the previous run left off.

If a user wants to use their checkpoint to make a prediction:

```python
lastest_cp = Checkpoint(monitor=None)
net = NeuralNet(..., callbacks=[latest_cp, load_state]).initialize()
net.load_params(checkpoint=lastest_cp)
net.predict(X)
```

All of the above works for filenames with that uses `net`, `last_epoch` and `last_batch`. This was done by having `net.load_params` load the history first before consulting checkpoint for the filenames.

thomasjpfan:
With the addition of `FinalCheckpoint`, it is possible to do this:

```python
dirname = 'exp1'
cp_monitor = Checkpoint(dirname=dirname)
cp_final = FinalCheckpoint(dirname=dirname)
load_state = LoadInitState(cp_final)

net = NeuralNet(..., callbacks=[cp_monitor, cp_final, load_state])
net.fit(X, y)
```

The `FinalCheckpoint` saves the model at the end of training. If one wants to continue training from the the final model, the final checkpoint is passed to `LoadInitState` and is loaded on the second run. The default prefix for `FinalCheckpoint` is `final_` so it does not conflict with `cp_monitor` saving to the same directory.

This means the user does not need to run `net.load_params` or `net.save_params. Everything is handled by the checkpoints.

BenjaminBossan:
Thank you for this fantastic addition. I will have a deep look at this tomorrow.

The explanations you gave in your previous posts seem to be too useful to bury them in a closed PR, especially since this topic keeps coming up repeatedly. Would you mind adding those to the docs? I could see a new section about saving and loading, where you could put the above almost verbatim.

thomasjpfan:
I will be happy to add a "Saving and Loading" section into the docs.

skorch/callbacks/training.py
```diff
@@ -140,41 +172,91 @@ def on_epoch_end(self, net, **kwargs):
...
+
         if self.f_pickle:
-            f_pickle = self._format_target(net, self.f_pickle)
+            f_pickle = self._format_target(net, self.f_pickle, -1)
             with open_file_like(f_pickle, 'wb') as f:
                 pickle.dump(net, f)
 
-    def _format_target(self, net, f):
+    @property
+    def f_history_(self):
```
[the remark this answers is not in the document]
thomasjpfan: I was considering the following edge case:

```python
cp = Checkpoint(dirname='exp1')

net = NeuralNet(..., callbacks=[]) # does not include cp
net.initialize()
net.load_params(checkpoint=cp)
net.predict(X)
```

The checkpoint is not initialized, but can still be loaded.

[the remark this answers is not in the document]
benjamin-work: Okay, makes sense, could you add this as a comment?

thomasjpfan:
To deal with the dirname and `f_*` being file-like objects issue, a `_valid_filenames` function was placed in the `__init__` and the `initialize`. The check is in  `__init__`, to catch the error early. It is in `initialize` for users that used `net.set_params` to set a `f_*`  to a file-like object.

thomasjpfan:
The justification for `LoadInitState` is as follows:

#### Case 1 - No `LoadInitState`

```python
cp = Checkpoint(dirname='exp1')
net = NeuralNet(..., callbacks=[cp]).initialize()
net.load_params(checkpoint=cp) # update to gracefully handle the first run
net.partial_fit(X, y)
```

#### Case 2 - With `LoadInitState`

```python
cp = Checkpoint(dirname='exp1')
load_state = LoadInitState(cp)
net = NeuralNet(..., callbacks=[cp, load_state])
net.fit(X, y)
```

It makes the API slightly better for new users.

skorch/callbacks/training.py
```diff
@@ -450,3 +532,135 @@ class Initializer(ParamMapper):
     def __init__(self, *args, **kwargs):
         kwargs['at'] = kwargs.get('at', 1)
         super().__init__(*args, **kwargs)
+
+
+class LoadInitState(Callback):
```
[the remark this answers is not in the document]
thomasjpfan: I tried to keep `LoadInitState` interface fairly small by restricting it to just checkpoints. This goes along with the idea that `Checkpoint` is the only object handling filenames. `LoadInitState` tries to provide a nicer interface for loading a Checkpoint into a `NeutralNet`.

The motivation for adding filenames to `LoadInitState` is for those not using `Checkpoint` and are using `NeuralNet.save_params` to save their model. In this case, I would think that they would use `NeuralNet.load_params` to load their models as well.

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+         10        0.1992       0.9453        0.2101        0.0182
+         11        0.2033       0.9453        0.1947     +  0.0211
+         12        0.1825       0.9104        0.2515        0.0185
+         13        0.2010       0.9453        0.1927     +  0.0187
+         14        0.1508       0.9453        0.1952        0.0198
+         15        0.1679       0.9502        0.1905     +  0.0181
+         16        0.1516       0.9453        0.1864     +  0.0192
+         17        0.1576       0.9453        0.1804     +  0.0184
+
+Since we started from the previous checkpoint ended at epoch 7,
```
ottonemo: > [...] checkpoint **which** ended at [...]

skorch/tests/test_net.py
```diff
@@ -320,13 +326,237 @@ def test_save_load_state_dict_str(
...
+    @pytest.fixture(scope='module')
+    def net_fit_adam(self, net_cls, module_cls, data):
+        net = net_cls(
+            module_cls, max_epochs=2, lr=0.1,
+            optimizer=torch.optim.Adam)
+        net.fit(*data)
+        return net
+
+    def test_save_load_state_dict_file_with_history_optimizer(
+            self, net_cls, module_cls, net_fit_adam, data, tmpdir):
```
benjamin-work: `data` is no longer needed

skorch/tests/test_net.py
```diff
@@ -320,13 +326,237 @@ def test_save_load_state_dict_str(
...
+
+            new_steps = [v['step'] for v in
+                         new_net.optimizer_.state_dict()['state'].values()]
+            new_loss = np.array(new_net.history[:, 'train_loss'])
+
+            assert np.allclose(orig_loss, new_loss)
+            assert orig_steps == new_steps
+
+    def test_save_load_state_dict_str_with_history_optimizer(
+            self, net_cls, module_cls, data, tmpdir, net_fit_adam):
```
benjamin-work: `data` is no longer needed

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+          4        0.2514       0.8905        0.3080     +  0.0237
+          5        0.2333       0.9154        0.2844     +  0.0203
+          6        0.2177       0.9403        0.2164     +  0.0215
+          7        0.2194       0.9403        0.2159     +  0.0220
+          8        0.2027       0.9403        0.2299        0.0202
+          9        0.1864       0.9254        0.2313        0.0196
+         10        0.2024       0.9353        0.2333        0.0221
+
+By default, the checkpoint observes ``valid_loss`` and will save
+the model when the `valid_loss` is lowest. This can be seen by
```
taketwo:
```suggestion
the model when the ``valid_loss`` is lowest. This can be seen by
```

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+the ``+`` mark in the ``cp`` column of the logs.
+
+On our first run, the valiation loss does not improve after the 7th
```
taketwo:
```suggestion
On our first run, the validation loss did not improve after the 7th
```

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+         16        0.1516       0.9453        0.1864     +  0.0192
+         17        0.1576       0.9453        0.1804     +  0.0184
+
+Since we started from the previous checkpoint which ended at epoch 7,
+the second run starts at epoch 8, continuing from the first checkpoint.
+With a lower learning rate, the validation loss was able to improve!
+
+Notice in the first run, we included a :class:`.FinalCheckpoint` in the
+callbacks. This checkpoint saves the model at the end of training. This
+checkpoint can be passed to :class:`.LoadInitState` continue training:
```
taketwo:
```suggestion
checkpoint can be passed to :class:`.LoadInitState` to continue training:
```

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+          8        0.1319       0.9303        0.1910        0.0220
+          9        0.1558       0.9254        0.1923        0.0189
+         10        0.1432       0.9303        0.2219        0.0192
+
+This stores the model into the ``'exp2'`` directory. Since this is the first run,
+the :class:`.LoadInitState` callback does not do anything. If we were to run the above
+script again, the :class:`.LoadInitState` callback will load the model from the
+checkpoint.
+
+In the run above, the best checkpoint was at epoch 6, we can load this checkpoint
```
taketwo:
```suggestion
In the run above, the last checkpoint was created at epoch 6, we can load this checkpoint
```

docs/user/save_load.rst
```diff
@@ -0,0 +1,171 @@
...
+    net = NeuralNetClassifier(
+        MyModule, lr=0.5, module__num_units=20,
+    )
+    net.initialize()
+    net.load_params(checkpoint=cp)
+
+    y_pred = net.predict(X)
+
+In this case, it is important to initialize the neutral net before running
+:meth:`.NeutralNet.load_params`
```
taketwo:
```suggestion
:meth:`.NeutralNet.load_params`.
```

skorch/callbacks/training.py
```diff
@@ -84,20 +87,36 @@ class Checkpoint(Callback):
...
-      Supports the same format specifiers as ``f_params``.
-
     f_pickle : file-like object, str, None (default=None)
       File path to the file or file-like object where the entire
       model object should be pickled. Pass ``None`` to disable
       pickling.
 
       Supports the same format specifiers as ``f_params``.
 
+    fn_prefix: str (default='')
```
taketwo: Shouldn't this be called `f_prefix` for consistency? Otherwise sometimes we "abbreviate" filename to "f" and sometimes to "fn".

skorch/callbacks/training.py
```diff
@@ -84,20 +87,36 @@ class Checkpoint(Callback):
...
+      Prefix for filenames. If ``f_params``, ``f_optimizer``, ``f_history``,
+      or ``f_pickle`` are strings, they will be prefixed by ``fn_prefix``.
+
+    dirname: str (default='')
```
taketwo: Can we also support `pathlib.Path` here?

skorch/callbacks/training.py
```diff
@@ -450,3 +541,139 @@ class Initializer(ParamMapper):
...
+    history when the validation loss is minimized. During the first run,
+    there are no files on disk, thus :class:`.LoadInitState` will
+    not load anything. When running the example a second time,
+    :class:`LoadInitState` will load the best model from the first run and
+    continue training from there.
+
+    Parameters
+    ----------
+    checkpoint: :class:`.Checkpoint`
+      Checkpoint to get filenames from
```
taketwo:
```suggestion
      Checkpoint to get filenames from.
```

skorch/callbacks/training.py
```diff
@@ -450,3 +541,139 @@ class Initializer(ParamMapper):
...
+      model object should be pickled. Pass ``None`` to disable
+      pickling.
+
+      Supports the same format specifiers as ``f_params``.
+
+    fn_prefix: str (default='final_')
+      Prefix for filenames. If ``f_params``, ``f_optimizer``, ``f_history``,
+      or ``f_pickle`` are strings, they will be prefixed by ``fn_prefix``.
+
+    dirname: file-like object, str (default='')
```
taketwo: `file-like` is not supported AFAIU.

benjamin-work:
>     1. I noticed a lot of cases in test when a `pathlib.Path` object is cast into `str` to be passed to our `f_*` parameters. Should we perhaps natively support `Path`s to reduce friction?

The reason for that was that Python 3.5 does not support passing `Path`s as filename identifiers.

Regarding the name `FinalCheckpoint`: I did not have the association with that implying the best model checkpoint (but am also not a native speaker). Would the best model checkpoint not simply be what `Checkpoint` does by default? I have to say I like the name, what do others think?

taketwo:
> The reason for that was that Python 3.5 does not support passing Paths as filename identifiers.

Right. So what I was suggesting is to consider moving this cast inside `skorch` functions such that the users don't need to think about which Python version they are using and just pass `Path` objects.

> Would the best model checkpoint not simply be what Checkpoint does by default?

Yes it is, and I feel this contributes to my confusion. Line of thought: `Checkpoint` saves best performing models. What does `FinalCheckpoint` do then? Saves the "final" best performing model. But what does "final" mean...? Hm....

skorch/callbacks/training.py
ottonemo: I think `f` is a file-like object while `fn` is the name specifically. So it makes sense to have `f_*` for destinations/paths/file-likes and `fn_prefix` for a common prefix *if* `f_*` params are actually names.

ottonemo:
I think final is fine since it is the last checkpoint that gets written no matter what the scores are (or ^C is hit). When considering alternatives such as `LatestCheckpoint` there is also a different notion where the user might be inclined to believe the checkpoint is automatically the latest best checkpoint. So I'm fine with `FinalCheckpoint`.

thomasjpfan:
> Side note: does FinalCheckpoint support ctrl-c exit from training?

This is a good idea. There are some details that would need to be ironed out, since stopping training in the middle of a epoch, leaves the history in a interesting state. Given this PR is already +1k lines, I can work on this in another PR.

>  Saves the "final" best performing model. But what does "final" mean...? Hm....

The `FinalCheckpoint` considers the final epoch the "best". Given that, maybe `FinalEpochCheckpoint` or `LastEpochCheckpoint`, would be better?

This is why there are marketing teams for naming products. 😅

benjamin-work:
> Given this PR is already +1k lines, I can work on this in another PR.

yeah

> maybe `FinalEpochCheckpoint` or `LastEpochCheckpoint`, would be better?

Maybe `LatestCheckpoint`?

thomasjpfan:
`FinalCheckpoint` and `LatestCheckpoint` are almost the same, since final and latest are synonyms: https://www.thesaurus.com/browse/final.

There are many words for `final`, we have considered `end`, `final`, `last`, and `latest`. It seems the confusion comes from what this word is describing. To avoid confusion, we can add what the `final` word is describing. `TrainEndCheckpoint` may be the only name that has no confusion. It is uglier, but if it makes the new user experience better, I would make this compromise.

benjamin-work:
From my point of view, this is good to go.

benjamin-work:
Great work everyone, this is the PR with the most comments so far :)

I merged this but this PR appears to be open probably due to this github issue: https://blog.github.com/2018-10-21-october21-incident-report/

We hope that this will fix itself when they restore order.

PR #366, opened 2018-10-19
[MRG]: Updates transfer learning notebook in Colab

thomasjpfan:
It seems like `DATALAB_ENV` is not defined in Colab anymore. This PR tests if `COLAB_GPU` is defined to check if the notebook is running on colab. 

Since the Freezer callback is in master, this PR installs skorch from master.

This PR can be tested out at https://colab.research.google.com/github/thomasjpfan/skorch/blob/transfer_learning_colab-fix/notebooks/Transfer_Learning.ipynb


benjamin-work:
@thomasjpfan There seems to be a merge conflict, would you resolve this?

benjamin-work:
It might also be worth taking a look at this for future updates in notebooks: https://www.reviewnb.com/

PR #367, opened 2018-10-23
Unify device handling in __setstate__ and load_params

ottonemo:
Before each of these functions had their own logic of
determining a fallback device if special devices such
as CUDA are not available. Now there is one common
way of handling this case.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -24,6 +24,7 @@ from skorch.exceptions import NotInitializedError
 from skorch.history import History
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import duplicate_items
+from skorch.utils import get_map_location
 from skorch.utils import is_dataset
 from skorch.utils import noop
 from skorch.utils import params_for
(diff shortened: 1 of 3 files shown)


skorch/net.py
```diff
@@ -1421,21 +1409,33 @@ def save_params(
...
             map_location = torch.device('cpu')
+        return map_location
 
-        return torch.load(f, map_location=map_location)
+    def _handle_mismatched_devices(self, net_state, map_device):
+        """Warn the user and update the network state to use a non-cuda device
+        in case the network state requests e.g. CUDA but CUDA is not available.
+        """
+        requested_device = net_state['device']
+        if torch.device(requested_device) != torch.device(map_device):
```
benjamin-work: Is it necessary to use `torch.device`?

```suggestion
        if requested_device != map_device:
```

Maybe `torch.device` covers more but I wouldn't know what (e.g. even with 1 GPU, `torch.device('cuda:0') != torch.device('cuda')`)

benjamin-work:
Currently, two warnings are shown when trying to load to CUDA while CUDA is unavailable. Is there a scenario that only triggers one of them. If so, do we need a test for this? If not, do we really need two warnings?

skorch/net.py
ottonemo: It handles the case where one of the values is a string instead of a device:

```
device(device('cuda')) == device('cuda')
```

The case of explicitly enumerated devices could be tricky, e.g. when switching from a single gpu machine to a multi gpu machine where it is now necessary to specify a device qualifier. We should probably resort to device type prefix matching to avoid this issue (instead of a equality test).

ottonemo:
@benjamin-work please review again

PR #368, opened 2018-10-24
DOC: Basic Usage - Adds support for colab 

thomasjpfan:
You can test this out here:

https://colab.research.google.com/github/thomasjpfan/skorch/blob/colab-basic-usage/notebooks/Basic_Usage.ipynb


benjamin-work:
LGTM

PR #369, opened 2018-10-25
RFC: runtime parameter setting (e.g. LR)

### LINKED ISSUE #349: More verbosity when re-initializing components
I often continue training my models using `partial_fit` preceded by a `set_params()` where I increase/decrease parameters such as the learning rate. However, certain parameter modifications trigger a re-initialization, e.g. setting the learning rate using `set_params` will re-initialize the optimizer. This can be problematic for optimizers that have a state, such as `RMSprop` or `Adam`.

So either 
- we enable the user to set parameters such as the learning rate without re-initializing the optimizer or
- we make this initialization visible.

ottonemo:
Attempt to resolve #349.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,10 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - [MNIST with scikit-learn and skorch][1811011230] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
-- Added the ability to register "virtual" parameters on the network to mark
-  them safe for setting without re-initialization. This is useful for specific
-  module, criterion or optimizer parameters such as the learning rate. As of now
-  you can do `net.set_params(lr=0.03)` or
+- Added an experimental API (`net._register_virtual_param`) to register "virtual"
+  parameters on the network with custom setter functions. (#369)
+- Setting parameters `lr`, `momentum`, `optimizer__lr`, etc. no longer resets 
+  the optmizer. As of now you can do `net.set_params(lr=0.03)` or
   `net.set_params(optimizer__param_group__0__momentum=0.86)` without triggering
   a re-initialization of the optimizer (#369)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - [MNIST with scikit-learn and skorch][1811011230] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- Added the ability to register "virtual" parameters on the network to mark
+  them safe for setting without re-initialization. This is useful for specific
+  module, criterion or optimizer parameters such as the learning rate. As of now
+  you can do `net.set_params(lr=0.03)` or
+  `net.set_params(optimizer__param_group__0__momentum=0.86)` without triggering
+  a re-initialization of the optimizer (#369)
 
 [1810251445]: https://colab.research.google.com/github/dnouri/skorch/blob/master/notebooks/Basic_Usage.ipynb
 [1810261633]: https://colab.research.google.com/github/dnouri/skorch/blob/master/notebooks/Advanced_Usage.ipynb
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -32,6 +32,7 @@ from skorch.utils import to_numpy
 from skorch.utils import to_tensor
 from skorch.utils import train_loss_score
 from skorch.utils import valid_loss_score
+from skorch.setter import optimizer_setter
 
 
 # pylint: disable=too-many-instance-attributes
@@ -493,27 +494,10 @@ class NeuralNet(object):
             for key, val in virtual_kwargs.items():
                 if not fnmatch.fnmatch(key, pattern):
                     continue
-                fn(key, val)
+                fn(self, key, val)
 
     def initialize_virtual_params(self):
         self.virtual_params_ = {}
-
-    def _set_lr_param(self, param, lr, optimizer_attr='optimizer_', optimizer_name='optimizer'):
-        # Extract param group from explicit param group parameter name (if needed)
-        pat = optimizer_name + '__param_groups__(?P<param_group>[0-9])__lr'
-        mat = re.compile(pat).fullmatch(param)
-        param_group = int(mat.groupdict().get('param_group')) if mat else 'all'
-
-        if param_group == 'all':
-            groups = getattr(self, optimizer_attr).param_groups
-        else:
-            groups = [getattr(self, optimizer_attr).param_groups[param_group]]
-
-        for group in groups:
-            group['lr'] = lr
-
-        if param == 'lr':
-            self.lr = lr
 
     def initialize_optimizer(self, triggered_directly=True):
         """Initialize the model optimizer. If ``self.optimizer__lr``
@@ -544,8 +528,8 @@ class NeuralNet(object):
         self.optimizer_ = self.optimizer(*args, **kwargs)
 
         self._register_virtual_param(
-            ['optimizer__param_groups__*__lr', 'optimizer__lr', 'lr'],
-            self._set_lr_param
+            ['optimizer__param_groups__*__*', 'optimizer__*', 'lr'],
+            optimizer_setter,
         )
 
     def initialize_history(self):
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -21,6 +21,7 @@ from skorch.dataset import uses_placeholder_y
 from skorch.exceptions import DeviceWarning
 from skorch.exceptions import NotInitializedError
 from skorch.history import History
+from skorch.setter import optimizer_setter
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import TeeGenerator
 from skorch.utils import duplicate_items
@@ -32,7 +33,6 @@ from skorch.utils import to_numpy
 from skorch.utils import to_tensor
 from skorch.utils import train_loss_score
 from skorch.utils import valid_loss_score
-from skorch.setter import optimizer_setter
 
 
 # pylint: disable=too-many-instance-attributes
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -3,7 +3,6 @@
 import fnmatch
 from itertools import chain
 from collections import OrderedDict
-import re
 import tempfile
 import warnings
 
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -3,6 +3,8 @@
 import fnmatch
 from itertools import chain
 from collections import OrderedDict
+from functools import partial
+import re
 import tempfile
 import warnings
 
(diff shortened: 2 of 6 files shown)


skorch/net.py
```diff
@@ -456,6 +458,30 @@ def initialize_module(self):
         self.module_ = module.to(self.device)
         return self
 
+    def register_runtime_param(self, param_patterns, fn):
+        self.runtime_params_ = getattr(self, 'runtime_params_', {})
+        if not isinstance(param_patterns, list):
```
thomasjpfan: Since `register_runtime_param` is an internal api, it could be good to force `param_patterns` to just always be a list. (Reduces the lines of code)

skorch/net.py
```diff
@@ -469,6 +495,16 @@ def initialize_optimizer(self):
 
         self.optimizer_ = self.optimizer(*args, **kwargs)
 
+        self.register_runtime_param(
+            ['optimizer__lr', 'lr'],
+            partial(self._set_lr, optimizer_attr='optimizer_', param_group=0)
+        )
+
+        self.register_runtime_param(
+            'optimizer__param_groups__(?P<param_group>[0-9]*)__lr',
```
thomasjpfan: This introduces a new interface for targeting parameters in a parameter group. Maybe this should be tested, with a simple case such as: 

```python
optimizer__param_groups=[
        ('rnn*.period', {'lr': 0.3, 'momentum': 0}),
        ('rnn0', {'lr': 0.1})
]
```

And then setting `optimizer__param_groups__0__lr`.

Currently, this only coverts `lr` and other parameters such as `momentum`. The regex can most likely be expanded to cover any keyword, such as `momentum`.

skorch/net.py
```diff
@@ -1225,6 +1263,12 @@ def set_params(self, **kwargs):
 
         BaseEstimator.set_params(self, **normal_params)
 
+        # Keys that were applied as runtime params should not trigger
+        # re-initialization by definition, so remove them.
+        used_keys = self.apply_runtime_params(kwargs)
+        for key in list(used_keys):
```
thomasjpfan:
```suggestion
        for key in used_keys:
```

ottonemo:
Just to make sure: this will be an internal API for the time being since it is not clear whether this approach is viable. Once it is clear that this approach covers most cases we can think about the user-facing side more.

skorch/net.py
```diff
@@ -473,6 +476,45 @@ def initialize_module(self):
...
+        for pattern, fn in self.virtual_params_.items():
+            for key, val in virtual_kwargs.items():
+                if not fnmatch.fnmatch(key, pattern):
+                    continue
+                fn(key, val)
+
+    def initialize_params(self):
+        self.virtual_params_ = {}
+
+    def _set_lr_param(self, param, lr, optimizer_attr='optimizer_', optimizer_name='optimizer'):
```
thomasjpfan: Since the virtual parameters may expand to objects, maybe `_set_lr_param` could be moved to `skorch.utils`. Otherwise, additional functions for other objects would expand the amount of code in `NeutralNet`. This would change the name to something like:

```python
def _set_optimizer_lr_param(net, param, lr, ...)
```

Do you see `optimizer_attr` and `optimizer_name` to be other values other than the default?

skorch/net.py
```diff
@@ -473,6 +476,45 @@ def initialize_module(self):
...
+            self.virtual_params_[pattern] = fn
+
+    def _apply_virtual_params(self, virtual_kwargs):
+        for pattern, fn in self.virtual_params_.items():
+            for key, val in virtual_kwargs.items():
+                if not fnmatch.fnmatch(key, pattern):
+                    continue
+                fn(key, val)
+
+    def initialize_params(self):
```
thomasjpfan: `initialize_virtual_params` would be a little clearer.

ottonemo:
The ambiguity is a bit on purpose since virtual params do more than just being "pre init". In fact, you can set virtual params whenever you want it is just not treated specially (e.g. no init will be issued). The sole purpose of plain virtual parameters is to be exempt from the initialization scheme in cases where you don't want a re-initialization (e.g. in case of the learning rate or optimizer momentum, etc.).

In the future you could add virtual parameters for your module which don't require re-initialization, e.g. changing dropout probabilities on the fly:
```python
net = Net(
    module, 
    virtual_params=['module__p_dropout'],
)
net.set_params(module__p_dropout=0.2)
net.partial_fit(X, y)
net.set_params(module__p_dropout=0.5)
net.partial_fit(X, y)
# [...]

```

skorch/net.py
ottonemo: > Since the virtual parameters may expand to objects, maybe `_set_lr_param` could be moved to `skorch.utils`. Otherwise, additional functions for other objects would expand the amount of code in `NeutralNet`. This would change the name to something like:
> 
> ```python
> def _set_optimizer_lr_param(net, param, lr, ...)
> ```

I think it might make sense to move these definitions to their own module since there may be more to come (criteria might want this ability, optimizer alone has several attributes like learning rate, momentum, averaging parameters).

> Do you see `optimizer_attr` and `optimizer_name` to be other values other than the default?

Yes, in the case when the user defines their own optimizer in `initialize_optimizer` this might happen.

ottonemo:
I addressed the comments and changed the setter to support more parameters than just the learning rate. Please review again.

skorch/tests/test_net.py
```diff
@@ -858,24 +858,23 @@ def test_setting_lr_after_init_reflected_in_optimizer(
...
-             "parameters were re-set: momentum.")
-        ),
         (
             {'module__input_units': 12, 'module__hidden_units': 34,
              'optimizer__momentum': 0.56},
             ("Re-initializing module because the following "
              "parameters were re-set: hidden_units, input_units.\n"
-             "Re-initializing optimizer because the following "
-             "parameters were re-set: momentum.")
+             "Re-initializing optimizer.")
```
benjamin-work: Would we not expect this to not trigger optimizer re-initialization anymore?

skorch/setter.py
```diff
@@ -0,0 +1,42 @@
...
+    ):
+    """Handle setting of optimizer parameters such as learning rate and
+    parameter group specific parameters such as momentum.
+    """
+    if param == 'lr':
+        param_group = 'all'
+        param_name = 'lr'
+        net.lr = value
+    else:
+        pat_1 = '__param_groups__(?P<group>[0-9])__(?P<name>.+)'
```
benjamin-work: I find this whole section a little bit complicated. Perhaps it would already help if it is moved to a sub-function with a good name.

skorch/utils.py
```diff
@@ -466,3 +466,16 @@ def get_map_location(target_device, fallback_device='cpu'):
             ), DeviceWarning)
         map_location = torch.device(fallback_device)
     return map_location
+
+
+def set_optimizer_param(optimizer, param_group, param_name, value):
```
benjamin-work: Would this function not be a better fit for `setter.py`?

skorch/setter.py
```diff
@@ -0,0 +1,42 @@
+"""Setter functions for virtual params such as ``optimizer__lr``."""
+import re
+
+from skorch.utils import set_optimizer_param
+
+
+def optimizer_setter(
+        net, param, value, optimizer_attr='optimizer_', optimizer_name='optimizer'
```
benjamin-work: When would we set other values for `optimizer_attr` and `optimizer_name`? Perhaps a docstring complete with parameter description would help here.

skorch/net.py
```diff
@@ -32,6 +32,7 @@
 from skorch.utils import to_tensor
 from skorch.utils import train_loss_score
 from skorch.utils import valid_loss_score
+from skorch.setter import optimizer_setter
```
benjamin-work: import is not alphabetical

skorch/tests/test_setter.py
```diff
@@ -0,0 +1,71 @@
...
+        setter(net_optim_dummy, 'lr', new_lr)
+
+        assert net_optim_dummy.lr == new_lr
+
+    def test_wrong_name_raises(self, setter, net_optim_dummy):
+        # should be 'param_groups' instead
+        param = 'optimizer__param_group__0__lr'
+        value = 0.1
+        with pytest.raises(AttributeError):
+            setter(net_optim_dummy, param, value)
```
benjamin-work: Could you test the error message as well?

skorch/utils.py
ottonemo: I wouldn't leave the setter exposed as a public function in `setter.py` since it might become confusing what a 'setter' is when mixing `set_` and `thing_setter` functions. Therefore I marked it as protected and moved it there to avoid this confusion.

ottonemo:
@benjamin-work I addressed your change requests, please review again.

ottonemo:
I agree that we should probably document the logic behind the param group addressing but I don't think that a user *needs* to know about param groups when setting optimizer parameters. Consider changing the learning rate or momentum: as long as the user does not use param groups, they can simply do

```python
net.set_params(optimizer__lr=0.3, optimizer__momentum=0.98)
```

and it just works (i.e. it sets these parameters for all param groups). I would argue that as soon as the user employs param groups they also know how to address and use them.

skorch/setter.py
```diff
@@ -0,0 +1,70 @@
+"""Setter functions for virtual params such as ``optimizer__lr``."""
+import re
+
+
+def _extract_param_name_and_group(optimizer_name, param):
```
benjamin-work: Does this function have any application besides for the optimizer? If not, would it not be better to have `optimizer` somewhere in the name?

skorch/tests/test_setter.py
```diff
@@ -0,0 +1,76 @@
+"""Tests for virtual parameter setters"""
+from unittest.mock import Mock
+
+import pytest
+
+
+def net_dummy():
```
benjamin-work: why not a fixture (perhaps move to conftest)

skorch/tests/test_setter.py
```diff
@@ -0,0 +1,76 @@
...
+    from skorch import NeuralNet
+    net = Mock(spec=NeuralNet)
+    net.lr = 0.01
+    return net
+
+
+def optimizer_dummy():
```
benjamin-work: see above

skorch/tests/test_net.py
```diff
@@ -1971,3 +1973,49 @@ def test_predefined_split(self, net_cls, module_cls, data):
...
+        lr_pgroup_1 = 0.2
+        lr_pgroup_0_new = 0.3
+        lr_pgroup_1_new = 0.4
+
+        net = net_cls(
+            module_cls,
+            lr=lr_pgroup_1,
+            optimizer__param_groups=[
+                ('sequential.0.*', {'lr': lr_pgroup_0}),
+            ])
```
benjamin-work: do we really need `max_epochs=10` for this test?

skorch/tests/test_net.py
```diff
@@ -858,23 +858,23 @@ def test_setting_lr_after_init_reflected_in_optimizer(
...
-        ),
-        (
-            {'module__input_units': 12, 'module__hidden_units': 34, 'lr': 0.56},
+            {'module__input_units': 12, 'module__hidden_units': 34,
+             'optimizer__momentum': 0.56},
             ("Re-initializing module because the following "
              "parameters were re-set: hidden_units, input_units.\n"
-             "Re-initializing optimizer because the following "
-             "parameters were re-set: lr.")
+             "Re-initializing optimizer.")
```
benjamin-work: I would propose changing the error message to `"Re-initializing optimizer (triggered by another component)."` in this case. This should be straightforward to add [here](https://github.com/dnouri/skorch/blob/1700cbb22e2c49b1a6431a27d4f17e3bdfe54b09/skorch/net.py#L453).

CHANGES.md
```diff
@@ -14,6 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - [MNIST with scikit-learn and skorch][1811011230] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- Added the ability to register "virtual" parameters on the network to mark
+  them safe for setting without re-initialization. This is useful for specific
```
benjamin-work: Maybe add that this feature is experimental and a "public" API for user-defined parameters will be added later?

I would also split this message:
1) Add this experimental feature...
2) Setting lr, momentum etc. no longer resets optimizer.

That fact that the two are related does not matter that much from the user's point of view.

skorch/tests/test_setter.py
ottonemo: @fixture done.
@move I think these functions are very specific to the test case. I'll move it to the test class instead.

skorch/tests/test_net.py
ottonemo: But that would be the default then. Whenever the optimizer is re-initialized it would say "by another component", even if the user calls `initialize_optimizer` directly. Do we want that? It's technically correct but are there cases where this might be confusing? I couldn't find any but I thought I bring this up first.

benjamin-work: Since `triggered_directly=True` is the default, the proposed message would not trigger by default but only when `initialize_optimizer` is triggered indirectly, so this should be good or am I missing something?

That said, when the user calls `initialize_optimizer` directly, it will trigger the other case ("...because the following parameters were re-set: ..."), which may be confusing?

benjamin-work:
@ottonemo LGTM, the point above with the output message can also be addressed later.

PR #372, opened 2018-10-26
Bugfix: Reinitialize optimizer when set_params is called with lr.

benjamin-work:
Previously, when the net was already initialized, calling
set_params(lr=new_lr) didn't trigger a re-initialization because the
method only looked for optimizer__* parameters. However, lr is
indirectly also an optimizer parameter, so setting it must
reinitialize the optimizer.


githubnemo:
LGTM, can you add a note in the `CHANGES.md` file?

PR #374, opened 2018-10-26
Feature/better user facing messages

benjamin-work:
Depends on #372.

Implements:

1. Replace placeholder error message with useful one.

This is a safeguard that should usually never be reached but still
there should be a useful error message.

2. Improve user-facing messages when re-initialization occurs.

Specifically, alert the user to the parameters that were responsible
for re-initialization. As a by-product, this change also prevents the
optimizer from being re-initialized twice when both it and the module
changed.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -8,7 +8,6 @@ import warnings
 
 import numpy as np
 from sklearn.base import BaseEstimator
-from sklearn.utils import deprecated
 import torch
 from torch.utils.data import DataLoader
 
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -431,6 +431,24 @@ class NeuralNet(object):
         self.criterion_ = self.criterion(**criterion_params)
         return self
 
+    def _format_reinit_msg(self, name, kwargs=None, triggered_directly=True):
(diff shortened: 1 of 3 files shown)


skorch/net.py
```diff
@@ -449,24 +448,48 @@ def initialize_module(self):
...
 
+        Parameters
+        ----------
+        triggered_directly : bool (default=True)
+
+          Only relevant when optimizer is re-initialized.
+          Initialization of the optimizer can be triggered directly
+          (e.g. when lr was changed) or indirectly (e.g. when the
+          module was re-initialized). If the former happens, the user
+          should receive a message alterting them to the parameters
```
taketwo:
```suggestion
          should receive a message alerting them about the parameters
```

skorch/net.py
```diff
@@ -449,24 +448,48 @@ def initialize_module(self):
...
+          that caused the re-initialization. If the latter happens,
```
taketwo: This last sentence is... strange :D
I'm not sure if it's needed at all.

ottonemo:
LGTM, can you document this addition in `CHANGES.md`?

PR #375, opened 2018-10-26
DOC: Add links to colab in notebook

thomasjpfan:
Adds links to colab and github in the beginning of notebooks.


benjamin-work:
Thanks!

PR #379, opened 2018-10-29
ENH: Adds TeeGenerator for named_parameters

thomasjpfan:
Fixes https://github.com/dnouri/skorch/issues/378

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,8 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Basic usage notebook][1810251445] now runs on Google Colab
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
-- The `on_grad_computed` callback function will yield an iterable for
-  `named_parameters` only when it is used.
+- `named_parameters` only when it is used to reduce the run-time overhead of the call (#379)
 
 
 [1810251445]: https://colab.research.google.com/github/dnouri/skorch/blob/master/notebooks/Basic_Usage.ipynb
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Basic usage notebook][1810251445] now runs on Google Colab
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- The `on_grad_computed` callback function will yield an iterable for
+  `named_parameters` only when it is used.
 
 
 [1810251445]: https://colab.research.google.com/github/dnouri/skorch/blob/master/notebooks/Basic_Usage.ipynb
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Basic usage notebook][1810251445] now runs on Google Colab
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
-- `named_parameters` only when it is used to reduce the run-time overhead of the call (#379)
+- The `on_grad_computed` callback function will yield an iterable for `named_parameters` only when it is used to reduce the run-time overhead of the call (#379)
 
 
 [1810251445]: https://colab.research.google.com/github/dnouri/skorch/blob/master/notebooks/Basic_Usage.ipynb
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -1829,6 +1829,19 @@ class TestNeuralNet:
         net = net_cls(module_cls, max_epochs=1, callbacks=[mock_cb])
         net.fit(*data)
 
+    def test_callback_on_grad_computed(self, net_cls, module_cls, data):
+
+        module = module_cls()
+        expected_names = set(name for name, _ in module.named_parameters())
+
+        def on_grad_computed(*args, named_parameters, **kwargs):
+            names = set(name for name, _ in named_parameters)
+            assert expected_names == names
+
+        mock_cb = Mock(on_grad_computed=on_grad_computed)
+        net = net_cls(module, max_epochs=1, callbacks=[mock_cb])
+        net.fit(*data)
+
     @pytest.mark.parametrize('training', [True, False])
     def test_no_grad_during_evaluation_unless_training(
             self, net_cls, module_cls, data, training):
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -500,15 +500,10 @@ class TestTeeGenerator:
 
         def list_gen():
             yield from expected_list
-        mock = Mock(return_value=list_gen())
-        lazy_gen = lazy_generator_cls(mock)
-
-        assert mock.call_count == 0
+        lazy_gen = lazy_generator_cls(list_gen())
 
         first_return = list(lazy_gen)
         second_return = [item for item in lazy_gen]
 
         assert first_return == expected_list
-        assert mock.call_count == 1
         assert second_return == expected_list
-        assert mock.call_count == 1
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -1,4 +1,5 @@
 """Test for utils.py"""
+from unittest.mock import Mock
 
 import numpy as np
 import pytest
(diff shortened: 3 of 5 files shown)


benjamin-work:
I have some comments on this:

1. I'm not sure if `LazyGenerator` is the right name, because "generator" already implies "lazy". Furthermore, it's not so much about laziness and more about caching.

2. I have a strong feeling that such a solution already exists, though a quick search did not reveal a ready solution.

3. The generator creates the whole list right away when called first. I'm not sure if that's a problem but it is possible to avoid it, e.g.:

```
from itertools import tee
class BetterName:
    def __init__(self, gen):
        self.gen = gen
        self.gen_ = None

    def __iter__(self):
        if self.gen_ is None:
            self.gen_, it = tee(self.gen())
        else:
            it, *_ = tee(self.gen_, 1)
        yield from it
```

4. Did you measure how much the overhead is reduced?

ottonemo:
> 1. I'm not sure if `LazyGenerator` is the right name, because "generator" already implies "lazy". Furthermore, it's not so much about laziness and more about caching.

Well, both actually. The function is evaluated lazily and the generator is cached (and/or multiplexed).
But since the lazy part is not actually the important part, it shouldn't influence the name.

> 3. The generator creates the whole list right away when called first. I'm not sure if that's a problem but it is possible to avoid it, e.g.:
> 
> 
> ```
> from itertools import tee
> class BetterName:
>     def __init__(self, gen):
>         self.gen = gen
>         self.gen_ = None
> 
>     def __iter__(self):
>         if self.gen_ is None:
>             self.gen_, it = tee(self.gen())
>         else:
>             it, *_ = tee(self.gen_, 1)
>         yield from it
> ```

You need to always copy the generator:
```python
it, *_ = tee(self.gen_, 1)
```
will fail if the generator is evaluated more than twice. Better:
```python
from itertools import tee
class LazyGenerator_:
    def __init__(self, gen):
        self.gen = gen()

    def __iter__(self):
        self.gen, it = tee(self.gen)
        yield from it
```
(this way you don't need the callback at all)

benjamin-work:
> will fail if the generator is evaluated more than twice

What do you mean by that?

> Better:

Yes, this also works. I wanted to avoid calling the generator during `__init__` because that is what the original code does, but I don't know whether it's a requirement.

ottonemo:
> > will fail if the generator is evaluated more than twice
> 
> What do you mean by that?

Sorry, should have written an example:
```python
g = BetterName(lambda: [1,2,3])

g1 = list(g) # == [1,2,3]
g2 = list(g) # == [1,2,3]
g3 = list(g) # == [] <- bad
```

> > Better:
> 
> Yes, this also works. I wanted to avoid calling the generator during `__init__` because that is what the original code does, but I don't know whether it's a requirement.

I think it is no requirement. While there might be an overhead by simply creating the generator yielded by `named_parameters` it is probably less than evaluating the generator fully (as the old code did). Of course this should be evaluated in a benchmark.

benjamin-work:
Interesting, I didn't see this.

If initialization of the generator during `__init__` is not desired, this should still work:

```
class LazyGenerator:
    def __init__(self, gen):
        self.gen = gen
        self.gen_ = None

    def __iter__(self):
        if self.gen_ is None:
            self.gen_ = self.gen()
        self.gen_, it = tee(self.gen_)
        yield from it
```

thomasjpfan:
Here are the results of the microbenchmark comparing the `list` version and the `tee` version:

```python
from itertools import tee

class ListLazyGenerator:
    def __init__(self, gen): 
        self.gen = gen
        self.gen_ = None
    def __iter__(self): 
        if self.gen_ is None:
            self.gen_ = list(self.gen())
        yield from self.gen_

class TeeLazyGenerator:
    def __init__(self, gen):
        self.gen = gen
        self.gen_ = None

    def __iter__(self):
        if self.gen_ is None:
            self.gen_ = self.gen()
        self.gen_, it = tee(self.gen_)
        yield from it

class TeeCallInInitLazyGenerator:
    def __init__(self, gen):
        self.gen = gen()

    def __iter__(self):
        self.gen, it = tee(self.gen)
        yield from it
```

## Small Module
```python
import torchvision
m = torchvision.models.squeezenet1_0()

# No generator
%timeit list(m.named_parameters())
# 214 µs ± 10.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

list_lazy = ListLazyGenerator(m.named_parameters)
%timeit list(list_lazy)
# 3.04 µs ± 61.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

tee_lazy = TeeLazyGenerator(m.named_parameters)
%timeit list(tee_lazy)
# 3.51 µs ± 87.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

tee_call_in_init = TeeCallInInitLazyGenerator(m.named_parameters)
%timeit list(tee_call_in_init)
# 3.29 µs ± 47 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

## Big Module
```python
import torchvision
m = torchvision.models.resnet152()

# No generator
%timeit list(m.named_parameters())
# 1.67 ms ± 75.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

list_lazy = ListLazyGenerator(m.named_parameters)
%timeit list(list_lazy)
# 22.5 µs ± 405 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tee_lazy = TeeLazyGenerator(m.named_parameters)
%timeit list(tee_lazy)
# 24.8 µs ± 1.36 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tee_call_in_init = TeeCallInInitLazyGenerator(m.named_parameters)
%timeit list(tee_call_in_init)
# 23.8 µs ± 253 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

I think `list` is fastest with a tradeoff in memory usage. Given the microbenchmark, `tee` with the generator called in `init` would be my preferred solution.

thomasjpfan:
This PR was updated to use the tee version of the generator, and renames the class to `TeeGenerator`.

skorch/tests/test_utils.py
```diff
@@ -485,3 +486,28 @@ def type_truth_table():
...
+        expected_list = [1, 2, 3]
+
+        def list_gen():
+            yield from expected_list
+        mock = Mock(return_value=list_gen())
+        lazy_gen = lazy_generator_cls(mock)
+
+        assert mock.call_count == 0
+
+        first_return = list(lazy_gen)
```
benjamin-work: The test would be a little clearer if the call count would also be tested immediately after the first call (i.e. here).

skorch/utils.py
```diff
@@ -466,3 +467,19 @@ def get_map_location(target_device, fallback_device='cpu'):
             ), DeviceWarning)
         map_location = torch.device(fallback_device)
     return map_location
+
+
+class TeeGenerator:
+    """Stores a function that produces a generator and calls list on the
```
benjamin-work: The docstring is no longer up-to-date. Also, could you add 2 newlines after the first sentence?

skorch/tests/test_net.py
```diff
@@ -1829,6 +1829,19 @@ def check_grad(*args, loss, training, **kwargs):
         net = net_cls(module_cls, max_epochs=1, callbacks=[mock_cb])
         net.fit(*data)
 
+    def test_callback_on_grad_computed(self, net_cls, module_cls, data):
```
benjamin-work: Could you shortly explain why you added this test?

thomasjpfan: This PR changes the `self.notify('on_grad_computed',  ...)` to use the `TeeGenerator`. This test makes sure that the named_parameters are still correctly passed to the callback.

skorch/utils.py
```diff
@@ -466,3 +467,19 @@ def get_map_location(target_device, fallback_device='cpu'):
...
+    generator on the first iteration. All future iteration calls will
+    iterate over the list.
+    """
+    def __init__(self, gen):
+        self.gen = gen
+        self.gen_ = None
+
+    def __iter__(self):
+        if self.gen_ is None:
+            self.gen_ = self.gen()
```
ottonemo: I think the TeeGenerator should use a generator, not a callable. The overhead is negligible but it allows for easy passing of partially consumed generators (e.g. when using models with two parameter sets) and is more intuitive.

CHANGES.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Basic usage notebook][1810251445] now runs on Google Colab
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- The `on_grad_computed` callback function will yield an iterable for
+  `named_parameters` only when it is used.
```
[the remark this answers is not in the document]
thomasjpfan: I am starting to see how you like the CHANGELOG to be worded. 😅

skorch/utils.py
```diff
@@ -466,3 +467,16 @@ def get_map_location(target_device, fallback_device='cpu'):
             ), DeviceWarning)
         map_location = torch.device(fallback_device)
     return map_location
+
+
+class TeeGenerator:
+    """Stores a generator and calls ``tee`` on it to create new generators
+    when ``TeeGenerator`` is iterated over.
```
ottonemo:
```suggestion
    when ``TeeGenerator`` is iterated over to let you iterate over the given generator more than once.
```

benjamin-work:
@thomasjpfan Would you mind looking at the open comments?

thomasjpfan:
> What do you think about my suggestion to adjust the unit test?

`TeeGenerator` no longer calls the function, it now accepts a generator. The test was refactored to not use mocks, thus removing the `call_count` assertion.

benjamin-work:
> `TeeGenerator` no longer calls the function, it now accepts a generator. The test was refactored to not use mocks, thus removing the `call_count` assertion.

Ah okay, I was looking at the outdated state shown by github. With the given change in the test, does the function name still make sense, though?

thomasjpfan:
The test was renamed to `test_returns_copies_of_generator`.

benjamin-work:
Thanks.

PR #381, opened 2018-11-03
ENH: Early exit for train_loss_score and valid_loss_score already in history

thomasjpfan:
When `train_loss_score` or `valid_loss_score` is set to be the `scoring` function, it already assumes that `train_loss` or `valid_loss` is already in the batch history. If the metric is not in the history, this will raise an `KeyError` which is ignored.

This PR checks for these scoring functions and returns early.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - [MNIST with scikit-learn and skorch][1811011230] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- Early exits `BatchScoring` when using `train_loss_score` or
+  `valid_loss_score`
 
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Advanced usage notebook][1810261633] now runs on Google Colab
 - [MNIST with scikit-learn and skorch][1811011230] now runs on Google Colab
 - Better user-facing messages when module or optimizer are re-initialized
+- Early exits `BatchScoring` when using `train_loss_score` or
```
ottonemo:
```suggestion
- Reduce overhead of `BatchScoring` when using `train_loss_score` or `valid_loss_score` by skipping superfluous inference step (#381)
```

CHANGES.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `valid_loss_score`
```
ottonemo:
```suggestion
```

benjamin-work:
Thanks.

PR #390, opened 2018-11-20
Feature: quick CLIs

benjamin-work:
This feature adds functionality to build quick CLIs for scripts based on skorch. The intent is to encourage users to adopt skorch by reducing the boilerplate required to make experiments reproducible.

This basically leverages the features of [fire](https://github.com/google/python-fire) to automatically create CLIs. In combination with skorch, almost all possible model parameters are immediately available via the command line, without all the boilerplate typically required to parse the arguments. This way, things like `max_epochs`, `lr`, `device`, etc. are available for free. Moreover, all arguments on the module are also available as usual (say `module__num_units`).

Furthermore, if there are docstrings in the numpydoc format, those can be displayed via `--help`, again without any boilerplate code.

A restriction right now is passing complex python objects via the command line. Things like `--module__nonlin 'torch.nn.RReLU(0.1, upper=0.4)'` work, but if the object contains other non-primitive objects, parsing will fail.

@thomasjpfan @taketwo Please also share your views on whether you find this feature useful.

### DIFF
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -5,6 +5,7 @@ Helper
 This module provides helper functions and classes for the user. They
 make working with skorch easier but are not used by skorch itself.
 
+
 SliceDict
 ---------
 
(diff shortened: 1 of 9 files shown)


taketwo:
I have my own utility class that helps me to minimize CLI boilerplate. It's not as advanced when it comes to automatic discovery of parameters, so I need to manually add module parameters that are of interest. On the other hand, it helps me organize my `runs` directory, automatically creating directories/filenames for different experiments based on the parameters.

I guess what I want to say is, sure, a CLI helper is great to have and can give a productivity boost. Unfortunately I don't have enough time to play with your proposed implementation now. Nevertheless, it looks good and relies on an established third-party library, so I only see benefits in including this.

docs/user/helper.rst
```diff
@@ -16,3 +17,140 @@ length of the arrays and not the number of keys, and you get a
...
+
+This even works if your neural net is part of an sklearn pipeline, in
+which case the help extends to all other estimators of your pipeline.
+
+Restrictions
+^^^^^^^^^^^^
+
+Almost all arguments should work out of the box. Therefore, you get
+command line arguments for the number of epochs, learning rate, batch
+size, etc. for free. Morevoer, you can access the module paremeters
```
taketwo:
```suggestion
size, etc. for free. Moreover, you can access the module parameters
```

docs/user/helper.rst
```diff
@@ -16,3 +17,140 @@ length of the arrays and not the number of keys, and you get a
...
+        my_model = parsed(my_model)
+
+        my_model.fit(X, y)
+
+
+    if __name__ == '__main__':
+        fire.Fire(main)
+
+
+This even works if your neural net is part of an sklearn pipeline, in
```
taketwo:
```suggestion
This even works if your neural net is part of a sklearn pipeline, in
```

examples/cli/README.md
```diff
@@ -0,0 +1,122 @@
...
+which case the help extends to all other estimators of your pipeline.
+
+For more information on how to use fire, follow [this
+link](https://github.com/google/python-fire).
+
+## Restrictions
+
+Almost all arguments should work out of the box. Therefore, you get
+command line arguments for the number of epochs, learning rate, batch
+size, etc. for free. Morevoer, you can access the module paremeters
```
taketwo: Same here. I wonder if this documentation duplication can be avoided somehow?

docs/user/helper.rst
benjamin-work: I pronounce it "ess-kay-learn", therefore I would use "an". Or do most people "translate" it to "scy-kit-learn"?

examples/cli/README.md
benjamin-work: Yes, I wondered about that too. I wanted to avoid sending readers this way and that way to collect all relevant information, hence the duplication. Any ideas?

benjamin-work:
Thanks for the feedback.

> it helps me organize my `runs` directory, automatically creating directories/filenames for different experiments based on the parameters

This sounds like something that could be combined with the current feature. Do you think that would be a good idea?

skorch/cli.py
```diff
@@ -0,0 +1,335 @@
...
+    set_params : callable
+      If --help is not in the options, apply command line arguments to
+      the estimator and return it.
+
+    """
+    defaults = defaults or {}
+
+    def print_help_and_exit(estimator):
+        print_help(estimator, defaults=defaults)
+        sys.exit()
```
ottonemo: Maybe use `exit(1)` to indicate non-normal termination?

examples/cli/README.md
```diff
@@ -0,0 +1,144 @@
...
+python train.py pipeline --help
+```
+
+### Training a Model
+
+Run
+
+```bash
+python train.py net  # only the net
+python train.py pipeline  # net with pipeline
```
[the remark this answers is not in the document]
benjamin-work: It tells you to use `net` or `pipeline` option, though this is under the control of fire.

skorch/cli.py
```diff
@@ -0,0 +1,335 @@
...
+    if not match:
+        return s
+
+    # ideally, we would like to replace the 'default' group directly
+    # but I haven't found a way to do this
+    try:
+        i, j = _get_span(match)
+    except IndexError:
+        pass
+    return '{}{}{}'.format(s[:i], new_value, s[j:])
```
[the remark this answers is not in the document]
benjamin-work: It seems not to be possible to pass `group` to a `span` call.

skorch/cli.py
benjamin-work: But should the error code not be 0 for a normal help option?

ottonemo: Well, not sure, the usual tools don't seem to agree on that issue:
```
$ cat -h 2>/dev/null; echo $?
1
$ sed -h 2>/dev/null; echo $?
1
$ dd -h 2>/dev/null; echo $?
1
$ grep --help >/dev/null; echo $?
0
$ make -h >/dev/null; echo $?
0
$ bzcat -h; echo $?
0
$ bash --help; echo $?
0
```

But it might be good to exit with an error code for cases when deploying the script to distributed/automated training. An illegal option might be detected sooner if the script fails to execute due to an unknown option or a left-over `-h`.

benjamin-work:
@ottonemo pls review again

PR #391, opened 2018-11-22
RFC: Changes default fn_prefix in TrainEndCheckpoint to train_end_

thomasjpfan:
The default prefix for `TrainEndCheckpoint` is `final_`. This PR adjusts it to `train_end_` along with all the references to it.


benjamin-work:
This change makes sense. In theory, it could break backwards compatibility for some users. I guess that's not too big of a concern since this code was introduced not too long ago. But if the next skorch release is far away, more people could be affected by then. Do we want to do anything about that?

thomasjpfan:
The only good option I see is to leave the default as `final_` and issue a warning stating that the default will change in 0.5.0 or 0.6.0, depending when this PR gets released. I would hope a point release such as 0.4.1, so this default can change at 0.5.0.

thomasjpfan:
This PR was updated to keep the default as `final_` but issue a `FutureWarning` stating the default will change in `0.5.0`.

benjamin-work:
LGTM, @ottonemo what's your take?

ottonemo:
LGTM, please document the change in the change log :)

PR #400, opened 2018-12-04
ENH: Adds epoch_completed to history

### LINKED ISSUE #387: Support Ctrl+C in TrainEndCheckpoint
Continuing from PR https://github.com/dnouri/skorch/pull/360

When one presses Ctrl+C in the middle of an epoch, the latest epoch would most likely be in an incomplete state. When the checkpoint is reloaded and training starts again, I propose throwing away the incomplete epoch and start training on that epoch.

For example, if one Ctrl+C's during epoch 9, only the history up to epoch 9, i.e. 1-8, will be saved. All the batch data in epoch 9 would be discarded.

thomasjpfan:
Resolves #387 

When dealing with histories without the `epoch_completed` signal, this PR keeps the old behavior of just creating the new epoch in history. `test_on_epoch_begin_adds_new_epoch_when_completed_event_does_not_exist` tests for this behavior.

After this PR, the history will now include a `epoch_completed` flag that is set to False at the creation of the epoch. When the epoch finishes, then this flag is set to True.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -5,6 +5,7 @@ from itertools import chain
 from collections import OrderedDict
 import tempfile
 import warnings
+from contextlib import suppress
 
 import numpy as np
 from sklearn.base import BaseEstimator
@@ -296,13 +297,19 @@ class NeuralNet:
     # pylint: disable=unused-argument
     def on_epoch_begin(self, net,
                        dataset_train=None, dataset_valid=None, **kwargs):
-        self.history.new_epoch()
-        self.history.record('epoch', len(self.history))
+        epoch_completed = True
+        with suppress(KeyError):
+            epoch_completed = self.history[-1, 'epoch_completed']
+
+        if epoch_completed:
+            self.history.new_epoch()
+            self.history.record('epoch', len(self.history))
+            self.history.record('epoch_completed', False)
 
     # pylint: disable=unused-argument
     def on_epoch_end(self, net,
                      dataset_train=None, dataset_valid=None, **kwargs):
-        pass
+        self.history.record('epoch_completed', True)
 
     # pylint: disable=unused-argument
     def on_batch_begin(self, net,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reduce overhead of `BatchScoring` when using `train_loss_score` or `valid_loss_score` by skipping superfluous inference step (#381)
 - The `on_grad_computed` callback function will yield an iterable for `named_parameters` only when it is used to reduce the run-time overhead of the call (#379)
 - Default `fn_prefix` in `TrainEndCheckpoint` is now `train_end_` (#391)
+- Adds `epoch_completed` to history to signal when an epoch has completed without interruption. (#387)
 
 ### Fixed
 
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -145,7 +145,7 @@ class PrintLog(Callback):
 
         for key in sorted(keys):
             if not (
-                    (key in ('epoch', 'dur')) or
+                    (key in ('epoch', 'dur', 'epoch_completed')) or
                     (key in self.keys_ignored_) or
                     key.endswith('_best') or
                     key.startswith('event_')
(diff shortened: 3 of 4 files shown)


skorch/net.py
```diff
@@ -296,13 +297,19 @@ def on_train_end(self, net,
     # pylint: disable=unused-argument
     def on_epoch_begin(self, net,
                        dataset_train=None, dataset_valid=None, **kwargs):
-        self.history.new_epoch()
-        self.history.record('epoch', len(self.history))
+        epoch_completed = True
+        with suppress(KeyError):
+            epoch_completed = self.history[-1, 'epoch_completed']
```
benjamin-work: would `epoch_completed = self.history[-1].get('epoch_completed', False)` not suffice?

benjamin-work:
You chose to not add a new epoch in case `on_epoch_end` was not triggered. I wonder if this could cause trouble. E.g. say a user interrupts training and then calls another fit. The new information would be added to the last epoch of the previous training. Is this the intended effect?

benjamin-work:
You chose to not add a new epoch in case `on_epoch_end` was not triggered. I wonder if this could cause trouble. E.g. say a user interrupts training and then calls another fit. The new information would be added to the last epoch of the previous training. Is this the intended effect?

benjamin-work:
You chose to not add a new epoch in case `on_epoch_end` was not triggered. I wonder if this could cause trouble. E.g. say a user interrupts training and then calls another fit. The new information would be added to the last epoch of the previous training. Is this the intended effect?

benjamin-work:
You chose to not add a new epoch in case `on_epoch_end` was not triggered. I wonder if this could cause trouble. E.g. say a user interrupts training and then calls another fit. The new information would be added to the last epoch of the previous training. Is this the intended effect?

thomasjpfan:
That is the intended effect. If the epoch didn’t end, this PR assumes that it did not complete. This means if the user calls `partial_fit` after interrupting, it does not create a new epoch, but adds to the last epoch of the previous training history. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

benjamin-work:
> if the user calls partial_fit after interrupting, it does not create a new epoch, but adds to the last epoch of the previous training history

Okay, let's discuss if that is what's best. First of all, this will change the current behavior, so if code assumes that interrupting creates a new epoch, it will break (though this is very unlikely to really happen in the wild). 

Moreover, my intuition would have been that interrupting creates a new epoch (with the old one being incomplete). In particular, calling `partial_fit` will trigger a new `on_epoch_begin`, so this effect would be triggered a second time on the epoch and could potentially break things. Is there a particular reason why continuing with the last epoch is better?

thomasjpfan:
The motivation for this PR was to iron out what happens when `TrainEndCheckpoint` saves a model that is interrupted in the middle of an epoch. Currently, when training resumes, a new epoch is creating, leaving the old one incomplete.

>Moreover, my intuition would have been that interrupting creates a new epoch (with the old one being incomplete).

If this is the norm, then this PR overcomplicates things and should not be merged.

benjamin-work:
Okay, I reread the issue and looked at the changes and am still confused. Could you please repeat why starting a new epoch is problematic with `TrainEndCheckpoint`?

thomasjpfan:
When leaving an epoch incomplete, the indexing the history object looks like the following:

```python
from skorch.history import History

h = History()
h.new_epoch()
h.record("loss", 1)
h.new_epoch() # incomplete epoch
h.new_epoch()
h.record("loss", 3)

print(h[:, "loss"])
# [1, 3]

print(len(h))
# 3
```

When querying the `loss` the incomplete epoch will be ignored.  When a user graphs the loss, they would see that they are missing an epoch. To figure out where this incomplete epoch is, a user would need to write custom code to find it.

thomasjpfan:
Including nans would make it clear that the value is missing. The biggest downside is that changing the default would break backwards compatibility.

BenjaminBossan:
> The biggest downside is that changing the default would break backwards compatibility.

For sure we would need a deprecation period for that change. The current behaviour is a little bit surprising, though, so the change might be for the better. Overall, I guess that most code should be unaffected by the change.

thomasjpfan:
Closing this PR in favor of changing the behavior of adding nans when querying the history.

PR #401, opened 2018-12-06
[MRG] SliceDict's copy now returns a SliceDict

### LINKED ISSUE #388: SliceDict's copy returns a normal dict
```python
>>> type(SliceDict(foo=np.zeros((10, 5))).copy())
dict
```

The result should be another `SliceDict`. Perhaps there are even more methods that need fixing.

thomasjpfan:
Resolves #388

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Re-initialize optimizer when `set_params` is called with `lr` argument (#372)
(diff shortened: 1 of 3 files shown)


benjamin-work:
Thanks for this contribution. I've worked on this issue as well (#402). I should have put a note on the issue about that but I had falsely assumed that this would be a quick fix when it actually took me longer. The reason is that I had a look at some other methods that were broken and fixed them as well (`__eq__` in particular).

I believe my PR supersedes yours. Again, sorry for the trouble.

thomasjpfan:
That’s okay! Your PR looks good!

PR #402, opened 2018-12-07
Fixes bugs with copy and eq on SliceDict (#388).

benjamin-work:
* Calling .copy() now correctly returns another SliceDict instead of a
  dict.
* == should now work as long as values are numpy arrays or torch
  tensors.
* raise an error when `fromkeys` is called. It is generally not useful
  to use this method with SliceDicts.

### DIFF
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -6,8 +6,11 @@ They should not be used in skorch directly.
 from functools import partial
 import warnings
 
+import numpy as np
+
 from skorch.utils import _make_split
 from skorch.utils import _make_optimizer
+from skorch.utils import is_torch_data_type
 
 
 class SliceDict(dict):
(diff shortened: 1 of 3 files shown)


benjamin-work:
Fixes #388 and then some.

skorch/helper.py
```diff
@@ -84,6 +89,46 @@ def __repr__(self):
...
+                    return False
+                continue
+
+            # numpy arrays
+            if isinstance(val, np.ndarray):
+                if not isinstance(val_other, np.ndarray):
+                    return False
+                if not (val == val_other).all():
+                    return False
+                continue
```
ottonemo: What about DataFrames? Can we use a more generic way for collections, e.g. `all(a == b)` for `array` and `DataFrame`s? (torch tensors don't work that way though)

benjamin-work: I thought about pandas `DataFrame`s but they probably don't matter. As I see it, they serve a similar purpose as a dict, namely to have named features (keys/columns) whose values may have different meanings. Therefore, I feel that you don't really need to mix both. In a way, a `DataFrame` is a dict with 1-dim arrays as values; `SliceDict` is a dict with n-dim arrays/tensors as values.

PR #403, opened 2018-12-07
[MRG] Warn user when using possibly wrong checkpoint monitor

### LINKED ISSUE #399: Warn user when using possibly wrong checkpoint monitor
When the user supplies a `monitor` value for the `Checkpoint` callback that does not end on `_best` and the history contains `<monitor>_best` we should warn the user about the fact that they probably meant the `_best` value.

thomasjpfan:
Resolves #399

Updates some tests to use the `<monitor>_best` syntax for `Checkpoints`'s `monitor` parameter.

### DIFF
--- a/skorch/tests/callbacks/test_training.py
+++ b/skorch/tests/callbacks/test_training.py
@@ -157,11 +157,11 @@ class TestCheckpoint:
 
         from skorch.callbacks import EpochScoring
         scoring = EpochScoring(
-            scoring=epoch_3_scorer, on_train=True)
+            scoring=epoch_3_scorer, on_train=True, lower_is_better=False)
 
         sink = Mock()
         cb = checkpoint_cls(
-            monitor='epoch_3_scorer',
+            monitor='epoch_3_scorer_best',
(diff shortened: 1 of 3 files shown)


skorch/tests/callbacks/test_training.py
```diff
@@ -170,22 +170,23 @@ def epoch_3_scorer(net, *_):
...
+             call(f_params='model_3_10.pt'),
              call(f_optimizer='optimizer_3_10.pt'),
              call(f_history='history.json')]
         )
-        assert sink.call_count == 1
-        assert all((x is False) for x in net.history[:2, 'event_cp'])
-        assert net.history[2, 'event_cp'] is True
-        assert all((x is False) for x in net.history[3:, 'event_cp'])
+        assert sink.call_count == 2
+        assert [True, False, True] + [False] * 7 == net.history[:, 'event_cp']
```
ottonemo: I think it would be beneficial to have a short explanation why the result looks like it does (what makes the result True in those cases?).

ottonemo:
Great, thanks!

PR #404, opened 2018-12-08
[MRG] Adds test for pytorch 1.0

thomasjpfan:
Originally, the tests only ran on python 3.6, since the `environment.yml` file fixes the python version to 3.6. This PR correctly tests for the different version of python.

`skorch/tests/test_net.py` had to be adjusted because of how `python 3.5` handles the ordering of parameters.

When newer versions of pytorch comes out we can add additional versions to the `env` item in the `.travis.yml` file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+dist: xenial
 python:
   - "3.7"
   - "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "3.6"
   - "3.5"
+env:
+  - PYTORCH_DEP=pytorch-cpu==0.4.1
 cache:
   apt: true
   timeout: 1000
@@ -26,6 +28,6 @@ install:
   - pip install tabulate
   - conda install --file=requirements-dev.txt
   - python setup.py install
-  - conda install -c pytorch 'pytorch>=1.0.0'
+  - conda install -c pytorch '${PYTORCH_DEP}'
 script:
   - pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,10 @@ install:
   - conda update -q conda
   - conda info -a
 
-  - conda create -y -n skorch-env python=${TRAVIS_PYTHON_VERSION} scikit-learn tqdm
+  - conda create -y -n skorch-env python=${TRAVIS_PYTHON_VERSION}
   - source activate skorch-env
-  - pip install tabulate
-  - conda install --file=requirements-dev.txt
+  - cat requirements.txt requirements-dev.txt > reqs.txt
+  - conda install --file=reqs.txt
   - python setup.py install
   - conda install -c pytorch pytorch-cpu==${PYTORCH_VERSION}
 script:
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for this addition, Thomas. Thankfully, pytorch 1.0 seems to work out of the box. I will have a closer look at this on Monday.

.travis.yml
```diff
@@ -1,7 +1,12 @@
 language: python
+dist: xenial
 python:
+  - "3.7"
   - "3.6"
   - "3.5"
```
ottonemo: Is there any relevance to specifying more than one python version?

.travis.yml
```diff
@@ -21,10 +26,11 @@ install:
   - conda update -q conda
   - conda info -a
 
-  - conda env create -q -n skorch-env -f environment.yml python=${TRAVIS_PYTHON_VERSION}
+  - conda create -y -n skorch-env python=${TRAVIS_PYTHON_VERSION} scikit-learn tqdm
```
ottonemo: Why drop the `environment.yml`? Isn't CI with a fixed environment preferable?

.travis.yml
thomasjpfan: Sometimes we get bugs for specific python versions: https://github.com/dnouri/skorch/issues/333

This change ensures we test for 3.7 as well.

.travis.yml
thomasjpfan: The current version of `environment.yml` file fixes the python version. This meant all test runs used python 3.6. 

This PR was updated to use the `requirements.txt` to install dependencies.

.travis.yml
ottonemo: Sorry, I've overlooked the line installing the travis python version. That makes perfect sense.

.travis.yml
```diff
@@ -21,10 +26,12 @@ install:
...
-  - conda env create -q -n skorch-env -f environment.yml python=${TRAVIS_PYTHON_VERSION}
+  - conda create -y -n skorch-env python=${TRAVIS_PYTHON_VERSION}
   - source activate skorch-env
-  - conda install --file=requirements-dev.txt
+  - pip install tabulate==0.7.7
+  - cat requirements.txt requirements-dev.txt > reqs.txt
+  - conda install --file=reqs.txt
   - python setup.py install
-  - conda install -c pytorch 'pytorch-cpu>=0.4.1'
+  - conda install -c pytorch pytorch-cpu==${PYTORCH_VERSION}
```
ottonemo: Can you move tabulate to the `requirements.txt` instead of installing it here separately?

thomasjpfan: Placing tabulate in `requirements.txt` would be preferable, but tabulate is not in conda. Thus we would not be able to use `conda install --file=reqs.txt` to install all the dependencies at once.

thomasjpfan: There was an issue with this approach:  When pytorch gets installed, it will install the conda version of numpy. Having a pip version and a conda version of numpy installed may cause weird issues down the line.

thomasjpfan:
Since the `setup.py` file includes tabulate, it will automatically be installed when skorch is installed. Thus the line to install tabulate can be removed.

ottonemo:
LGTM

PR #405, opened 2018-12-11
Highlight some features in README

ottonemo:
I propose to highlight some features in the README after the examples.

This is a WIP since the CLI PR is not merged yet.


benjamin-work:
I like this idea. Maybe we could also add a "features" or "highlights" section to the docs (in another PR).

PR #406, opened 2018-12-11
Add support for scipy sparse CSR matrices as input data.

benjamin-work:
Among other use cases, this allows to use transformers that return
sparse matrices in the model pipeline, e.g. sklearn's
`CountVectorizer`. Previously, data needed to be cast to dense arrays
before passing it to skorch.

Note that during batching, data is cast to dense because pytorch's
`DataLoader` cannot deal with sparse matrices. But even if it could, it
would be of little use because torch support for sparse arrays is very
incomplete at this time (e.g. nn.Linear does not work in
backwards). Still this allows to avoid having to cast a potentially
huge input sparse matrix to dense so the feature should be useful.

This PR also updates to_tensor to allow casting a scipy sparse matrix
to a torch sparse COO tensor when accept_sparse=True. Even though
skorch does not use this feature, it seemed to be useful for this
function in general, since with pytorch 1.0, sparse support became
better overall.

### DIFF
--- a/skorch/utils.py
+++ b/skorch/utils.py
@@ -13,6 +13,7 @@ import pathlib
 import warnings
 
 import numpy as np
+from scipy import sparse
 from sklearn.utils import safe_indexing
 import torch
 from torch.nn.utils.rnn import PackedSequence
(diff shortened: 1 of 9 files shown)


skorch/utils.py
```diff
@@ -58,6 +60,9 @@ def to_tensor(X, device):
       tensors will be pushed to cuda tensors before being sent to the
       module.
 
+    accept_sparse : bool (default=False)
+      Whether to accept sparse matrices as input.
```
ottonemo:
```suggestion
      Whether to accept sparse matrices as input and convert them to torch sparse tensors.
```

PR #408, opened 2018-12-12
Attempt to merge setup dependencies

ottonemo:
WIP to merge the requirements.txt and requirements in setup.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,9 @@ install:
 
   - conda create -y -n skorch-env python=${TRAVIS_PYTHON_VERSION}
   - source activate skorch-env
-  - cat requirements.txt requirements-dev.txt > reqs.txt
-  - conda install --file=reqs.txt
-  - pip install .
+  - pip install -r requirements.txt
+  - pip install -r requirements-dev.txt
+  - python setup.py install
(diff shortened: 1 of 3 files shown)


thomasjpfan:
The only issue blocking this is how tabulate is not in conda. Moving scikit-learn and tqdm to `requirements-dev.txt` and running `conda install -file=requirements-dev.txt` would work. Since scikit-learn depends on scipy and numpy, they will be installed. This will satisfy the requirements from setup.py.

Do we need to pin the versions of tqdm and tabulate?

ottonemo:
Can't we do the following:
```
conda create -n skorch-dev
pip install -r requirements.txt
pip install -r requirements-dev.txt
python setup.py install
conda install pytorch -c pytorch
```
I think that solves all the issues?

ottonemo:
Sorry, you also have to include the python version of course:
```
conda create -n skorch-dev
conda install python==3.7
pip install -r requirements.txt
pip install -r requirements-dev.txt
python setup.py install
conda install pytorch -c pytorch
```

thomasjpfan:
When using pip, we should enable the travis cache in the `.travis.ci` file:

```yaml
cache:
  pip: true
```

Issue #412, opened 2018-12-14
conda package cant find requirements.txt


dhirschfeld:
Having a problem over at https://github.com/conda-forge/skorch-feedstock/pull/2
```
Processing $SRC_DIR
  Created temporary directory: /tmp/pip-req-build-op7959fv
  Added file://$SRC_DIR to build tracker '/tmp/pip-req-tracker-3oildf45'
  Running setup.py (path:/tmp/pip-req-build-op7959fv/setup.py) egg_info for package from file://$SRC_DIR
    Running command python setup.py egg_info
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-op7959fv/setup.py", line 9, in <module>
        with open('requirements.txt') as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
```

...and indeed the requirements.txt doesn't seem to be in the pypi tar.gz meaning the `setup.py` can't be run from the sdist.

I *think* `requirements.txt` needs to be in the `MANIFEST.in`?

dhirschfeld:
***xref:*** https://github.com/dnouri/skorch/pull/408

benjamin-work:
That sounds like the problem, do you want to create a PR?

dhirschfeld:
I'm on holiday until the new year so will be unlikely to find the time before then. Happy to do so if no one has beaten me to it though...

BenjaminBossan:
No problem, I'll do this.

ottonemo:
This was fixed in #413. We just released `0.5.0.post0`.

RoyiAvital:
Is there a chance to update the `conda-forge` package?
It seems to be stuck on version [`0.11.0`](https://anaconda.org/conda-forge/skorch/files).

BenjaminBossan:
> Is there a chance to update the `conda-forge` package?

@dhirschfeld Would you mind merging the two open PRs for skorch 0.12.0 aund 0.12.1? You could also give me permissions to the repo if you want me to do it.

dhirschfeld:
I published the 0.12.1 version, but in general I'm stretched too thin to keep up with dependencies not on my own critical path.

As such, your help would be much appreciated @BenjaminBossan! If you'd like to be added as a maintainer just open an issue on the feedstock as described below:
***xref:*** https://conda-forge.org/docs/maintainer/infrastructure.html#conda-forge-admin-please-add-user-username

[...]

PR #413, opened 2018-12-17
Fix/include requirements in manifest

benjamin-work:


ottonemo:
This should resolve #412.

Issue #419, opened 2019-01-09
Shuffle iterator_train by default


EthanRosenthal:
Please correct me if I am wrong, but it looks like `iterator_train` does not shuffle the dataset by default. This came as a surprise to me when I noticed a degradation in model performance when migrating custom PyTorch training code to `skorch`. I have now used `skorch` on a number of projects and never realized that I should be shuffling`iterator_train`, and I imagine that others have had the same issue.

I was wondering what you think about changing the default behavior of `skorch` to shuffle `iterator_train`? I would be happy to submit a PR for this; or, a warning could be displayed if `iterator_train` does not shuffle.

BenjaminBossan:
Thank you for your proposal. The reason why shuffling is not the default is because [`DataLoader`](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader) does not shuffle by default. I agree that often shuffling is the desired behaviour but am not sure if we should change the default. Perhaps a middle ground would be to highlight this behaviour more in the docs and examples, what do you think?

PS: Here is the easiest way to enable shuffling on train data:

https://skorch.readthedocs.io/en/latest/user/FAQ.html#how-do-i-shuffle-my-train-batches

EthanRosenthal:
Fair point - I assumed that was the reasoning behind your decision. I guess the hard thing is competing between PyTorch defaults and sklearn defaults (e.g. [SGDClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html#sklearn.linear_model.SGDClassifier) shuffles by default). Looking through the skorch [tutorials](https://skorch.readthedocs.io/en/latest/user/tutorials.html), it seems that only two tutorials shuffle the data during training, so I wonder how widespread this is.

I understand wanting to keep things as is, though. Highlighting this in the docs would definitely be helpful. Maybe making sure the Quickstart and README both include this parameter with an inline comment explaining why? Also, maybe the examples should be rerun with shuffling (if nothing else, the performance should improve 🙂 )

EthanRosenthal:
Sure thing - I'll try to submit something this week.

ZaydH:
[...]

PR #421, opened 2019-01-11
Highlight shuffling of training data

EthanRosenthal:
Small change in reference to #419


BenjaminBossan:
LGTM, @ottonemo what do you think?

PR #425, opened 2019-01-20
[MRG] Update transform learning to use class in NeutralNet.__init__

thomasjpfan:
Updates tutorial to use a class in `NeutralNet.__init__`.

### DIFF
--- a/notebooks/Transfer_Learning.ipynb
+++ b/notebooks/Transfer_Learning.ipynb
@@ -39,7 +39,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Could you quickly explain why you made that change?

thomasjpfan:
Previously, the module was created first and passed to `NeuralNet`. This PR pushes the use of the stateless class to creating the `NeuralNet` object.

BenjaminBossan:
Okay, so this is to encourage more "skorchy" usage. From my perspective, this is good to go, @ottonemo what do you think?

ottonemo:
LGTM

PR #427, opened 2019-01-28
Better error msg when dataset incompatible

### LINKED ISSUE #422: Facilitate the use of custom Datasets
Custom `Dataset`s are somewhat second class citizens in skorch land. For instance, see this blogpost:

https://nodata.science/no-fit-comparing-high-level-learning-interfaces-for-pytorch.html

I could imagine that others also face problems when they want to transition their existing pytorch code to skorch.

The solution is not straightforward, since we probably cannot change skorch in a way that allows to use any kind of `Dataset`. However, we can probably give more helpful error messages. E.g., when `__getitem__` does not return exactly two values, we could point that out to the user and tell them how to fix the problem.

I personally work mostly with numpy arrays or dicts of arrays, thus I rarely encounter this sort of problem. So if anyone can share their trouble they had with `Dataset`s, that would be helpful.

BenjaminBossan:
Resolves #422 

Note: I successfully ran the `mnist.py` benchmark to check for performance regression.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -20,6 +20,20 @@ from skorch.utils import multi_indexing
 from skorch.utils import to_numpy
 
 
+ERROR_MSG_1_ITEM = (
+    "You are using a non-skorch dataset that returns 1 value. "
+    "Remember that for skorch, Dataset.__getitem__ must return exactly "
+    "2 values, X and y (more info: "
+    "https://skorch.readthedocs.io/en/stable/user/dataset.html).")
+
+
+ERROR_MSG_MORE_THAN_2_ITEMS = (
+    "You are using a non-skorch dataset that returns {} values. "
+    "Remember that for skorch, Dataset.__getitem__ must return exactly "
+    "2 values, X and y (more info: "
+    "https://skorch.readthedocs.io/en/stable/user/dataset.html).")
+
+
 def _apply_to_data(data, func, unpack_dict=False):
     """Apply a function to data, trying to unpack different data
     types.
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thinking about this, would `unpack_batch` be a better name? Although the function would also work in other contexts, here we are dealing with batches. Or is there another name that would fit better here?

skorch/dataset.py
```diff
@@ -75,6 +89,41 @@ def uses_placeholder_y(ds):
...
+    "2 values, X and y (more info: "
+    "https://skorch.readthedocs.io/en/stable/user/dataset.html).")
+
+
+ERROR_MSG_MORE_THAN_2_ITEMS = (
+    "You are using a non-skorch dataset that returns {} values. "
+    "Remember that for skorch, Dataset.__getitem__ must return exactly "
+    "2 values, X and y (more info: "
+    "https://skorch.readthedocs.io/en/stable/user/dataset.html).")
+
```
ottonemo: Why duplicate the messages?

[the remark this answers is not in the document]
BenjaminBossan: The error messages are not exactly the same, so they are not duplicates. Or do you suggest to define the common part separately, or merge the two and format the distinct part?

One of the reasons I chose to define the messages as such was to minimize the overhead due to creating/formatting the string during the `unpack_data` call because it could be called very often. But one could argue that this formatting would only when there is an error, in which case it is only called once. Idk if there are cases where the error would be triggered repeatedly and thus cause overhead.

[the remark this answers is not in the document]
thomasjpfan: Lines 23-34 are the same as lines 92-103.

skorch/dataset.py
```diff
@@ -75,6 +89,41 @@ def uses_placeholder_y(ds):
...
+
+def unpack_data(data):
```
thomasjpfan: Exposing this as an argument in `NeuralNet` may give the flexibility needed to control how data is "unpacked".

BenjaminBossan: I was wondering about that too. Is there a case where you would want to change this specific behavior? My impression is that the assumption is so fundamental that you would never change it anyway, but perhaps I'm missing a case.

thomasjpfan: The use case would be if one uses a pure PyTorch Dataset where its `__getitem__` function returns more than 2 values. A custom `unpack_data` would transform the data to a 2-tuple. 

I think this is really edge case. Advising a user to adjust the `__getitem__` to return a 2-tuple as  in https://github.com/dnouri/skorch/issues/422 would be better.

It is probably better not to expose `unpack_data`.

ottonemo: It is also pretty easy to define your own data iterator (e.g. a generator) which could do the altered unpacking. I'm against exposing `unpack_data` as well.

skorch/tests/test_net.py
```diff
@@ -1593,6 +1593,74 @@ def test_fit_with_dataset_stratified_without_explicit_y_raises(
...
+    def test_fit_with_dataset_one_item_error(
+            self, net_cls, module_cls, dataset_1_item):
+        net = net_cls(module_cls, train_split=None)
+        with pytest.raises(ValueError) as exc:
+            net.fit(dataset_1_item(), None)
+
+        msg = ("You are using a non-skorch dataset that returns 1 value. "
+               "Remember that for skorch, Dataset.__getitem__ must return "
+               "exactly 2 values, X and y (more info: "
+               "https://skorch.readthedocs.io/en/stable/user/dataset.html).")
```
ottonemo: Just noticed this: can't you omit duplicating this message here as well and simply reference the corresponding variable?
```python
msg = skorch.dataset.ERROR_MSG_MORE_THAN_2_ITEMS
```

skorch/dataset.py
BenjaminBossan: Okay, then we leave it as is.

skorch/tests/test_net.py
BenjaminBossan: I'm a little bit torn about this kind of thing, because it makes the test look circular (the message can never be wrong when it is just referenced). But I do also see the advantages, so not sure overall.

PR #431, opened 2019-01-31
Fix unpickling issues with CUDA sub-attributes

ottonemo:
Solves issues with, for example,
* `criterion__weight = <cuda tensor>`
* `optimizer__state`
* `module__some_weird_parameter = <cuda_tensor>`

Before, `criterion` was never considered a CUDA dependent
attribute which is also fixed.


ottonemo:
I think there was also a bug where `net.cuda_dependent_attributes_` was overwritten with the torch byte array of said attributes when unpickling. This is also fixed by introducing the intermediate state variable `__cuda_dependent_attributes__`.

BenjaminBossan:
LGTM. Could you quickly comment on why that part is no longer required:

https://github.com/dnouri/skorch/blob/master/skorch/net.py#L1422

(or was it ever?)

ottonemo:
I've added a compatibility check so that we can direct users to this version of skorch if they have problems loading their old pickle files. In future versions this fallback can be removed.

ottonemo:
@BenjaminBossan done, please review again!

PR #432, opened 2019-02-04
Callbacks are now stored in a CallbacksContainer

fcharras:
This is a quick POC to fix https://github.com/dnouri/skorch/issues/14


fcharras:
See https://github.com/dnouri/skorch/issues/434 for better solutions.

PR #435, opened 2019-02-07
Workaround for misused cuda dependent attributes

ottonemo:
the `cuda_dependent_attributes_` member was misused as storage in
`__getstate__` which results in errors after re-pickling the unpickled
model since `__getstate__` expects strings in
`cuda_dependent_attributes_`.


BenjaminBossan:
Sorry I don't quite get it. In what way is `cuda_dependent_attributes_` being misused as a storage?

ottonemo:
> Sorry I don't quite get it. In what way is `cuda_dependent_attributes_` being misused as a storage?

It *was*, before PR #431, see https://github.com/dnouri/skorch/pull/431/files#diff-5d1f8e13ab8b822b72e9b295cf7301c5L1412. The result of `torch.save` was stored in `state['cuda_dependent_attributes_']`.

BenjaminBossan:
Okay, I misread that.

Is it possible to make a unit test that triggers the bug this PR fixes?

ottonemo:
OK I left this issue open for too long so I had to roll this up from the beginning to reproduce it.
Essentially, there are two bugs here that either prevent loading old pickle files or re-pickling them with 0.5.1:

1. loading old pickle with 0.5.1, then pickling it again breaks since `self.cuda_dependent_attributes_` is mangled by pre-0.5.1 pickling. This is hard to unit test and a temporary issue, so I'll attach a script to reproduce this issue but won't add a unit test
2. there is a broken check in `__setstate_050__` which might fail when the module of the pickled net is not initialized

I'm currently working on the script to reproduce (1) and a fix for (2).

ottonemo:
It *might* be a good idea to test for issue (2) though. I'll see if I can write up a test tomorrow.

ottonemo:
@BenjaminBossan Please review again.

BenjaminBossan:
Nice, thanks for making the effort to reproduce the error with a test. LGTM

Issue #436, opened 2019-02-07
Misleading message if fit triggers re-initialization


BenjaminBossan:
Take the example from the README.md. If you call `net.initialize()` before calling `net.fit(X, y)`, the following message pops up:

```
Re-initializing module because the following parameters were re-set: .
Re-initializing optimizer because the following parameters were re-set: .
```

This isn't helpful at all. The message should be something like:

`The module and optimizer are reset because `fit` was called. To prevent that, use partial_fit`

thomasjpfan:
> `The module and optimizer are reset because `fit` was called. To prevent that, use partial_fit`

This may be confusing in the context of the `Pipeline` example in the readme. That message makes calling `pipe.fit(X, y)` a mistake. For context here is the code outputting the message:

https://github.com/dnouri/skorch/blob/00eaaf10e135c6d69bf3b36d339b5a3a9df3fcc2/skorch/net.py#L466-L474

I propose a simpler message when `kwargs` is empty:

```
Re-initializing module because the module was already initialized.
```

BenjaminBossan:
I agree that proposing to use `partial_fit` may be bad in case of a pipeline (although few people use pipelines *and* pre-initialize the net).

> Re-initializing module because the module was already initialized.

As a user with little skorch experience, I would find this confusing. If the module is already initialized, why re-initialize it?

How about this message instead:

`fit call has triggered re-initialization of the module and optimizer. To prevent that, either use warm_start=True or call partial_fit`

thomasjpfan:
The only issue I have with referencing `fit` is how the message can trigger in when calling other methods, such as call `initialize`.

BenjaminBossan:
The idea was to display the message specifically iff `fit` was called. I haven't investigated yet how feasible that is. If it isn't, then yes, the message would need to be different.

thomasjpfan:
It is possible. We would need to change the `initialize_module` and `initialize` methods to accept a "caller" `keyword. This way the `initialize_*` methods know what function is calling it.

BenjaminBossan:
My hope is to achieve the goal without the need to have a different signature on `initialize`. It is a method that at least a couple of users call directly. But having esoteric parameters like `triggered_by_fit` is somewhat exposing skorch internals. I would need to investigate more to see if that change would be necessary.

thomasjpfan:
I was thinking of adding a `caller=None` option in `initialize`. 

[...]

PR #440, opened 2019-02-14
Improve user facing messages when parameters are re-set

BenjaminBossan:
WIP

This addresses #436 

I'm not yet sure about the implementation (see discussion) but have added some tests to cover some cases that are currently not handled well. Also, I wanted to try the new github draft PRs :)

Generally, I believe that these messages are important because re-initialization could sometimes be unexpected for users and without good messages, they could get tripped.

What I'm not sure about yet is what message to display in the following case:

```python
net = NeuralNet(MyModule, module__hidden_units=123).initialize()
net.set_params(module__dropout=0.2)
# prints:
# Re-initializing module because the following parameters were re-set: dropout, hidden_units.
# Re-initializing optimizer.
```

Obviously, the optimizer is re-set because `dropout` was set but currently the message would claim that `hidden_units` was also responsible. That is technically true but confusing. One solution would be to drop all parameters whose values did not actually change during initialization.

### DIFF
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -1216,7 +1216,48 @@ class TestNeuralNet:
         stdout = capsys.readouterr()[0]
         assert "Re-initializing module!" not in stdout
 
-    def test_with_initialized_module_partial_fit(
+    def test_message_with_initialized_optimizer_reset(
+            self, net_cls, module_cls, capsys):
+        net = net_cls(module_cls, optimizer__momentum=0.5).initialize()
+        net.set_params(optimizer__momentum=0.6)
+
+        stdout = capsys.readouterr()[0]
+        expected = (
+            "Re-initializing module.\n"
+            "Re-initializing optimizer because the following "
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@ottonemo I was working on this PR when I noticed an inconvenience. I was wondering why there was no message when resetting an optimizer parameter with `set_params`. The reason is that it is now treated as a `virtual_params` and not a `special_params`, therefore never reaching this part:

https://github.com/skorch-dev/skorch/blob/6433348930c49882dc3c7c450746726ba11f1924/skorch/net.py#L1350

Consequently, there is no message being displayed. I'm not sure what the right solution is.

In general, this all seems very complicated, perhaps more complicated than it needs be. Even without the mentioned issue, there wouldn't be any easy solution (at least none that I found). I wonder if we can find a way to simplify this part of the code.

BenjaminBossan:
This issue was addressed in another PR, closing for now. We may revisit some of the things addressed here at a later time.

PR #445, opened 2019-03-17
[MRG] Adds Batch Count in History

thomasjpfan:
Fixes https://github.com/dnouri/skorch/issues/418

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -119,8 +119,7 @@ class LRScheduler(Callback):
             try:
                 self.batch_idx_ = sum(net.history[:, 'train_batch_count'])
             except KeyError:
-                self.batch_idx_ = sum(len(b)
-                                      for b in net.history[:, 'batches'])
+                self.batch_idx_ = sum(len(b) for b in net.history[:, 'batches'])
         self.lr_scheduler_ = self._get_scheduler(
             net, self.policy_, **self.kwargs
         )
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -119,7 +119,8 @@ class LRScheduler(Callback):
             try:
                 self.batch_idx_ = sum(net.history[:, 'train_batch_count'])
             except KeyError:
-                self.batch_idx_ = len(net.history[:, 'batches'][0])
+                self.batch_idx_ = sum(len(b)
+                                      for b in net.history[:, 'batches'])
         self.lr_scheduler_ = self._get_scheduler(
             net, self.policy_, **self.kwargs
         )
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -116,7 +116,10 @@ class LRScheduler(Callback):
 
     def on_train_begin(self, net, **kwargs):
         if net.history:
-            self.batch_idx_ = len(net.history[:, 'batches']) - 1
+            try:
+                self.batch_idx_ = sum(net.history[:, 'train_batch_count'])
+            except KeyError:
+                self.batch_idx_ = len(net.history[:, 'batches'][0])
         self.lr_scheduler_ = self._get_scheduler(
             net, self.policy_, **self.kwargs
         )
@@ -138,15 +141,17 @@ class LRScheduler(Callback):
         else:
             self.lr_scheduler_.step(epoch)
 
-    def on_batch_begin(self, net, **kwargs):
+    def on_batch_begin(self, net, training, **kwargs):
         if (
+                training and
                 hasattr(self.lr_scheduler_, 'batch_step') and
                 callable(self.lr_scheduler_.batch_step)
         ):
             self.lr_scheduler_.batch_step(self.batch_idx_)
 
-    def on_batch_end(self, net, **kwargs):
-        self.batch_idx_ += 1
+    def on_batch_end(self, net, training, **kwargs):
+        if training:
+            self.batch_idx_ += 1
 
     def _get_scheduler(self, net, policy, **scheduler_kwargs):
         """Return scheduler, based on indicated policy, with appropriate
--- a/skorch/tests/callbacks/test_lr_scheduler.py
+++ b/skorch/tests/callbacks/test_lr_scheduler.py
@@ -104,18 +104,52 @@ class TestLRCallbacks:
             policy,
             kwargs,
     ):
-        num_examples = 1000
         batch_size = 100
-        max_epochs = 2
+        max_epochs = 1
 
         X, y = classifier_data
+        num_examples = len(X)
+
         lr_policy = LRScheduler(policy, **kwargs)
         net = NeuralNetClassifier(classifier_module(), max_epochs=max_epochs,
                                   batch_size=batch_size, callbacks=[lr_policy])
         net.fit(X, y)
-        expected = (num_examples // batch_size) * max_epochs - 1
+        net.partial_fit(X, y)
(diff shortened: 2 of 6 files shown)


ottonemo:
Why are we not leveraging `new_batch` in `History` for this? It is basically *the* entry point for registering a new batch, we could just count from there.

BenjaminBossan:
I haven't look at the whole PR but wanted to comment on some parts:

> The total number of samples is another beast

We do have `train_batch_size` and `valid_batch_size`, which can be used to sum the total number of samples seen.

> Why are we not leveraging `new_batch` in `History` for this? It is basically _the_ entry point for registering a new batch, we could just count from there.

The proposed change does not look all that elegant and requires to change the signature of `new_batch`. I understand the desire to keep the fit loop lean, but I'm not convinced this is the best approach. Perhaps we can somehow use `net.on_batch_end`, since that already receives `training=False/True`:

https://github.com/skorch-dev/skorch/blob/72a3c8763f03ae4436b4e34a94798a7f7743c8c1/skorch/net.py#L316-L318

 (and we should really fix the line break there!)

ottonemo:
> > The total number of samples is another beast
> 
> We do have `train_batch_size` and `valid_batch_size`, which can be used to sum the total number of samples seen.

Oh well, I completely overlooked that we track the batch size for each batch. So the total sample count could be easily computed via `training_samples = sum(history[:, 'batches', :, 'train_batch_size'])`. Never mind me, then :)

> > Why are we not leveraging `new_batch` in `History` for this? It is basically _the_ entry point for registering a new batch, we could just count from there.
> 
> The proposed change does not look all that elegant and requires to change the signature of `new_batch`. I understand the desire to keep the fit loop lean, but I'm not convinced this is the best approach. Perhaps we can somehow use `net.on_batch_end`, since that already receives `training=False/True`:

If we do it `on_batch_end`, the code would look like something like this, I presume:

```python
key = 'train_batch_count' if training else 'valid_batch_count'
self.history.record(key, self[-1, key] + 1)
```

I think I prefer the fit-loop solution over this one.

> (and we should really fix the line break there!)

+1

thomasjpfan:
Maybe `LRScheduler` should be recording the batch counts, since it is the only one using it.

ottonemo:
If you mean that the LR scheduler should write the batch count to the history I would object since this would simply move the 'bad coupling' argument to the LR scheduler.

`NeuralNet` is a good place to do this since the batch count is a very general thing and `NeuralNet` is a very general class. A specific callback for this would work as well but the case is too small to justify this, I think. Integrating it to LR scheduler only makes sense if it doesn't "leak" to the outside, e.g. by writing to the history (except for what it is supposed to do, namely learning rate scheduling of course) or else we risk having callbacks depend on LRScheduler just for the batch count.

The best solution for me would be to have everything that is necessary to resume training at the same point for the `LRScheduler` callback to be stored locally in the callback. This would work for pickled networks but not for freshly initialized networks that continue training from loading a non-pickle checkpoint. 

In my opinion we should just take the solution as is, think about the batch count being updated during looping instead of at the end and go on.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -104,18 +104,52 @@ def test_lr_callback_batch_steps_correctly(
             policy,
             kwargs,
     ):
-        num_examples = 1000
         batch_size = 100
-        max_epochs = 2
+        max_epochs = 1
```
ottonemo: Wouldn't `max_epochs == 2` be a better test for this since it also covers the case of multiple epochs?

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -104,18 +104,52 @@ def test_lr_callback_batch_steps_correctly(
...
         X, y = classifier_data
+        num_examples = len(X)
+
         lr_policy = LRScheduler(policy, **kwargs)
         net = NeuralNetClassifier(classifier_module(), max_epochs=max_epochs,
                                   batch_size=batch_size, callbacks=[lr_policy])
         net.fit(X, y)
-        expected = (num_examples // batch_size) * max_epochs - 1
+        net.partial_fit(X, y)
+        expected = int(1.6 * num_examples / batch_size) * max_epochs
```
ottonemo: Can you clarify this expression? I find it non-obvious (where does the 1.6 come from?) and having `max_epochs` here with `max_epochs=1` doesn't make sense.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -104,18 +104,52 @@ def test_lr_callback_batch_steps_correctly(
...
+    ])
+    def test_lr_callback_batch_steps_correctly_fallback(
+            self,
+            classifier_module,
+            classifier_data,
+            policy,
+            kwargs,
+    ):
+        batch_size = 100
+        max_epochs = 1
```
ottonemo: Same argument as above.

BenjaminBossan:
@thomasjpfan Are you still on this?

thomasjpfan:
@BenjaminBossan Thanks for the ping! I have updated this PR to address the comments.

skorch/tests/callbacks/test_lr_scheduler.py
thomasjpfan: The original intention was to use `fit` and `partial_fit` to simulate two epochs.

This PR has been updated to used `max_epochs=2` and use one fit`. The `test_lr_callback_batch_steps_correctly_fallback` tests for partial fit and that the fallback works.

skorch/tests/callbacks/test_lr_scheduler.py
thomasjpfan: The PR was updated to break the steps up when calculating the expected batch_idx.

skorch/callbacks/lr_scheduler.py
```diff
@@ -116,7 +116,11 @@ def kwargs(self):
 
     def on_train_begin(self, net, **kwargs):
         if net.history:
-            self.batch_idx_ = len(net.history[:, 'batches']) - 1
+            try:
+                self.batch_idx_ = sum(net.history[:, 'train_batch_count'])
+            except KeyError:
+                self.batch_idx_ = sum(len(b)
```
BenjaminBossan: Very minor nitpick: I don't like the indentation here, this should fit in one line (88 chars is okay).

BenjaminBossan:
Apart from the minor nitpick concerning the line break, this LGTM. @ottonemo what do you think?

skorch/callbacks/lr_scheduler.py
thomasjpfan: I have gotten too use to sklearn's 80 characters limit. :)

This PR has been updated.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -104,18 +104,70 @@ def test_lr_callback_batch_steps_correctly(
...
+        X, y = classifier_data
+        num_examples = len(X)
+
+        lr_policy = LRScheduler(policy, **kwargs)
+        net = NeuralNetClassifier(classifier_module(), max_epochs=max_epochs,
+                                  batch_size=batch_size, callbacks=[lr_policy])
+        net.fit(X, y)
+
+        # Removes batch count information in the last two epochs
+        for i in range(2):
```
githubnemo: shouldn't 2 be `max_epochs` here?

thomasjpfan: Yup! PR updated to reflect this.

githubnemo:
LGTM @BenjaminBossan can you merge this?

PR #446, opened 2019-03-21
Add a wrapper class for datasets to work with sklearn

### LINKED ISSUE #443: How to create a Skorch Dataset from a Torch Dataset for Grid search Cross-Validation using Skorch.GridSearchCV()
> We can extend `TrainValidDataset` to support the full `torch.utils.data.Dataset` interface:
> 
> ```python
> class TrainValidDataset(torch.utils.data.Dataset):
>     def __init__(self, train_ds, valid_ds):
>         self.train_ds = train_ds
>         self.valid_ds = valid_ds
> 
>     def __len__(self):
>         return len(self.train_ds) + len(self.valid_ds)
> 
>     def __getitem__(self, i):
>         if i < len(self.train_ds):
>             return self.train_ds[i]
>         i = i - len(self.train_ds)
>         return self.valid_ds[i]
> ```
Hello @thomasjpfan I attempted applying this procedure for K-fold CV with torch ImageFolder datasets, but ran into a TypeError with the following stack trace:
image

Here is the code sample:
```
data = load_data.load('./poorly_cohessive/ck_dataset')
train_ds, valid_ds = data[0], data[1]

net = NeuralNetClassifier(
    PretrainedModel,
    criterion=nn.CrossEntropyLoss,
    lr=0.001,
    batch_size=32,
    max_epochs=25,
    module__output_features=2,
    optimizer=optim.SGD,
    optimizer__momentum=0.9,
    iterator_train__shuffle=True,
    iterator_train__num_workers=4,
    iterator_valid__shuffle=True,
    iterator_valid__num_workers=4,
    train_split=predefined_split(valid_ds),
    callbacks=[lrscheduler, checkpoint, freezer],
    device='cuda' # comment to train on cpu
)
from sklearn.model_selection import GridSearchCV
params = {
    'lr': [0.01, 0.02],
    'max_epochs': [10, 20]
    #'module__num_units': [10, 20],
}
gs = GridSearchCV(net, params, refit=False, cv=5, scoring='accuracy')
gs.fit(train_ds, y=None)
print(gs.best_score_, gs.best_params_)
```
Did I miss any other information? I am quite new to Skorch. Feedback much appreciated.
I sent you a tweet on this :)

_Originally posted by @Tsakunelson in https://github.com/dnouri/skorch/issues/282#issuecomment-470359481_

BenjaminBossan:
When wrapping a torch dataset with `SliceDatasetX`, it can be passed to sklearn grid search et al.

Fixes #443

### DIFF
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -19,6 +19,35 @@ length of the arrays and not the number of keys, and you get a
 with :class:`.SliceDict`, this works.
 
 
+.. _slicedictx:
+
+SliceDatasetX
+-------------
+
+A :class:`.SliceDatasetX` is a wrapper for
+:class:`torch.utils.data.Dataset`\s that makes them behave a little
+bit like :class:`numpy.ndarray`\s. That way, you can slice your
+dataset with lists and arrays, and you get a ``shape`` attribute.
+These properties are useful because if your data is in a dataset, you
+would normally not be able to use sklearn
+:class:`~sklearn.model_selection.GridSearchCV` and similar things;
+with :class:`.SliceDatasetX`, this works.
+
+Note that even if your dataset already contains the y values, you must
+still provide them separately for sklearn. Most of the time, it should
+be pretty straightforward to extract the y values from your dataset,
+so that this won't be an issue. Something like this should do the job:
+
+.. code:: python
+
+    ds = MyCustomDataset()
+    # assume that y is the second value returned by indexing into ds
+    y_from_ds = np.asarray([ds[i][1] for i in range(len(ds))])
+    # wrap the dataset into the new helper class
+    ds_sliceable = SliceDatasetX(ds)
+    gs.fit(ds_sliceable, y_from_ds)
+
+
 Command line interface helpers
 ------------------------------
 
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -7,6 +7,7 @@ from functools import partial
 import warnings
 
 import numpy as np
+from torch.utils.data.dataloader import default_collate
 
 from skorch.utils import _make_split
 from skorch.utils import _make_optimizer
@@ -129,6 +130,82 @@ class SliceDict(dict):
 
     def __ne__(self, other):
         return not self.__eq__(other)
+
+
+def _unpack_first(x):
+    if isinstance(x, tuple):
+        return x[0]
+    return x
+
+
+class SliceDatasetX:
+    """Helper class that wraps a torch dataset to make it work with
+    sklearn.
+
+    Sometimes, sklearn will touch the input data, e.g. when splitting
+    the data for a grid search. This will fail when the input data is
+    a torch dataset. To prevent this, use this wrapper class for your
+    dataset.
+
+    Note that this class will only return the X value (i.e. the first
+    value returned by indexing the original dataset). Sklearn, and
+    hence skorch, always require 2 values, X and y. Therefore, you
+    still need to provide the y data separately.
+
+    Examples
+    --------
+    >>> X = MyCustomDataset()
+    >>> search = GridSearchCV(net, params, ...)
+    >>> search.fit(X, y)  # raises error
+    >>> ds = SliceDatasetX(X)  # or Xs = SliceDict(**X)
+    >>> search.fit(ds, y)  # works
+
+    Parameters
+    ----------
+    dataset : torch.utils.data.Dataset
+      A valid torch dataset.
+
+    collate_fn : callable (default=torch.utils.data.dataloader.default_collate)
+      A function that merges a list of samples to form a
+      mini-batch. This is typically the same function that torch's
+      DataLoader class uses to collate batches.
+
+    """
+    def __init__(self, dataset, collate_fn=default_collate):
+        self.dataset = dataset
+        self.collate_fn = collate_fn
+
+        self._indices = list(range(len(self.dataset)))
+
+    def __len__(self):
+        return len(self.dataset)
+
+    @property
+    def shape(self):
+        return (len(self),)
+
+    def transform(self, X):
+        return X
+
+    def __getitem__(self, i):
+        if isinstance(i, (int, np.integer)):
+            Xi = _unpack_first(self.dataset[i])
+            return self.transform(Xi)
+
+        if isinstance(i, slice):
+            i = self._indices[i]
+
+        if isinstance(i, np.ndarray):
+            if i.ndim != 1:
+                raise IndexError("SliceDatasetX only supports slicing with 1 "
+                                 "dimensional arrays, got {} dimensions instead."
+                                 "".format(i.ndim))
+            if i.dtype == np.bool:
+                i = np.flatnonzero(i)
+
+        Xi = self.collate_fn([
+            self.transform(_unpack_first(self.dataset[j])) for j in i])
+        return Xi
 
 
 # TODO: remove in 0.5.0
--- a/skorch/tests/test_helper.py
+++ b/skorch/tests/test_helper.py
@@ -254,6 +254,125 @@ class TestSliceDict:
         assert sldict0 != sldict1
 
 
+class TestSliceDatasetX:
+    @pytest.fixture
+    def data(self):
+        X, y = make_classification(100, 20, n_informative=10, random_state=0)
+        return X.astype(np.float32), y
+
+    @pytest.fixture
+    def X(self, data):
+        return data[0]
+
+    @pytest.fixture
+    def y(self, data):
+        return data[1]
+
+    @pytest.fixture
+    def custom_ds(self, data):
+        """Return a custom dataset instance"""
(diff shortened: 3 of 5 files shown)


docs/user/helper.rst
```diff
@@ -19,6 +19,35 @@ length of the arrays and not the number of keys, and you get a
 with :class:`.SliceDict`, this works.
 
 
+.. _slicedictx:
```
taketwo: The class name is `SliceDatasetX`, but the anchor name has _dict_ in it. Is this on purpose?

docs/user/helper.rst
```diff
@@ -19,6 +19,35 @@ length of the arrays and not the number of keys, and you get a
...
+
+SliceDatasetX
+-------------
+
+A :class:`.SliceDatasetX` is a wrapper for
+:class:`torch.utils.data.Dataset`\s that makes them behave a little
```
taketwo: I think in order to avoid convoluted names in the running text, our convention was to use unqualified name and spell PyTorch in front, i.e. in this case:

> is a wrapper for PyTorch :class:`~torch.utils.data.Dataset`s

skorch/helper.py
```diff
@@ -131,6 +132,82 @@ def __ne__(self, other):
...
+    def shape(self):
+        return (len(self),)
+
+    def transform(self, X):
+        return X
+
+    def __getitem__(self, i):
+        if isinstance(i, (int, np.integer)):
+            Xi = _unpack_first(self.dataset[i])
+            return self.transform(Xi)
```
thomasjpfan: Currently, `self.transform` returns `X`, do we intend on letting users to subclass `SliceDatasetX` to provide their own transform?

docs/user/helper.rst
BenjaminBossan: Thanks, this was unintended

docs/user/helper.rst
BenjaminBossan: Will fix it.

skorch/helper.py
BenjaminBossan: That's the intent. I will document it better.

BenjaminBossan:
I made several improvements to the implementation.

* I updated docstrings and docs according to the reviewer comments above.
* I changed the implementation to make it easy to also wrap y in this class (and hence renamed it from `SliceDatasetX` to `SliceDataset`.
* I noticed that with the old implementation, when data is e.g. split into train and valid durng grid search, all data is loaded into memory, which we don't want. Now the implementation is closer to a PyTorch `Subset`. I.e. when a `SliceDataset` is indexed by a list, array, or slice, it creates a `SliceDataset` subset that doesn't actually load any data. Only when it is indexed by an int will it load the data. This avoids the mentioned issue.
Of course, this makes the assumption that when sklearn touches the data, it will never slice using an int. AFAIK, this assumption should hold, but sklearn makes no guarantees about that. If anyone knows a more robust solution, please chime in. What we need is basically a way to know during slicing whether the actual data is needed at this moment or not.

BenjaminBossan:
@ottonemo Could you review this pls?

skorch/helper.py
```diff
@@ -131,6 +132,105 @@ def __ne__(self, other):
...
+    >>> search.fit(X, y)  # raises error
+    >>> ds = SliceDataset(X)
+    >>> search.fit(ds, y)  # works
+
+    Parameters
+    ----------
+    dataset : torch.utils.data.Dataset
+      A valid torch dataset.
+
+    n : int (default=0)
```
thomasjpfan: Is there a better variable name for this?

skorch/helper.py
```diff
@@ -131,6 +132,105 @@ def __ne__(self, other):
...
+      :class:`~torch.utils.data.Subset`.
+
+    """
+    def __init__(self, dataset, n=0, indices=None):
+        self.dataset = dataset
+        self.n = n
+        self.indices = indices
+
+        self.indices_ = (self.indices if self.indices is not None
+                         else np.arange(len(self.dataset)))
```
thomasjpfan: Is the `indices=None` case tested?

skorch/tests/test_helper.py
```diff
@@ -254,6 +254,189 @@ def test_equals_different_keys(self, sldict_cls):
...
+        slice(0, 2),
+        slice(0, None),
+        slice(-1, None),
+        slice(None, None, -1),
+        [0],
+        [55],
+        [-3],
+        [0, 10, 3, -8, 3],
+        np.ones(100, dtype=np.bool),
+        np.array([0, 0, 1, 0] * 25, dtype=np.bool),
```
thomasjpfan: `* 25` and cast to bool?

skorch/helper.py
```diff
@@ -131,6 +132,105 @@ def __ne__(self, other):
...
+        :class:`~torch.utils.data.DataLoader`, the latter will call
+        the dataset for each row separately, which means that the
+        incoming ``data`` is a single rows.
+
+        """
+        return data
+
+    def __getitem__(self, i):
+        if isinstance(i, (int, np.integer)):
+            Xi = self.dataset[self.indices_[i]][self.n]
```
thomasjpfan: Do we want a custom error message when `self.n` out of bounds?

skorch/tests/test_helper.py
```diff
@@ -254,6 +254,189 @@ def test_equals_different_keys(self, sldict_cls):
...
+        ([55], 0),
+        (slice(0, 1), 0),
+        (slice(-1, None), 0),
+        ([55], -1),
+        ([0, 10, 3, -8, 3], 1),
+        (np.ones(100, dtype=np.bool), 5),
+        (np.array([0, 0, 1, 0] * 25, dtype=np.bool), 6),
+    ])
+    def test_slice_twice(self, slds_cls, custom_ds, X, y, sl0, sl1, n):
+        data = y if n else X
```
thomasjpfan: `n==1` is slightly clearer, but this is okay.

skorch/tests/test_helper.py
```diff
@@ -254,6 +254,189 @@ def test_equals_different_keys(self, sldict_cls):
         assert sldict0 != sldict1
 
 
+class TestSliceDataset:
+    @pytest.fixture
```
thomasjpfan:
```suggestion
    @pytest.fixture(scope='class')
```

skorch/helper.py
BenjaminBossan: I have no good idea, any suggestions?

skorch/helper.py
BenjaminBossan: Yes, as it's the default. But I added a test for `indices != None`

BenjaminBossan:
@thomasjpfan I addressed your comments

skorch/tests/test_helper.py
thomasjpfan: Oops. Noticed `[0, 0, 1, 0]` is a list, (Not an numpy array)

BenjaminBossan:
@ottonemo Is this good to go?

skorch/helper.py
githubnemo: How about renaming the parameter to `key` to be in line with builtins like `sorted` and `max`?

BenjaminBossan: The argument doesn't serve the same purpose, though, since `key` in those functions is a callable.

A more similar function would be `itemgetter`, where the argument is called `item`. But `itemgetter` also has `items` for returning a tuple, so now I wonder if we should just use `itemgetter` under the hood to make it possible to return multiple items.

skorch/tests/test_helper.py
```diff
@@ -254,6 +254,209 @@ def test_equals_different_keys(self, sldict_cls):
...
+        return data[1]
+
+    @pytest.fixture
+    def custom_ds(self, data):
+        """Return a custom dataset instance"""
+        from skorch.dataset import Dataset
+        class MyDataset(Dataset):
+            """Simple pytorch dataset that returns 2 values"""
+            def __init__(self, X, y):
+                super().__init__(X, y)
```
githubnemo: Why have the `__init__` implementation at all?

githubnemo:
@BenjaminBossan I'm currently out of office until Friday so I can't merge it but there wasn't anything except for the awkward name of parameter `n` for which I just made an alternative suggestion (but we can keep it if there are objections to my suggestion) and my question regarding `__init__`.

skorch/tests/test_helper.py
BenjaminBossan: True, I will remove this.

BenjaminBossan:
So would `item` be better than `n`?

thomasjpfan:
`index` works too. (I am okay with `item` or `index`)

BenjaminBossan:
My first association with `index` would be the `index` method on strings, lists, tuples, but maybe that's just me. I could see `idx` though, that might be better than `item`.

thomasjpfan:
It is indexing a tuple returned by the `Dataset`. Let's do `idx`?

PR #454, opened 2019-04-10
Increase notebook version to >=5.7.8 in nuclei example

BenjaminBossan:
This is because of a github security note. 5.7.8 seems to be the most recent version of notebook. I haven't actually checked what changed but I assume that the notebook should still work as expected.

### DIFF
--- a/examples/nuclei_image_segmentation/requirements.txt
+++ b/examples/nuclei_image_segmentation/requirements.txt
@@ -3,4 +3,4 @@ torch==0.4.0
 torchvision==0.2.1
 matplotlib==2.2.2
 kaggle==1.4.2
-notebook==5.6.0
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
I wonder if it wouldn't be better to only include the dependencies needed in addition to those that skorch already requires. That way, there would be less redundancy. Of course the drawback is that this requirement.txt doesn't stand on its own anymore.

thomasjpfan:
> only include the dependencies needed in addition to those that skorch already requires.

I am okay with that. The notebook will depend on skorch, so it assume the skorch dependencies are installed.

BenjaminBossan:
I made the discussed changes to the requirements.txt.

githubnemo:
LGTM

PR #455, opened 2019-04-11
Move criterion to compute device automatically

ottonemo:
Previously the user had to make sure that criterion parameters
such as class weights are on the correct computing device.
Now the criterion is moved to the compute device using `.to`
which moves the parameters as well.


ottonemo:
One issue with this is that we would then force `criterion` to be a subclass of `torch.nn.Module` since whatever is passed needs to have the `.to` method. We could live with that or add something like

```python
        self.criterion_ = self.criterion(**criterion_params)
        if isinstance(self.criterion_, torch.nn.Module):
            self.criterion_ = self.criterion_.to(self.device)
```

What do you think?

BenjaminBossan:
Yes, I believe we should allow any kind of `nn.Module`. Furthermore, we might add a note to the docstring of criterion that tells the user that if they wish to have proper device handling on the criterion, it should be an instance of `nn.Module`.

ottonemo:
@BenjaminBossan Fixed, please review again.

Issue #428, opened 2019-01-28
How to extract features in forward method.


junwoony:
I am using `NeuralNetRegressor` to predict output values and I am wondering If there is a way to extract/visualize intermediate features in the forward method (such as `crys_fea` in the forward). It seems like `NeuralNetRegressor` does not support multiple return values in the forward method. Please give me some suggestions to extract those intermediate features.

```python
class ConvNet(nn.Module):

    def __init__(self, ...)
        super(CrystalGraphConvNet, self).__init__()
        self.embedding = nn.Linear(orig_atom_fea_len, atom_fea_len)
        self.convs = nn.ModuleList([ConvLayer()])
        self.conv_to_fc = nn.Linear(atom_fea_len, h_fea_len)
        self.conv_to_fc_softplus = nn.Softplus()
        self.fc_out = nn.Linear(h_fea_len, 1)

    def forward(self, ...):
        atom_fea = self.embedding(atom_fea)
        for conv_func in self.convs:
            atom_fea = conv_func(atom_fea, nbr_fea, nbr_fea_idx)
        crys_fea = self.pooling(atom_fea, crystal_atom_idx)
        crys_fea = self.conv_to_fc(self.conv_to_fc_softplus(crys_fea))
        out = self.fc_out(crys_fea)
        return out
```

```python
from skorch import NeuralNetRegressor 
from model import CrystalGraphConvNet

net = NeuralNetRegressor(
    CrystalGraphConvNet,
    modules...)

net.initialize()
net.fit(SDT_training,target_training)
```

ottonemo:
For a quick inspection you can also use a PyTorch [foward hook][1] which will be called whenever the `forward` method of a specific module is called:

```python
net.initialize()
net.module_.conv_to_fc.register_forward_hook(lambda mod, inp, out: print(out))
net.fit(SDT_training,target_training)
```

[1]: https://pytorch.org/docs/master/nn.html#torch.nn.Module.register_forward_hook

junwoony:
Thanks for your inputs, @BenjaminBossan and @ottonemo.
I tried to set `return out, crys_fea` in the forward, but it gives me an error if I try to run `net.fit(SDT_training, traget_training)`.

The error says that,
```
TypeError                                 Traceback (most recent call last)
<ipython-input-11-368234cd854e> in <module>()
      1 net.initialize()
      2 # net.predict(SDT_training)
----> 3 net.fit(SDT_training,target_training)

[...]

PR #457, opened 2019-04-16
Better document how to return multiple values from forward

BenjaminBossan:
Add an elaborate example to the `Avanced_Usage.ipynb`
notebook. Furthermore, add a short entry in the FAQ that also links to
the notebook.


thomasjpfan:
For reviewers, you can run the notebook by going to: https://colab.research.google.com/github/skorch-dev/skorch/blob/feature/document-multiple-return-values/notebooks/Advanced_Usage.ipynb

BenjaminBossan:
Fixes #428

notebooks/Advanced_Usage.ipynb
```diff
@@ -1144,11 +1150,380 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Since the module's `forward` method returns two values, we have to adjust our objective to do the right thing with those values. If we don't do this, the criterion wouldn't know what to do with the two values and would raise an error.\n",
+    "\n",
+    "One strategy would be to only use the decoded state for the loss and discard the encoded state. For this demonstration, we have a different plan: We would like the encoded state to be sparse. Therefore, we add an L1 loss of the encoded state to the reconstruction loss. This way, the net will try to reconstruct the input as accurately as possible while keeping the encoded state as sparse as possible.\n",
+    "\n",
+    "To implement this, the right method to override is called `get_loss`, which is where `skorch` computes and returns the loss. It gets the prediction (our tuple) and the target as input, as well as some other things that we don't need to bother with. We create a subclass of `NeuralNetRegressor` that overrides said method and implements our idea for the loss."
```
thomasjpfan: "Don't need to be bother with" is a bit aggressive? Maybe:

> It gets the prediction (our tuple) and the target as input, as well as other arguments and keywords that we passthrough.

docs/user/FAQ.rst
```diff
@@ -243,3 +243,25 @@ The ``RandomDataset`` can be passed directly to
...
+can deal with the output or you need to override
+:func:`~skorch.net.NeuralNet.get_loss` and handle the unpacking of the
+tuple.
+
+To inspect all output values, you can use either the
+:func:`~skorch.net.NeuralNet.forward` method (eager) or the
+:func:`~skorch.net.NeuralNet.forward_iter` method (lazy).
+
+For an example of how this works, have a look at this `notebook
+<https://nbviewer.jupyter.org/github/dnouri/skorch/blob/master/notebooks/Advanced_Usage.ipynb#Multiple-return-values-from-forward>`_.
```
thomasjpfan:
```suggestion
<https://nbviewer.jupyter.org/github/skorch-dev/skorch/blob/master/notebooks/Advanced_Usage.ipynb#Multiple-return-values-from-forward>`_.
```

notebooks/Advanced_Usage.ipynb
```diff
@@ -37,7 +37,11 @@
     "  * [Accessing callback parameters](#Accessing-callback-parameters)\n",
```
ottonemo: <p>This cell does not reach the promised accuracy of 70%, maybe it would make sense to write an assertion so that such errors are caught directly?</p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/skorch-dev/skorch/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

ottonemo: <p>Maybe instead of writing 'would raise an error', quote the actual error (makes it searchable).</p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/skorch-dev/skorch/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='84'/>

ottonemo: <p>I would have liked a heatmap of the activations to have a bit of a visual element but this is OK.</p>

 _Reply via <a href='https://app.reviewnb.com/skorch-dev/skorch/pull/457/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='100'/>

BenjaminBossan: The error would be

`AttributeError: 'tuple' object has no attribute 'size'`

It is raised within pytorch code. We could catch that but this seems overly specific. We could think of another strategy to deal with that situation, but that would be out of scope for this PR.

BenjaminBossan: The activations have shape (batch_size, 5), that doesn't really lend itself too well for a visualizations. Also, running the notebook currently doesn't require matplotlib, it's probably not worth for this to add this requirement.

BenjaminBossan: :+1:

BenjaminBossan:
@ottonemo pls review again

notebooks/Advanced_Usage.ipynb
ottonemo: Yeah I just thought that quoting the error in the notebook would make it searchable through google if someone searches for `skorch AttributeError has not attribute size` or something similar.

Issue #467, opened 2019-05-02
PyTorch 1.1 release


BenjaminBossan:
There has been a new PyTorch release ([notes](https://github.com/pytorch/pytorch/releases)) and we should update skorch to keep pace. On my notebook (no GPU), all but 3 tests passed. These one failed:
https://github.com/skorch-dev/skorch/blob/17256a1325ea463003a0e927a09a11065dc0b21a/skorch/tests/test_net.py#L313-L322
This one could probably be removed, since it tests whether we can unpickle an old state (<=0.5).

https://github.com/skorch-dev/skorch/blob/17256a1325ea463003a0e927a09a11065dc0b21a/skorch/tests/callbacks/test_lr_scheduler.py#L22-L27
The test fails for both parametrizations. Since this uses `LRScheduler`, it might have to do with the breaking change introduced [here](https://github.com/pytorch/pytorch/pull/7889).

I will also run the tests on GPU once I have time.

thomasjpfan:
The change in https://github.com/pytorch/pytorch/pull/7889 changes the order of calling `step` in the scheduler. Looking at the [StepLR](https://pytorch.org/docs/stable/optim.html?highlight=steplr#torch.optim.lr_scheduler.StepLR) (and the other schedulers) documentation, it suggest to run `step` at the end of a epoch now.

thomasjpfan:
The fix to `simulate` would most likely break on other versions of PyTorch. We would also need to move the `step` call to `on_epoch_end` which will change behavior for users on PyTorch 1.0.

BenjaminBossan:
No, let's not have code that branches based on PyTorch version. If we made a change to `StepLR` and at the same time increase the minimum PyTorch version to 1.1.0, would that result in the same behavior as before? If so, I would argue that it's not a breaking change.

thomasjpfan:
> If we made a change to StepLR and at the same time increase the minimum PyTorch version to 1.1.0, would that result in the same behavior as before?

Yes, as long as we dump the minimum version, everything should work as before. With this I will make a PR to fix the scheduler part.

On a side note, we can run our tests nightly on TravisCI with the nightly version of PyTorch. This way we can better prepare for a new PyTorch release.

BenjaminBossan:
> Yes, as long as we dump the minimum version, everything should work as before. With this I will make a PR to fix the scheduler part.

Sounds good.

[...]

PR #474, opened 2019-05-15
Remove support for unpickling old files

BenjaminBossan:
With PyTorch 1.1.0, we can no longer unpickle old models that have an
optimizer (i.e. basically all models). Therefore, we may as well drop
the support for unpickling old files now, when moving to PyTorch
1.1.0.


BenjaminBossan:
This addresses some points in #467, but there are additional parts of the code that should be removed with the skorch 0.6.0 release.

BenjaminBossan:
Closing this in favor of #476

PR #488, opened 2019-06-23
More helpful error message when predicting with unfitted net

### LINKED ISSUE #487: Use sklearn's check_is_fitted
We should use sklearn [`check_is_fitted`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html). This leads to better usability.

BenjaminBossan:
Fixes #487 

* Use check_is_fitted similar to what is used in sklearn, except that
  a skorch.exceptions.NotInitializedError is raised.
* Only assumes the presence of 'module_' so that nets remain hackable,
  except where a different attribute is specifically required.
* Re-wrote existing checks to now use check_is_fitted.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -25,6 +25,7 @@ from skorch.history import History
 from skorch.setter import optimizer_setter
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import TeeGenerator
+from skorch.utils import check_is_fitted
 from skorch.utils import duplicate_items
 from skorch.utils import get_map_location
 from skorch.utils import is_dataset
@@ -679,6 +680,7 @@ class NeuralNet:
         like dropout by setting ``training=True``.
 
         """
+        self.check_is_fitted()
         with torch.set_grad_enabled(training):
             self.module_.train(training)
             return self.infer(Xi)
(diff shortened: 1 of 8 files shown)


skorch/utils.py
```diff
@@ -483,6 +484,31 @@ def get_map_location(target_device, fallback_device='cpu'):
     return map_location
 
 
+def check_is_fitted(estimator, attributes, msg=None, all_or_any=all):
```
thomasjpfan: We can also use the sklearn version directly:

```py
from sklearn.utils.validation import check_is_fitted as sk_check_is_fitted

def check_is_fitted(...):
    try:
        sk_check_is_fitted(...)
    except NotFittedError as e:
        raise NotInitializedError(str(e))
```

skorch/net.py
```diff
@@ -852,6 +854,27 @@ def fit(self, X, y=None, **fit_params):
         self.partial_fit(X, y, **fit_params)
         return self
 
+    def check_is_fitted(self, attributes=None, *args, **kwargs):
```
thomasjpfan: Is `check_is_fitted` public to signal that it can be overridden by subclasses?

skorch/utils.py
BenjaminBossan: This doesn't quite work because the error message now says "...is not **initialized** yet...". Or do you believe this is not enough to justify not re-using sklearn's `check_is_fitted`?

skorch/net.py
BenjaminBossan: Yes, the idea was to make it easy for a user to change the attributes that they consider to be necessary for a net to count as fitted.

skorch/utils.py
thomasjpfan: I believe the scikit-learn version has a `msg` argument to change the message.

PR #491, opened 2019-07-06
BUG Adds eps to loss to prevent nans

### LINKED ISSUE #481: torch.log(y_pred) in NeuralNetClassifier.get_loss may cause NaN
When I was testing the `examples/MNIST.ipynb` notebook I stumbled upon an NaN-issue whenever I changed the optimizer to RMSprop or Adam or changed the learning rate from the original value. After using the anomaly detection I saw that the underlying issue are extreme values (close to 1 or 0) in `y_pred` after softmax which resulted in `torch.log(y_pred)` yielding `-inf`. This was already a concern in one of the comments of #136 and now we have a "real-world" example of this.

There is the obvious fix of adding an epsilon value to `y_pred` before calling `torch.log` but this *feels* like justifying a bad decision. PyTorch recommends using `log_softmax` exactly for this reason and we are preventing people from using it by expecting `softmax` output. At the least we should make it easy to do it 'properly' (a flag maybe? `yes_i_want_log_softmax`, not sure about this)

Obvious fix:
```diff
diff --git a/skorch/classifier.py b/skorch/classifier.py
index 95fcad2..e955e03 100644
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -52,8 +52,10 @@ class NeuralNetClassifier(NeuralNet):
             *args,
             criterion=torch.nn.NLLLoss,
             train_split=CVSplit(5, stratified=True),
+            epsilon=1e-8,
             **kwargs
     ):
+        self.epsilon = epsilon
         super(NeuralNetClassifier, self).__init__(
             module,
             *args,
@@ -102,7 +104,7 @@ class NeuralNetClassifier(NeuralNet):
     # pylint: disable=arguments-differ
     def get_loss(self, y_pred, y_true, *args, **kwargs):
         if isinstance(self.criterion_, torch.nn.NLLLoss):
-            y_pred = torch.log(y_pred)
+            y_pred = torch.log(y_pred + self.epsilon)
         return super().get_loss(y_pred, y_true, *args, **kwargs)
 
     # pylint: disable=signature-differs

```

thomasjpfan:
Fixes #481

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Adds an eps to `y_pred` when calculating loss for a `NeuralNetClassifer` (#491)
+
 ### Fixed
 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Adds an eps to `y_pred` when calculating loss for a `NeuralNetClassifer` (#491)
+- Improve numerical stability when using `NLLLoss` in `NeuralNetClassifer` (#491)
 
 ### Fixed
 
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -103,7 +103,8 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
     # pylint: disable=arguments-differ
     def get_loss(self, y_pred, y_true, *args, **kwargs):
         if isinstance(self.criterion_, torch.nn.NLLLoss):
-            y_pred = torch.log(y_pred)
+            eps = torch.finfo(y_pred.dtype).eps
+            y_pred = torch.log(y_pred + eps)
         return super().get_loss(y_pred, y_true, *args, **kwargs)
 
     # pylint: disable=signature-differs
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -98,6 +98,13 @@ class TestNeuralNet:
             assert not (y_out < 0).all()
             assert torch.isclose(torch.ones(len(y_out)), y_out.sum(1)).all()
 
+    # classifier-specific test
+    def test_high_learning_rate(self, net_cls, module_cls, data):
+        # regression test for nan loss with high learning rates issue #481
+        net = net_cls(module_cls, max_epochs=2, lr=2, optimizer=torch.optim.Adam)
+        net.fit(*data)
+        assert np.any(~np.isnan(net.history[:, 'train_loss']))
+
 
 class TestNeuralNetBinaryClassifier:
     @pytest.fixture(scope='module')


BenjaminBossan:
Very nice, I hope this solves the issue once and for all. Could you add an entry to the CHANGES.md?

CHANGES.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Adds an eps to `y_pred` when calculating loss for a `NeuralNetClassifer` (#491)
```
BenjaminBossan: I would prefer a more user-centric description of the change, say:

```suggestion
- Improve numerical stability when using `NLLLoss` in `NeuralNetClassifer` (#491)
```

thomasjpfan: Good point.

BenjaminBossan:
Much appreciated.

skorch/classifier.py
```diff
@@ -103,7 +103,8 @@ def check_data(self, X, y):
     # pylint: disable=arguments-differ
     def get_loss(self, y_pred, y_true, *args, **kwargs):
         if isinstance(self.criterion_, torch.nn.NLLLoss):
-            y_pred = torch.log(y_pred)
+            eps = torch.finfo(y_pred.dtype).eps
+            y_pred = torch.log(y_pred + eps)
```
Atcold: This is **not** a smart operation.
Running `log(exp( · ))` is not computationally stable.

thomasjpfan: Since this is guarded by the `isinstance(self.criterion_, torch.nn.NLLLoss)`, this only happens with this negative log loss. You can use a different `criterion_` such as `CrossEntropyLoss`, and skorch would not take this log.

Fundamentally, this comes from a constraint where the output of `module(X)` are probabilities: https://github.com/skorch-dev/skorch/issues/481#issuecomment-497454767

> The idea was that at that time, it was standard (keras, nolearn) that if you define a classifier, your output nonlinearity should be softmax.

BenjaminBossan: @Atcold did you have any practical issues resulting from this?

Atcold: I was following these tutorials ([1](https://skorch.readthedocs.io/en/stable/user/quickstart.html#training-a-model) and [2](https://github.com/skorch-dev/skorch/blob/master/notebooks/MNIST.ipynb)), which I've assigned to my mentee, but then was surprised to see the soft(arg)max inside the model. So, I've looked up the source, and saw that nasty thing 😬 
I have not run any code, but it's bad practice. Especially using it in an example / tutorial.

**If** you really like a soft(arg)max inside a model (since this appears to be *standard*, for some), then I'd rather get the logit, by indexing the `L - 1` layer, and use that in a `CrossEntropy` criterion.

BenjaminBossan: We discussed the issue at length (see the linked comment above). There are other design decisions that we need to take into account (how to go from `module_.forward` to `net.predict_proba` for instance), so you cannot judge this decision in isolation. And if we changed it now, it would be a backwards incompatible change, breaking a lot of existing code. Moreover, I have yet to encounter a problem with that -- if this really causes numerical instability, it almost certainly means that you've done something wrong with your model in the first place. Finally, if you change the loss, you'll not run into the problem, so it's easy to circumvent if it should really cause trouble. For all those reasons, my point of view is to leave it as it is.

Atcold: Okay, added my 2¢ there.

Issue #489, opened 2019-06-24
Possible disconnect between X and y values


ottonemo:
There is a possible disconnect between ordering of `y` and `X` when using a skorch 
net as transformer in a sklearn pipeline while enabling shuffling on the valid 
iterator which is a very easy mistake to make (by coincidence).

To explain the basic issue, consider the following setup:

```python
class TransformingNet(NeuralNet):
    def transform(self, X):
        return forward(X)

net = TransformingNet(module, iterator_valid__shuffle=True)
knn = KNearestNeighbor(n_neighbor=3)

pipe = Pipeline(net, knn)
pipe.fit(X_train, y_train)
pipe.predict(X_valid, y_valid)
```

The ordering of the embedded `X_valid` by `TransformingNet` will not correlate with
the ordering of `y_valid` since the data gets shuffled by `iterator_valid`.
Normally there is no need to shuffle the validation data which is why I wouldn't
think of this as a big issue. However this behavior might not be as obvious
all the time. Consider the following scenario:

```python
class MyFancySampler(Sampler):
    """Basically torch.utils.data.sampler.RandomSampler"""
    def __init__(self, dataset):
        self.dataset = dataset
    def __iter__(self):
        return (i for i in torch.randperm(len(self.dataset))

class MyFancyLoader(DataLoader):
    def __init__(self, ds, **kw):
        kw['sampler'] = MyFancySampler(ds)
        super().__init__(ds, **kw)

    def __iter__(self):
        # ... <intricate iteration code> ...

net = TransformingNet(
          module, 
          iterator_train=MyFancyLoader,
          iterator_valid=MyFancyLoader,
)
```

This code hides the sorting disconnect pretty well and I spent a couple of hours
huntin for a bug that was pretty similar to this.

I think that this (among other things) warrants a separate `iterator_eval` which
is completely disconnected from the training process.

BenjaminBossan:
First of all, this is not restricted to using the net as a transformer, even for the classical use, this could come and bite you. Furthermore, this is an issue that can easily happen with vanilla sklearn or other such frameworks.

I wonder if a separate `iterator_eval` helps. What would prevent the user from using `MyFancyLoader` on `iterator_eval` that wouldn't prevent the user from using it on `iterator_valid`? If they have this special loader, chances are that predicting wouldn't work correctly without it.

[...]

PR #493, opened 2019-07-15
(WIP) Notebook that shows how to use triplet loss

BenjaminBossan:
After some private discussions with @ottonemo and the discussion in #489, I attempted to implement a net with triplet loss using skorch. The result is shown in the notebook below. It is a first draft, intended for discussion.

At the moment, triplets are mined naively, i.e. for each sample, find a random sample of the same class and a random sample of another class. However, this could be used as a base to implement more sophisticated methods like online mining, though others who are more familiar with the topic probably need to comment.

As a first instinct, I wanted to implement the triplet mining on the `Sampler` level of pytorch. However, this quickly turned out not to work well. The main problem is that although we can pass samplers to the `DataLoader` in skorch, pytorch expects the sampler to be instantiated. This requires the sampler to have a reference to the data. However, at the time we initialize the net, we may not have a reference to the data yet (e.g. because the data is not split into train/valid yet). Using samplers thus didn't work.

The most practical approach then seemed to be within the `Dataset`, which is what is shown in the notebook. Currently, the dataset will return a triplet of input data and a triplet of dummy target data. The module then has to deal with the triplet.

Another thing that I wanted to achieve initially was to just stack each triplet before passing it to the module and then to unstack it in `get_loss`. The main advantage of that approach is that it wouldn't need any special treatment within the `Module` itself. However, this would result in the `Dataset` class returning three samples at once, which can lead to all kinds of problems. Also, it would effectively triple the batch size, resulting in unexpectedly high memory needs. Therefore, I didn't take this route at the end and had to modify the module to be able to work with triplets.

If anyone has a better suggestion, please comment. Overall, apart from the logic of finding the positive/negative sample itself, the code didn't seem very complicated.


BenjaminBossan:
ping @yv

BenjaminBossan:
[Colab link](https://colab.research.google.com/github/skorch-dev/skorch/blob/feature/triplet-loss-example/notebooks/triplet_loss.ipynb)

ottonemo:
I haven't looked into it in detail but one thing that stood out to me was:

> As a first instinct, I wanted to implement the triplet mining on the Sampler level of pytorch. However, this quickly turned out not to work well. The main problem is that although we can pass samplers to the DataLoader in skorch, pytorch expects the sampler to be instantiated. This requires the sampler to have a reference to the data. However, at the time we initialize the net, we may not have a reference to the data yet (e.g. because the data is not split into train/valid yet). Using samplers thus didn't work.

What I often do in this case is to create my own `DataLoader` instance (or define a data loader generator), for example:

```python
def dataloader(ds, **kwargs):
    sampler = MySampler(ds)
    dl = DataLoader(ds, sampler=sampler, **kwargs)
    for x, y in dl: yield x, y
```

Would that work for you?

BenjaminBossan:
@ottonemo Thanks for the hint, this does indeed work. I adjusted the notebook to now use the custom `DataLoader` approach that simply stacks the triplets. That way, there is no need to modify the module in any way, which I find more satisfying.

Overall, the amount of custom code, apart from the sampling logic itself, is really quite small. The next step would be to figure out how to perform things like online triplet mining and to maybe check that the performance impact if this approach is not too large.

yv:
One thing I did differently than here was to have a custom Dataset subclass that kept track of the classes and their example indices so that you could get the moral equivalent of find_positive and find_negative in O(1) time. Will share it here, it's slightly awkward to have a subclass just to get "another example with the same/a different class" but then again it's useful for many use cases (triples, pairs, whatever) and it's hard to get to O(1) without it.

I find the `TripletSampler` that performs two different things depending on whether y is filled or not very awkward - and it's easy to think of an example where you'd `fit_transform` on one dataset and `transform` another while (accidentally or deliberately) also passing a `y`, or a case where a triple is broken in half due to the batch size not being a multiple of three. This ties in with the previous discussion of having three different DataLoaders/Samplers for train/eval/transform time, if you can have separate samplers for train/eval and for transform it may be possible to get this in a more transparent fashion.

yv:
Here's the ClassIndexedDataset code: https://gist.github.com/yv/829c6d13ac509ac330e9dc6272e65ea2

BenjaminBossan:
@yv Thanks for sharing your code, my implementation was just a quick and dirty trial and is obviously not the most efficient. I will look at your implemenatation in the following days

> I find the TripletSampler that performs two different things depending on whether y is filled or not very awkward

This is true.

>  and it's easy to think of an example where you'd `fit_transform` on one dataset and `transform` another while (accidentally or deliberately) also passing a `y`

I don't believe this can happen, sklearn won't pass the `y` to the transform step.

> if you can have separate samplers for train/eval and for transform it may be possible to get this in a more transparent fashion

One thing that could make this difficult is that the decision which route to take is often based on `training=True/False`, adding a third option is thus difficult. If you have an idea, please share it.

yv:
This is make_offsets: 
```python
def make_offsets(y_sorted):
    """
    given a sorted list, returns a list of starting (and implicitly ending) offsets
    """
    idx = 0
    offsets = []
    for i, y in enumerate(y_sorted):
        while y >= len(offsets):
            offsets.append(i)
    offsets.append(len(y_sorted))
    return offsets
```

BenjaminBossan:
I played a little bit with your code, though I couldn't quite figure out how you actually use it. Do you use a special `DataLoader` as well?

Regarding the performance, after a little bit of tweaking, my code (only the data loading) runs 50% slower than yours. On the plus side, my code doesn't require to sort the data and thus supports shuffling. Also, I hope that it should not be too hard to add features like online triplet mining through the use of `sample_weight`. If those things are not required, your code is better, though.

yv:
My version does use a custom sampler and a custom DataLoader, both of which are used only at train/eval time and not at transformation time, and can be used to generate fresh triplets in every epoch. I don't see a point yet of having a per-example sample weight, is this used somewhere in the literature?

BenjaminBossan:
> I don't see a point yet of having a per-example sample weight, is this used somewhere in the literature?

The idea was that this can be used to perform more sophisticated sampling, e.g. downsampling samples that are too easy or enforcing a more uniform distribution. Do you use any kind of more sophisticated sampling?

BenjaminBossan:
When `find_positive` and `find_negative` are called, the sample `i` is known, so for each sample, there could be a sample-specific `sample_weight`, as you suggested. The difficult part is how to determine this weight within the sampler without making things too complicated.

PierreGtch:
Hi all,
I don't know if this is still active but I'm currently using skorch with a triplet loss and I ended up putting the triplet selection in the criterion. So I have something like this : 

```python
from skorch import NeuralNet
from torch import nn

class TripletMiner(nn.Module):
    def forward(self, y):
        # compute all triplets of indices from the elements in the batch
        return idx_anchor, idx_pos, idx_neg

class TripletMiningLoss(nn.Module):
    def __init__(self):
        self.miner = TripletMiner()
        self.loss_fn = nn.TripletMarginLoss()
    def forward(self, X_emb, y):
        idx_anchor, idx_pos, idx_neg = self.miner(y)
        return self.loss_fn(X_emb[idx_anchor], X_emb[idx_pos], X_emb[idx_neg])

net = NeuralNet(
    MLPModule,
    criterion=TripletMiningLoss,
)
```
It keeps things pretty simple, only the criterion has to be custom. 
And it's similar to the training of FaceNet, the computation of the  embeddings (heavy part) is only done `batch_size` times but a lot of triplets can be formed within a batch, even more if the batches are well sampled. So it's not required but we can also add a custom `BatchSampler` that only samples batches from a subset of the classes (and change the classes sampled at each new batch) so that the number of positive matches inside each batch is increased.

BenjaminBossan:
@PierreGtch Thanks a lot for providing this snippet.

Do you have a full example lying around somewhere? If so, it could be a practical addition to the collection of notebooks in skorch. If you have an example of a custom sampler, that would also be greatly appreciated.

PierreGtch:
I filled the colab notebook with my code so that it can be compared:

https://gist.github.com/PierreGtch/8e5d81cf6e3ecf574f3ad6cc7a2aa5a1

a few comments:
- The `nan` in `valid_loss` are comming from `iterator_valid__drop_last=False` that allows the generation of batches containing only one class (so no valid triplets inside) but it can't be `True` for prediction
- here, the `SubsetBatchSampler` is not very relevant because there is only 5 classes but we use it for the demonstration
- with all the triplets we present to the loss, our model is overfitting very easily (see `train_loss` vs `valid_loss`)

BenjaminBossan:
Thanks, I will take a closer look at this soon. Perhaps we can add a link to our [list of notebooks](https://github.com/skorch-dev/skorch/blob/master/notebooks/README.md), that might be the easiest way to point users to it.

BenjaminBossan:
@PierreGtch The notebook looks good to me.

I have a question regarding the `TripletMiner` class: Is it necessary for it to be an `nn.Module` or could it be a function instead?

PierreGtch:
> I have a question regarding the `TripletMiner` class: Is it necessary for it to be an `nn.Module` or could it be a function instead?
Yes it could be a function.  This triplet miner returns all possible triplets so there is no parameters.
But someone could want a triplet miner that selects, for example, only "hard" triples, so there would be a threshold involved.

BenjaminBossan:
Okay, thanks for explaining.

So maybe we can add the link to the gist to the README in the notebook folder I mentioned above. That should afford more visibility than this issue.

BenjaminBossan:
This notebook is no longer up-to-date, please refer to the one provided in the [comment above](https://github.com/skorch-dev/skorch/pull/493#issuecomment-841640307).

PR #495, opened 2019-07-21
Remove some linting errors that accumulated over time

BenjaminBossan:
No functional code changes except for removing `_make_optimizer`, which is not used in skorch code anymore.


thomasjpfan:
Do we want the CI to fail when `pylint` fails? It would make it harder for new contributors.

BenjaminBossan:
> Do we want the CI to fail when `pylint` fails? It would make it harder for new contributors.

No, I wouldn't enforce this, there are too many false-positives and adding exceptions everywhere would be a bit too much. However, when new contributions are submitted, we should be more strict with checking compliance, so that we don't need to "clean up" regularly.

PR #503, opened 2019-08-03
Check init arguments for missing dunder notation

### LINKED ISSUE #500: Catching wrong parameters passed to NeuralNet
Right now, NeuralNet et al. are very forgiving in accepting parameters. Part of that has to do with the dynamic nature of skorch but some typos can probably be caught. E.g. in #498, there was the issue that a user passed `iterator_train_num_workers=4` (used single instead of double underscore), which we should be able to identify.

BenjaminBossan:
Fixes #500 

This catches wrong keyword arguments such as `optimizerlr` or `iterator_train_shuffle` and suggests the correct name.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- More careful check for wrong parameter names being passed to NeuralNet
+- More careful check for wrong parameter names being passed to NeuralNet (#500)
 
 ### Changed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- More careful check for wrong parameter names being passed to NeuralNet
+
 ### Changed
 
 - Improve numerical stability when using `NLLLoss` in `NeuralNetClassifer` (#491)
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1326,7 +1326,7 @@ class NeuralNet:
                     if not key.startswith(prefix + '__'):
                         missing_dunder_kwargs.append((prefix, key))
                     break
-            else:  # if no break
+            else:  # no break means key didn't match a prefix
                 unexpected_kwargs.append(key)
 
         msgs = []
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -229,7 +229,7 @@ class NeuralNet:
         initialized = kwargs.pop('initialized_', False)
         virtual_params = kwargs.pop('virtual_params_', dict())
 
-        kwargs = self.check_kwargs(kwargs)
+        kwargs = self._check_kwargs(kwargs)
         vars(self).update(kwargs)
 
         self.history = history
@@ -1272,7 +1272,7 @@ class NeuralNet:
         params.update(params_cb)
         return params
 
-    def check_kwargs(self, kwargs):
+    def _check_kwargs(self, kwargs):
         """Check argument names passed at initialization.
 
         Raises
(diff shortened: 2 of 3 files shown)


skorch/net.py
```diff
@@ -229,20 +229,7 @@ def __init__(
...
-            if any(key.startswith(p) for p in self.prefixes_):
-                continue
-            unexpected_kwargs.append(key)
-        if unexpected_kwargs:
-            msg = ("__init__() got unexpected argument(s) {}. "
-                   "Either you made a typo, or you added new arguments "
-                   "in a subclass; if that is the case, the subclass "
-                   "should deal with the new arguments explicitely.")
-            raise TypeError(msg.format(', '.join(sorted(unexpected_kwargs))))
+        kwargs = self.check_kwargs(kwargs)
```
thomasjpfan: Private for now? I do not think users would want to override this.

skorch/net.py
```diff
@@ -1285,6 +1272,55 @@ def get_params(self, deep=True, **kwargs):
         params.update(params_cb)
         return params
 
+    def check_kwargs(self, kwargs):
+        """Check argument names passed at initialization.
```
thomasjpfan: For our future selves, maybe add a small example on how `iterator_train_shuffle` will raise a `TypeError` that will suggest `iterator_train__shuffle`.

skorch/net.py
```diff
@@ -1285,6 +1272,55 @@ def get_params(self, deep=True, **kwargs):
...
+        missing_dunder_kwargs = []
+        for key in kwargs:
+            if key.endswith('_'):
+                continue
+            for prefix in self.prefixes_:
+                if key.startswith(prefix):
+                    if not key.startswith(prefix + '__'):
+                        missing_dunder_kwargs.append((prefix, key))
+                    break
+            else:  # if no break
```
thomasjpfan: Got to love this very obvious python feature.

skorch/tests/test_net.py
```diff
@@ -178,6 +178,48 @@ def test_net_init_two_unknown_argument(self, net_cls, module_cls):
                     "should deal with the new arguments explicitely.")
         assert e.value.args[0] == expected
 
+    def test_net_init_missing_dunder_in_prefix_argument(
```
thomasjpfan: We can `pytest.mark.parametrize` and test for some other prefixes like `optimizer_momentum`:

```py
@pytest.mark.parametrize("key, value", 
        [("iterator_train_shuffle", True), ("optimizer_momentum", 0.95)]
```

BenjaminBossan:
@thomasjpfan Thx for the feedback, I worked that in.

As an aside: I just noticed that this test now passes even though it should xfail:

https://github.com/skorch-dev/skorch/blob/f94466e272f6f325898359fecb9a7c004354af7f/skorch/tests/test_net.py#L1155-L1167

Not sure what change caused this.

thomasjpfan:
The `test_get_params_with_uninit_callbacks` test passes on master as well.

CHANGES.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- More careful check for wrong parameter names being passed to NeuralNet
```
thomasjpfan: Reference this PR or issue. (I have found reference the PR is more useful, since it usually also links to the issue)

BenjaminBossan: True. Added it.

BenjaminBossan:
> The `test_get_params_with_uninit_callbacks` test passes on master as well.

Yes, now it's hard to find out what change caused this except for bisecting. We may want to add `pytest.mark.xfail(strict=True)` in the future.

skorch/net.py
ottonemo:
```suggestion
            else:  # no break means key didn't match a prefix
```

Issue #501, opened 2019-07-29
Suggestion: Make DataFrames more useful


BenjaminBossan:
At the moment, passing a pandas DataFrame to `net.fit` seems like it would rarely be useful. It is treated similarly to a dictionary because each column is treated as key and the corresponding value as a 1-dim array, but dictionaries can also hold more dimensional data. I suspect that nobody uses DataFrames with skorch.

I would propose not to change this, but to add a useful transformer (inspired by [fastai's TabularDataset](https://docs.fast.ai/tabular.data.html#TabularDataset) that converts a DataFrame into a dictionary. This transformer could be used in a pipeline as a preprocessing step. It would do the following:

* float dtypes: concatenate to a single array with a specific key (e.g. `X`)
* categorical dtypes: leave each as separate values, with the key corresponding to the column name
* integer dtypes: either treat as cardinal (like float) or as categorical, depending on a parameter
* other dtypes: not sure, maybe raise an error

Using this transformer, the module could pass the cardinal features to a dense layer and the categorical features to an embedding layer (or one-hot encode).

The code could look somewhat like this:

```python
df = pd.DataFrame({
    'rooms': [2, 4, 3, ...],
    'area': [103.5, 42., 234.5, ...],
    'state': ['CA', 'FL', 'NY', ...],
})
df['state'] = df['state'].astype('category')

class MyModule(nn.Module):
    ...
    def forward(self, X, state):
        # X has shape (n, 2) and contains rooms, area
        # state has shape (n,) and contains ints
        X_state = self.embed(state)
        X = torch.cat((X, X_state), axis=0)
        ...

model = Pipeline([
    ('preproc', DataFrameTransformer(treat_int_as_categorical=False)),
    ('net', net),
])
```

My suggestion would be to add this transformer to `helper.py` and to not change any of the existing behavior. The transformer could also have a method that helps with building the right module by giving information about the correct keys and sizes.

Do you think this would be useful and fit the scope of skorch?

thomasjpfan:
I think this is pretty useful. For their [example with rossmann](https://github.com/fastai/course-v3/blob/master/nbs/dl1/lesson6-rossmann.ipynb), they heavy use `Embedding` for their categories.

On a ML level, it would be interesting to see how it compares to one-hot encoding.

To simplify the API, I do not think we need `treat_int_as_categorical`. If a user wants to have categories, they should convert the dtype of that column to a pandas category.

BenjaminBossan:
[...]

PR #507, opened 2019-08-08
Add DataFrameTransformer

BenjaminBossan:
Addresses #501


skorch/helper.py
```diff
@@ -256,3 +261,268 @@ def predefined_split(dataset):
...
+                "DataFrame contains a column named 'X', which clashes "
+                "with the name chosen for cardinal features; consider "
+                "renaming that column.")
+
+        wrong_dtypes = []
+
+        for col, dtype in zip(df, df.dtypes):
+            if isinstance(dtype, self.pd.core.dtypes.dtypes.CategoricalDtype):
+                continue
+            if dtype in self.int_dtypes_:
```
thomasjpfan: Does `np.issubdtype(dtype, np.integer)` work?

skorch/helper.py
```diff
@@ -256,3 +261,268 @@ def predefined_split(dataset):
...
+                continue
+            if dtype in self.float_dtypes_:
```
thomasjpfan: Does `np.issubtype(dtype, np.floating)` work?

skorch/helper.py
```diff
@@ -256,3 +261,268 @@ def predefined_split(dataset):
...
+      The dtype to cast the cardinal values to. If None, don't change
+      them.
+
+    int_dtype : numpy dtype or None (default=np.int64)
+      The dtype to cast the categorical values to. If None, don't
+      change them. If you do this, it can happen that the categorical
+      values will have different dtypes, reflecting the number of
+      unique categories.
+
+    signature_ : namedtuple
```
thomasjpfan: A named tuple defined inside a class seems a little harder to discover. Would at tuple or a dictionary be enough?

BenjaminBossan:
I changed the code to use `np.issubdtype` (learned something new). Moreover, I dropped the namedtuple but instead of using a tuple, I used a dict. The reason for that is the better repr, otherwise a lonely number is not very easy to interpret.

@thomasjpfan @ottonemo pls review again

BenjaminBossan:
@thomasjpfan @ottonemo any things to add to this?

skorch/helper.py
```diff
@@ -256,3 +261,250 @@ def predefined_split(dataset):
...
+    >>> # cast to category dtype to later learn embeddings
+    >>> df['col_cats'] = df['col_cats'].astype('category')
+    >>> y = np.asarray([0, 1, 0] * 4)
+
+    >>> class MyModule(nn.Module):
+    ...     def __init__(self):
+    ...         super().__init__()
+    ...         self.reset_params()
+
+    >>>     def reset_params(self):
```
thomasjpfan: Should this be directly in `__init__`?

skorch/helper.py
```diff
@@ -256,3 +261,250 @@ def predefined_split(dataset):
...
+        if 'X' in df:
+            raise ValueError(
+                "DataFrame contains a column named 'X', which clashes "
+                "with the name chosen for cardinal features; consider "
+                "renaming that column.")
+
+        wrong_dtypes = []
+
+        for col, dtype in zip(df, df.dtypes):
+            if isinstance(dtype, self.pd.core.dtypes.dtypes.CategoricalDtype):
```
thomasjpfan: Does `pd.api.types.CategoricalDtype` work?

skorch/helper.py
```diff
@@ -256,3 +261,250 @@ def predefined_split(dataset):
...
+          respective column names as keys.
+
+        """
+        self._check_dtypes(df)
+
+        X_dict = {}
+        Xf = []  # floats
+
+        for col, dtype in zip(df, df.dtypes):
+            values = df[col]
```
thomasjpfan: Nit: Having `values.values` later is a bit harder to follow. Maybe `X_col=df[col`?

skorch/helper.py
BenjaminBossan: It's just my personal preference, that way it's possible to reset the module without creating a new instance.

skorch/helper.py
BenjaminBossan: Not a nit :+1:

BenjaminBossan:
@ottonemo is this ready to be merged?

skorch/tests/test_helper.py
```diff
@@ -474,3 +482,230 @@ def test_pickle(self, predefined_split, data):
...
+    @pytest.fixture
+    def df(self):
+        """DataFrame containing float, int, category types"""
+        import pandas as pd
+
+        df = pd.DataFrame({
+            'col_floats': [0.1, 0.2, 0.3],
+            'col_ints': [11, 11, 10],
+            'col_cats': ['a', 'b', 'a'],
+        })
```
thomasjpfan: I am a little surprised some of the test pass without this in python 3.5:

```suggestion
        }, columns=['col_floats', 'col_ints', 'col_cats'])
```

skorch/helper.py
```diff
@@ -256,3 +261,250 @@ def predefined_split(dataset):
...
+
+            if isinstance(dtype, self.pd.api.types.CategoricalDtype):
+                x = X_col.cat.codes.values
+                if self.int_dtype is not None:
+                    x = x.astype(self.int_dtype)
+                X_dict[col] = x
+                continue
+
+            if (
+                    dtype in (np.dtype('int32'), np.dtype('int64'))
```
thomasjpfan:
```suggestion
                    np.issubdtype(dtype, np.integer):
```

skorch/tests/test_helper.py
BenjaminBossan: Why indicate the column names a second time?

ottonemo:
I will merge this after the tests complete since this seems to be a version we can work with / build up on. What I would like to see in the future is a way to deal with NULL-values for categorical columns (introduce a separate category for NULL) since this is a common enough problem.

thomasjpfan:
Since this PR uses the `cat.codes`, the `np.nan` values will be encoded as `-1`:

```py
import pandas as pd
df = pd.DataFrame({
    'col': pd.Categorical(['a', 'b', np.nan, 'b'])})

df['col'].cat.codes.values
# array([ 0,  1, -1,  1], dtype=int8)
```

PR #508, opened 2019-08-08
Update environment.yml

BenjaminBossan:
For some reason, the old one didn't work for me anymore. I got the following errors:

```bash
$ conda --version
conda 4.7.10
$ conda env create
ResolvePackageNotFound: 
  - setuptools==27.2.0=py36_0
  - xz==5.2.2=1
  - zlib==1.2.8=3
  - readline==6.2=2
  - sqlite==3.13.0=0
  - openssl==1.0.2l=0
  - tqdm=4.14.0
  - tk==8.5.18=0
  - python==3.6.2=0
```

Could someone please confirm this?

If this is true, I wonder why this happens. Shouldn't conda allow reproducible builds? Were all those packages removed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 dist: xenial
 python:
-  - "3.7"
+  - "3.7.3"
   - "3.6"
   - "3.5"
 env:
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
   - numpy-base=1.16.4
   - openssl=1.1.1c
   - pip=10.0.1
-  - python=3.6.9
+  - python=3.7.4
   - readline=7.0
   - scikit-learn=0.21.2
   - scipy=1.3.0
--- a/environment.yml
+++ b/environment.yml
@@ -1,20 +1,38 @@
 name: skorch
 channels:
-- defaults
+  - defaults
 dependencies:
-- mkl
-- openssl=1.0.2l=0
-- pip=10.0.1
-- python=3.6.2=0
-- readline=6.2=2
-- scikit-learn>=0.19.0
-- scipy>=1.1.0
-- setuptools=27.2.0=py36_0
-- sqlite=3.13.0=0
-- tk=8.5.18=0
-- wheel=0.32.2
-- xz=5.2.2=1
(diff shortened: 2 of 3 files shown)


thomasjpfan:
Yea [conda removed some packages from their default/free channel](https://www.anaconda.com/why-we-removed-the-free-channel-in-conda-4-7/), if we want to have access to the old packages we need to run:

```bash
conda config --set restore_free_channel true
```

BenjaminBossan:
Well in that case we should probably update the `environment.yml`. I removed the third part of the package versions (`foo==bar==BAZ`) because I found that this prevents builds from being reproducible. I'm not sure if they are needed.

Can you verify that the updated `environment.yml` works?

thomasjpfan:
On OSX, I get this:

```
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - libgcc-ng=9.1.0
  - libstdcxx-ng=9.1.0
  - libgfortran-ng=7.3.0
```

Removing them worked.

BenjaminBossan:
> On OSX, I get this:

Do you mean with the new or the old environment.yml?

thomasjpfan:
New one in this PR

BenjaminBossan:
I updated the env by removing the offending lines. I also ticked up the default Python version to 3.7.

Additionally, I changed the unit tests to be compatible with the latest pytorch release (1.2).

BenjaminBossan:
I had to change the Python 3.7 version to 3.7.3 from 3.7.4 (default) because numpy couldn't be installed otherwise. Locally, I could replicate this by just creating an env with Python 3.7.4 and calling `pip install numpy`. Different recent numpy versions all had the same effect. Oddly, I couldn't find a solution for this, even though this should affect many people.

BenjaminBossan:
> With the adjustments for PyTorch 1.2, lets update the PYTORCH_VERSION in travis as well?

I would prefer to separate this and then immediately make new release.

BenjaminBossan:
@ottonemo What's your take?

ottonemo:
LGTM. I'm for merging this right now and making a separate PR for PyTorch 1.2 related changes and doing a release after that.

PR #512, opened 2019-08-18
Tensorboard logging

### LINKED ISSUE #510: Proposal: Tensorboard callback
PyTorch seems to take the tensorboard direction in favor of visdom. Therefore, I propose to add a tensorboard callback. Below are the code and a screenshot from running it with the [MNIST](https://nbviewer.jupyter.org/github/dnouri/skorch/blob/master/notebooks/MNIST-torchvision.ipynb) notebook.

```python
class TensorBoard(Callback):
    def __init__(
            self,
            writer,
            add_graph=True,
            close_after_train=True,
    ):
        self.writer = writer
        self.add_graph = add_graph
        self.close_after_train = close_after_train

    def get_default_log_keys(self):
        return [
            ('train_loss', 'Loss/train'),
            ('valid_loss', 'Loss/valid'),
            ('valid_acc', 'Loss/valid_acc'),
            ('dur', 'duration'),
        ]

    def initialize(self):
        self.first_run_ = True
        return self

    def on_epoch_begin(self, net, dataset_train, **kwargs):
        if self.first_run_:
            self.first_run_ = False
            if self.add_graph:
                self.writer.add_graph(net.module_, dataset_train[0][0].to(net.device))

    def add_scalar_maybe(self, history, key, global_step=None, name=None):
        hist = history[-1]
        val = hist.get(key)
        if val is None:
            return

        name = name if name is not None else key
        global_step = global_step if global_step is not None else hist['epoch']
        self.writer.add_scalar(tag=name, scalar_value=val, global_step=global_step)

        history = net.history
        epoch = history[-1, 'epoch']
        for key, name in self.get_default_log_keys():
            self.add_scalar_maybe(history, key, name=name, global_step=epoch)

        if self.close_after_train:
            self.writer.close()
```

tensorboard_proposal

tensorboard_graph

The callback should be easy to extend. If a user wants to add a specific log, they can override one of the `on_*` methods, call `super()` on it, then add their own code.

Getting the module's graph wasn't straightforward, I'm not sure if the proposed method works in all cases.

Before doing a proper PR, I wanted to get the opinion of others, especially those with more experience in using tensorboard.

BenjaminBossan:
Fixes #510 

This includes the changes as proposed by @taketwo, namely to automatically log everything except keys that were opted out, the same as `PrintLog`.

I would like to update one of the existing notebooks to showcase the use. My candidate would be [MNIST-torchvision](https://nbviewer.jupyter.org/github/dnouri/skorch/blob/master/notebooks/MNIST-torchvision.ipynb), or do you think this would be too much for this notebook, @ottonemo? Of course, it's difficult to show the tensorboard inside the Jupyter notebook, I would probably add a screenshot.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -17,6 +17,33 @@ from skorch.callbacks import Callback
 __all__ = ['EpochTimer', 'PrintLog', 'ProgressBar']
 
 
+def filter_log_keys(keys, keys_ignored=None):
+    """Filter out keys that are generally to be ignored.
+
+    This is used by several callbacks to filter out keys from history
+    that should not be logged.
+
+    Parameters
+    ----------
+    keys : iterable of str
+      All keys.
+
+    keys_ignored : iterable of str or None (default=None)
+      If not None, collection of extra keys to be ignored.
+
+    """
+    keys_ignored = keys_ignored or ()
+    for key in keys:
+        if not (
+                key == 'epoch' or
+                (key in keys_ignored) or
+                key.endswith('_best') or
+                key.endswith('_batch_count') or
+                key.startswith('event_')
+        ):
+            yield key
+
+
 class EpochTimer(Callback):
     """Measures the duration of each epoch and writes it to the
     history with the name ``dur``.
@@ -62,7 +89,8 @@ class PrintLog(Callback):
     ----------
     keys_ignored : str or list of str (default=None)
       Key or list of keys that should not be part of the printed
-      table. Note that keys ending on '_best' are also ignored.
+      table. Note that keys starting with 'even_' or ending on '_best'
+      are ignored by default.
 
     sink : callable (default=print)
       The target that the output string is sent to. By default, the
@@ -92,8 +120,6 @@ class PrintLog(Callback):
             floatfmt='.4f',
             stralign='right',
     ):
-        if isinstance(keys_ignored, str):
-            keys_ignored = [keys_ignored]
         self.keys_ignored = keys_ignored
         self.sink = sink
         self.tablefmt = tablefmt
@@ -102,7 +128,11 @@ class PrintLog(Callback):
 
     def initialize(self):
         self.first_iteration_ = True
-        self.keys_ignored_ = set(self.keys_ignored or [])
+
+        keys_ignored = self.keys_ignored
+        if isinstance(keys_ignored, str):
+            keys_ignored = [keys_ignored]
+        self.keys_ignored_ = set(keys_ignored or [])
         self.keys_ignored_.add('batches')
         return self
 
@@ -140,24 +170,25 @@ class PrintLog(Callback):
           * all remaining keys are sorted alphabetically.
         """
         sorted_keys = []
+
+        # make sure 'epoch' comes first
         if ('epoch' in keys) and ('epoch' not in self.keys_ignored_):
             sorted_keys.append('epoch')
 
-        for key in sorted(keys):
-            if not (
-                    (key in ('epoch', 'dur')) or
-                    (key in self.keys_ignored_) or
-                    key.endswith('_best') or
-                    key.endswith('_batch_count') or
-                    key.startswith('event_')
-            ):
+        # ignore keys like *_best or event_*
+        for key in filter_log_keys(sorted(keys), keys_ignored=self.keys_ignored_):
+            if key != 'dur':
                 sorted_keys.append(key)
 
+        # add event_* keys
         for key in sorted(keys):
             if key.startswith('event_') and (key not in self.keys_ignored_):
                 sorted_keys.append(key)
+
+        # make sure 'dur' comes last
         if ('dur' in keys) and ('dur' not in self.keys_ignored_):
             sorted_keys.append('dur')
+
         return sorted_keys
 
     def _yield_keys_formatted(self, row):
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -14,7 +14,7 @@ from skorch.dataset import get_len
 from skorch.callbacks import Callback
 
 
-__all__ = ['EpochTimer', 'PrintLog', 'ProgressBar']
+__all__ = ['EpochTimer', 'PrintLog', 'ProgressBar', 'TensorBoard']
 
 
 def filter_log_keys(keys, keys_ignored=None):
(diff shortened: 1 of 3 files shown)


skorch/callbacks/logging.py
```diff
@@ -62,7 +89,8 @@ class PrintLog(Callback):
     ----------
     keys_ignored : str or list of str (default=None)
       Key or list of keys that should not be part of the printed
-      table. Note that keys ending on '_best' are also ignored.
+      table. Note that keys starting with 'even_' or ending on '_best'
```
taketwo:
```suggestion
      table. Note that keys starting with 'event_' or ending on '_best'
```

Note that this description is incomplete. The new filtering function also ignores `_batch_count` keys (line 41), and `PrintLog` adds `batches` key to the ignored list  (line 136).

BenjaminBossan:
@taketwo Tensorboard also ignores batches, they are added [here](https://github.com/skorch-dev/skorch/pull/512/files#diff-9e40ac6735bc2adfcba1a7d7153fcba5R447).

When it comes to the completeness, I didn't add every single detail, only those that I thought were most relevant to the users. But if you want to, I can change this.

Regarding duration, if there is an easy way to show time per epoch in tensorboard, I can remove the duration. Otherwise I think it's better to log too much than too little.

taketwo:
> @taketwo Tensorboard also ignores batches, they are added here.

Sorry, overlooked this. So both callbacks that use the `filter_log_keys()` utility function ignore the `batches` key. Then the question is: why is it not ignored "globally" in `filter_log_keys()` (like e.g. `epoch`)?

> When it comes to the completeness, I didn't add every single detail, only those that I thought were most relevant to the users. But if you want to, I can change this.

You are right, it's probably not necessary to list every single detail. But for me the sentence

> Note that keys starting with 'event_' or ending on '_best' are ignored by default.

reads like it's an exhaustive list of default ignores, which it is not. So perhaps it would be helpful to note that there are more default ignores. Also from the docstring (without checking code) it's not clear what will happen if user supplies `keys_ignored` argument. Will it replace the defaults or will they be merged?

> Regarding duration, if there is an easy way to show time per epoch in tensorboard, I can remove the duration. Otherwise I think it's better to log too much than too little.

No, I don't think it is possible to have a time per epoch plot, so the user will need to hover data points and compute time differences herself. Agree, too much is better than too little.

BenjaminBossan:
I added `TensorBoard` to the torchvision notebook, including an example that shows how to subclass. Also clarified the docstrings, as suggested by @taketwo

skorch/callbacks/logging.py
```diff
@@ -321,3 +353,163 @@ def on_epoch_begin(self, net, dataset_train=None, dataset_valid=None, **kwargs):
...
+    >>> # Example to log the bias parameter as a histogram
+    >>> def extract_bias(module):
+    ...     return module.hidden.bias
+
+    >>> class MyTensorBoard(TensorBoard):
+    ...     def on_epoch_end(self, net, **kwargs):
+    ...         bias = extract_bias(net.module_)
+    ...         epoch = net.history[-1, 'epoch']
+    ...         self.writer.add_histogram('bias', bias, global_step=epoch)
+    ...         super().on_epoch_end(net, **kwargs)  # call super last
```
thomasjpfan: This does not have to be last?

BenjaminBossan: True, it doesn't have to. The reason why I say this is that sometimes, calling `super` first can lead to subtle errors. E.g. during `on_batch_end`, `first_batch_` is set to false. If a user were to override this and call `super` first, then implement something that depends on `first_batch_` being true, it would never trigger.

ottonemo:
I'm not done testing but the first run of the notebook with PyTorch 1.2 failed at

    net.fit(mnist_train, y=y_train);

with

    Expected object of backend CUDA but got backend CPU for argument #4 'mat1'
    Error occurs, No graph saved

You can find the whole traceback [here](https://github.com/skorch-dev/skorch/files/3537678/tb.log).

I haven't changed anything in the notebook and `DEVICE` is set to `'cuda'`.

BenjaminBossan:
@ottonemo Not sure why you get this error and I don't, but you surely uncovered a more fundamental problem. Right now, `TensorBoard` just passes the module and X to `SummaryWriter.add_graph`, which does the calling. However, this circumvents the logic currently contained in `net.infer`, for instance the device handling.

The good news is that we could move that specific logic to `TensorBoard.add_graph`. The bad news is that we cannot move the logic of how `forward` is called to `TensorBoard`, since this is controlled on the pytorch side. This is a problem when X is a dict, because then we want to call `forward(**X_dict)`, but pytorch will just call `forward(*X_dict)` (meaning we only receive the keys).

I don't see an elegant solution to this. Does anyone of you have an idea? I opened a topic on the pytorch forums to ask for help there as well: https://discuss.pytorch.org/t/use-module-with-kwargs-in-summarywriter-add-graph/54263

BenjaminBossan:
I added a test that shows the discussed failing behavior with a dict input.

ottonemo:
> @ottonemo Not sure why you get this error and I don't, but you surely uncovered a more fundamental problem. Right now, `TensorBoard` just passes the module and X to `SummaryWriter.add_graph`, which does the calling. However, this circumvents the logic currently contained in `net.infer`, for instance the device handling.
> 
> The good news is that we could move that specific logic to `TensorBoard.add_graph`. The bad news is that we cannot move the logic of how `forward` is called to `TensorBoard`, since this is controlled on the pytorch side. This is a problem when X is a dict, because then we want to call `forward(**X_dict)`, but pytorch will just call `forward(*X_dict)` (meaning we only receive the keys).
> 
> I don't see an elegant solution to this. Does anyone of you have an idea? I opened a topic on the pytorch forums to ask for help there as well: https://discuss.pytorch.org/t/use-module-with-kwargs-in-summarywriter-add-graph/54263

Ideally we would have a single point of control for how the data is converted and passed to the model.
Currently, this is `net.infer` and, if possible, it would be nice to have the graph generated the same way.

I tried passing `net.infer` directly to `torch.jit.trace` which then fails with:

    RuntimeError: Cannot insert a Tensor that requires grad as a constant. 
    Consider making it a parameter or input, or detaching the gradient

There is some background information in https://github.com/pytorch/pytorch/issues/20101, basically it seems that the trace is not able to gather the module's parameters this way.

BenjaminBossan:
> Tensorboard logging is already pretty useful with only the metrics. We can have this PR include just the metrics, and have another PR for `add_graph`?

That's a good idea, I removed the graph functionality for now. I left the tests there (skipping them for now) and also show in the notebook how to add the graph yourself.

Please review again.

BenjaminBossan:
> there's explicitly no experiment management, running net.fit(mnist_tain, y_train) more than once will write into the same log dir

> we should definitely think about making it as easy as possible to manage experiments and possibly have the increase of experiment runs handled by skorch.

Since we leave the creation of the `SummaryWriter` to the user, I would consider this the user's job as well. Otherwise, we would need to create it ourselves, with all the trouble and dependencies that come with it. Maybe there is an easy way to achieve what you suggested, but I'm not aware of it.

skorch/callbacks/logging.py
```diff
@@ -321,3 +353,143 @@ def on_epoch_begin(self, net, dataset_train=None, dataset_valid=None, **kwargs):
...
+          Global step value to record.
+
+        """
+        hist = history[-1]
+        val = hist.get(key)
+        if val is None:
+            return
+
+        global_step = global_step if global_step is not None else hist['epoch']
+        try:
```
thomasjpfan:
```py
from contextlib import suppress

with suppress(NotImplementedError):
    self.writer...
```

skorch/tests/conftest.py
```diff
@@ -132,3 +132,11 @@ def data():
     pandas_installed = True
 except ImportError:
     pass
+
+
+tensorboard_installed = False
+try:
+    import tensorboard
```
thomasjpfan: Just to keep pylint happy in the future:

```
# pylint: disable=unused-import
import tensorboard
```

erap129:
I understand this feature has been added, Is there any usage example? Thanks

BenjaminBossan:
@erap129 please look here: https://github.com/skorch-dev/skorch/blob/f7e74afb5d69ac340aa962fb4a7c023c493bf39a/notebooks/MNIST-torchvision.ipynb

Issue #506, opened 2019-08-08
Recommended way to implement gradient accumulation


fabiocapsouza:
Hi,

I am new to Skorch and I am implementing BERT fine-tuning using Skorch.
One of the features that is missing is gradient accumulation, where `loss.backward()` is called on every batch but the optimizer is called only after N consecutive batches.
I believe I have to override `train_step_single` to remove `zero_grad()` before every step and `train_step` to perform `optimizer_.step()` conditionally. 
Also, the Net has to keep track of the number of batches during training and be able to keep a state that can be accessed inside `train_step`.
Where is the recommended place to save this kind of state?

Thanks,

BenjaminBossan:
I didn't know about gradient accumulation, so I don't have any code for you. That being said, the number of training batches can be inferred from the history: `len(self.history[-1, 'batches'])` (this assumes that training steps are performed before validation steps, which is what normally happens).

Regarding running `loss.backward()` several time before calling the optimizer: That I think is not so trivial. Apart from overriding `train_step`, you might even need to override `fit_loop`. The reason is that we assume there that the optimizer is called for each batch.

I don't know enough about the topic to give you a solution to this. But if you tinker a bit and find a working solution, please post it here. Maybe we can then point you towards improving it.

BenjaminBossan:
@fabiocapsouza Did you make progress on this?

fabiocapsouza:
@BenjaminBossan, thanks for your first reply and sorry for the delay.
After your comment, I decided to postpone my migration to Skorch for this task, since it reinforced my feelings that it would require a lot of work. I would not have the time to test all the needed changes that could easily introduce bugs.

Thanks again for your quick response, it really helped me.

BenjaminBossan:
> since it reinforced my feelings that it would require a lot of work.

I think it will be worth exploring if we can refactor skorch to make this easier. Until then, you are probably better off using your current solution. If you want to try with skorch again at some point and maybe help improving it, feel free to open the issue again or starting a new one.

PR #516, opened 2019-08-25
Small changes to net that make grad accumulation easier

BenjaminBossan:
In #506, there was a discussion about how to implement gradient accumulation with skorch. Unfortunately, it seemed to me that it is not easily possible. However, I believe that a small number of tweaks are enough to make it possible:

* move `self.optimizer_.zero_grad()` from `train_step_single` to `train_step` (might break some code but only of very advanced users)
* since `step` can now be None for train in `fit_loop`, add a conditional to only record things if step is not None (ugly but not back breaking)

I hope that this actually does what I read about gradient accumulation, since I have never used it myself. Maybe @fabiocapsouza can comment on that.

Overall, if this should really be enough, I think the changes would be worth implementing, given that popular architectures like BERT use it. However, I wouldn't elevate it to "first class" in skorch (yet), i.e. make it possible to control via a parameter. Instead, I would probably document how to do something similar to the test (possibly in the FAQ).

PS: Not quite sure what to do with `train_batch_count`? Only increment it each time the optimizer is called or for each batch? I guess the latter is more fitting?

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -282,3 +282,46 @@ To inspect all output values, you can use either the
 
 For an example of how this works, have a look at this `notebook
 <https://nbviewer.jupyter.org/github/skorch-dev/skorch/blob/master/notebooks/Advanced_Usage.ipynb#Multiple-return-values-from-forward>`_.
+
+How can I perform gradient accumulation with skorch?
+----------------------------------------------------
+
+There is no direct option to turn on gradient accumulation (at least
+for now). However, with a few modifications, you can implement
+gradient accumulation yourself:
+
+
+.. code:: python
+
+    ACC_STEPS = 2  # number of steps to accumulate before updating weights
+
+    class GradAccNet(net_cls):
+        """Net that accumulates gradients"""
+        def __init__(self, *args, acc_steps=ACC_STEPS, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.acc_steps = acc_steps
+
+        def get_loss(self, *args, **kwargs):
+            loss = super().get_loss(*args, **kwargs)
+            return loss / self.acc_steps  # normalize loss
+
+        def train_step(self, Xi, yi, **fit_params):
+            """Perform gradient accumulation
+
+            Only optimize every nth batch.
+
+            """
+            # note that n_train_batches starts at 1 for each epoch
+            n_train_batches = len(self.history[-1, 'batches'])
+            step = self.train_step_single(Xi, yi, **fit_params)
+
+            if n_train_batches % self.acc_steps == 0:
+                self.optimizer_.step()
+                self.optimizer_.zero_grad()
+            return step
+
+This is not a complete recipe. For example, if you optimize every 2nd
+step, and the number of training batches is uneven, you should make
+sure that there is an optimization step after the last batch of each
+epoch. However, this example can serve as a starting point to
+implement your own version gradient accumulation.
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -295,7 +295,7 @@ gradient accumulation yourself:
 
     ACC_STEPS = 2  # number of steps to accumulate before updating weights
 
-    class GradAccNet(net_cls):
+    class GradAccNet(NeuralNetClassifier):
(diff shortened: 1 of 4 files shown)


fabiocapsouza:
Hi Benjamin,
I am quite new to Skorch, so please correct me if I get something wrong.
I agree that moving `self.optimizer_.zero_grad()` from `train_step_single` is necessary to allow accumulation of gradients of many backward steps.
However, it looks like `GradAccNet.train_step` is not performing `loss.backward()` on all training forward steps, since `self.train_step_single` will only be called every 2 batches (when `self.optimizer.step(step_fn)` is called). Is there a way to test this? Maybe attaching a callback and asserting that `on_grad_computed` was called `b_total` times would work?
It would be necessary to perform `loss.backward()` on all training batches, but calling `optimizer.step()` conditionally. I believe it requires calling `step_fn()` directly, but I am not sure if it would break things.

It is also necessary to divide the loss by the number of accumulated steps to correct the effective batch size. That is, if we want an effective batch size of 8 using an instantaneous batch size of 2 (4 steps of grad. accumulation), the returned loss per step will be the mean loss considering only 2 samples instead of 8, so we have to divide it by 4. I think this correction can be performed by a Callback that divides the gradients directly inside `on_grad_computed` every 4 batches, before `optimizer.step` is called.

BenjaminBossan:
@fabiocapsouza Thanks for your feedback. Indeed my implementation was not quite right. Could you please have a look if it is now correct?

To implement the change, I had to sacrifice calling `optimizer.step(step_fn)` with the closure, not sure if the two are mutually exclusive.

> Is there a way to test this? Maybe attaching a callback and asserting that `on_grad_computed` was called `b_total` times would work?

This wouldn't be a direct test thus not really helpful. I think with the current code it's clear enough that `loss.backward()` is called for each batch.

> It is also necessary to divide the loss by the number of accumulated steps to correct the effective batch size.

I left this out at first since this PR is not about actually implementing gradient accumulation, just about showing implementing it is feasible. The normalization part shouldn't be the problem. Still I have changed the test to divide the loss, is that correct?
(What's also missing is that right now, if there is an uneven number of train batches, there should be an optimizer step for the last batch during the same epoch.)

BenjaminBossan:
>     1. Dividing the loss before `backward()` inside `train_step_single` or `get_loss` on every batch.

I changed the code to reflect this.

To make this PR ready, I will probably add code similar to the test to the FAQ.

docs/user/FAQ.rst
```diff
@@ -282,3 +282,46 @@ To inspect all output values, you can use either the
...
+There is no direct option to turn on gradient accumulation (at least
+for now). However, with a few modifications, you can implement
+gradient accumulation yourself:
+
+
+.. code:: python
+
+    ACC_STEPS = 2  # number of steps to accumulate before updating weights
+
+    class GradAccNet(net_cls):
```
thomasjpfan: Should we detail what `net_cls` is? Could this be `NeuralNetClassifer` to be slightly more concrete?

skorch/net.py
```diff
@@ -730,10 +730,12 @@ def fit_loop(self, X, y=None, epochs=None, **fit_params):
                 yi_res = yi if not y_train_is_ph else None
                 self.notify('on_batch_begin', X=Xi, y=yi_res, training=True)
                 step = self.train_step(Xi, yi, **fit_params)
+                train_batch_count += 1
+                if not step:
```
thomasjpfan: Does `train_step` return a falsy value for this to trigger?

BenjaminBossan: What do you mean? That we should check for `if step is not None`?

thomasjpfan: The examples in this PR that overwrite `train_step` we return something.

BenjaminBossan: Yes, you are right. This was a remnant of my earlier attempt. Fixed now.

BenjaminBossan:
> Regarding elevating to a parameter, this seems fairly popular to do:

I wouldn't go ahead just yet to implement this. As you said, maybe it goes out of fashion soon. Also, in contrast to at least fastai, skorch was never meant to deliver everything out of the box. Instead, it should be "hackable" enough that users can implement almost everything they want without too much hassle. I think this is achieved here.

PR #523, opened 2019-09-13
Tiny bit of more info in the CyclicLR deprecation warning 😅

dnouri:
The warning was pretty ambiguous.  First, you may think that you
should use torch.optim.lr_scheduler.CyclicLR as the net's optimizer,
which will fail.  Second, you may think that you can use
torch.optim.lr_scheduler.CyclicLR as a callback, which will obviously
fail, too.  I'm hoping this will save someone some time.  :-)


BenjaminBossan:
Thanks for the addition, it was indeed ambiguous.

PR #525, opened 2019-09-13
DOC Docs on how to use learning rate schedulers

dnouri:
More hints for confused people like me. ;-)

### DIFF
--- a/docs/user/callbacks.rst
+++ b/docs/user/callbacks.rst
@@ -230,3 +230,45 @@ starting with ``f_``:
 
 Please refer to :ref:`saving and loading` for more information about
 restoring your network from a checkpoint.
+
+
+Learning rate schedulers
+------------------------
+
+The :class:`.LRScheduler` callback allows the use of the various
+learning rate schedulers defined in :mod:`torch.optim.lr_scheduler`,
+such as :class:`~torch.optim.lr_scheduler.ReduceLROnPlateau`, which
+allows dynamic learning rate reducing based on a given value to
+monitor, or :class:`~torch.optim.lr_scheduler.CyclicLR`, which cycles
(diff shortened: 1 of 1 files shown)


docs/user/callbacks.rst
```diff
@@ -230,3 +230,34 @@ starting with ``f_``:
 
 Please refer to :ref:`saving and loading` for more information about
 restoring your network from a checkpoint.
+
+
+Learning rate schedulers
+------------------------
+
+The :class:`.LRScheduler` callback allows the use of the various
+learning rate scheduler defined in :mod:`torch.optim.lr_scheduler`,
```
BenjaminBossan:
```suggestion
learning rate schedulers defined in :mod:`torch.optim.lr_scheduler`,
```

docs/user/callbacks.rst
```diff
@@ -230,3 +230,34 @@ starting with ``f_``:
...
+such as :class:`~torch.optim.lr_scheduler.ReduceLROnPlateau`, which
+allows dynamic learning rate reducing based on a given value to
+monitor, or :class:`~torch.optim.lr_scheduler.CyclicLR`, which cycles
+the learning rate between two boundaries with a constant frequency.
+
+As with other callbacks, you can use `set_params`, and thus search
```
BenjaminBossan: You talk about grid search here but the code example just shows the normal use. Therefore, I would start this paragraph with the last sentence, "Here's a network ...", followed by the code example, then put the part "As with other callbacks..." below the code example.

dnouri: I followed your suggestion and in addition added a grid search code example below.

BenjaminBossan:
Thanks, especially for adding the grid search example :)

PR #526, opened 2019-09-13
DOC Update parallelism example to work with latest sklearn and dask

dnouri:
I present to you: "the new way".

The old way now fails in a very awkward way and I had to find out the hard way.  `joblib` defines a thread global to carry the current active "parallel backend".  Now, imagine what happens when you set the backend using `sklearn.externals.joblib.parallel_backend`, but at runtime scikit-learn will happily use the unchanged backend defined in `joblib`.  I guess yay for vendoring packages?


BenjaminBossan:
Thank you very much for this update.

Issue #521, opened 2019-09-07
Expectation when cloning a net


BenjaminBossan:
At the moment, when we use `sklearn.base.clone` on a fitted net, we receive a fitted net. In sklearn, however, we get back an unfitted estimator. The main reason is that out `net.get_params()` also returns learned attributes (ending on `_`, like `module_`).

For sklearn compliance, we might want to change this behavior. However, I wonder if this can have some unintended side-effects are might break backwards-compatibility for too many people (say, making it so that old pickle files no longer work?).

BenjaminBossan:
One possibility I could imagine: We don't change the behavior of `get_params` yet, only prepare everything for the eventual change. Then, after the next release, we change `get_params` to be more in line with what sklearn does.

Unfortunately, it's not really possible to slap a deprecation warning onto `get_params`. I would be afraid that it would trigger too many false positives. Instead, we have to make clear in the release text that the change will happen and what users need to do to prepare for it, then hope they read it. I believe we should encourage using `copy.deepcopy` for the purpose instead of `clone`.

Overall, most people shouldn't be affected by the change, but I can imagine it leads to some surprising bugs, as many are not even aware when `get_params` is actually used under the hood (e.g. for grid search).

thomasjpfan:
I agree with being compatible with the sklearn api of `clone`.

> making it so that old pickle files no longer work?

I do not think pickling will be affected by this change. Pickling serializes all the attribute regardless of how `get_params` work.

> as many are not even aware when get_params is actually used under the hood (e.g. for grid search).

Since grid search calls `fit`, even with the fitted net it will restarted from the beginning. If a user sets `warm_start=True` or passes in a module instance, I would say the grid search will give false results, since the training is continuing where the last parameter has left off.

BenjaminBossan:
> Since grid search calls `fit`, even with the fitted net it will restarted from the beginning. If a user sets `warm_start=True` or passes in a module instance, I would say the grid search will give false results, since the training is continuing where the last parameter has left off.

That would be my biggest worry with the current situation, especially since it's not easy to notice.

> I do not think pickling will be affected by this change. Pickling serializes all the attribute regardless of how `get_params` work.

Pickling will probably not be a problem but still some code could break because of the change.

PR #527, opened 2019-09-15
Preparation for more correct get params change

BenjaminBossan:
Introduce changes and tests in preparation for new get_params

The new behavior of get_params will be to not returned any "learned"
attributes such as "module_".

This PR implements the new behavior but doesn't switch to it yet to
give users time to adjust their code. This is a breaking change but it
is necessary since it is the "correct" behavior; the old one could
introduce subtle bugs in rare situations (e.g. `GridSearchCV` with a
net that has `warm_start=True`).

The PR also includes tests that are currently failing but that are
passing under the new behavior. When switching to the new behavior,
all tests, including these new ones, should pass (they currently
xfail).

Relates to #521


skorch/net.py
```diff
@@ -1266,6 +1266,15 @@ def _get_params_for_optimizer(self, prefix, named_parameters):
...
         return self.__dict__.keys()
 
+    def _get_param_names_new(self):
+        # TODO: This will be the new behavior for _get_param_names in
+        # a future release. This is to make get_params work as in
+        # sklearn, i.e. not returning "learned" attributes (ending on
+        # '_'). Once the transition period has passed, remove the old
+        # code and use the new one instead.
+        return (k for k in self.__dict__
+                if not k.endswith('_') and k != 'history')
```
thomasjpfan: `history` is... weird parameter.

BenjaminBossan:
> Why is the net_cuda.pkl udpated?

Accidental check in, I reset the file.

> Should we raise a FutureWarning when someone tries to access a attribute that ends in _ with get_params?

How would you achieve this? Return a custom dictionary that checks the key on `__getitem__`?

I would be afraid that such a warning would trigger in unexpected places, e.g. when someone uses `GridSearchCV` and doesn't do anything wrong, he or she could be confused by the warning. I'm all in favor of warning users more prominently about the upcoming change but I wouldn't want to create unnecessary confusion.

skorch/net.py
ottonemo: Maybe we should make `history` a property and store the values in `self.history_`?

BenjaminBossan: I made this change.

CHANGES.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve numerical stability when using `NLLLoss` in `NeuralNetClassifer` (#491)
 - Refactor code to make gradient accumulation easier to implement (#506)
 - NeuralNetBinaryClassifier.predict_proba now returns a 2-dim array; to access the "old" y_proba, take y_proba[:, 1] (#515)
+- net.history is now a property that accesses net.history_, which stores the History object
+
+### Future Changes
+
+- WARNING: In a future release, the behavior of method `net.get_params` will change to make it more consistent with sklearn: it will no longer return "learned" attributes like `module_`. Therefore, functions like `sklearn.base.clone`, when called with a fitted net, will no longer return a fitted net but instead an uninitialized net. If you want a copy of a fitted net, use `copy.deepcopy` instead. Note that `net.get_params` is used under the hood by many sklearn functions and classes, such as `GridSearchCV`, whose behavior may thus be affected by the change.
```
ottonemo:
```suggestion
- WARNING: In a future release, the behavior of method `net.get_params` will change to make it more consistent with sklearn: it will no longer return "learned" attributes like `module_`. Therefore, functions like `sklearn.base.clone`, when called with a fitted net, will no longer return a fitted net but instead an uninitialized net. If you want a copy of a fitted net, use `copy.deepcopy` instead. Note that `net.get_params` is used under the hood by many sklearn functions and classes, such as `GridSearchCV`, whose behavior may thus be affected by the change. (#521, #527)
```

CHANGES.md
```diff
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve numerical stability when using `NLLLoss` in `NeuralNetClassifer` (#491)
 - Refactor code to make gradient accumulation easier to implement (#506)
 - NeuralNetBinaryClassifier.predict_proba now returns a 2-dim array; to access the "old" y_proba, take y_proba[:, 1] (#515)
+- net.history is now a property that accesses net.history_, which stores the History object
```
ottonemo:
```suggestion
- net.history is now a property that accesses net.history_, which stores the History object (#527)
```

PR #531, opened 2019-09-18
support model input x containing extra str value

sakuranew:
it works well

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -46,6 +46,9 @@ def _apply_to_data(data, func, unpack_dict=False):
         return {k: apply_(v) for k, v in data.items()}
 
     if isinstance(data, (list, tuple)):
+        # added by sakuranew
+        if isinstance(data[0],str):
+            return func(data)
         try:
             # e.g.list/tuple of arrays
             return [apply_(x) for x in data]
--- a/skorch/utils.py
+++ b/skorch/utils.py
@@ -77,6 +77,9 @@ def to_tensor(X, device, accept_sparse=False):
 
     if is_torch_data_type(X):
         return X.to(device)
+    #  added by sakuranew 
+    if isinstance(X, str):
+        return X
     if isinstance(X, dict):
         return {key: to_tensor_(val) for key, val in X.items()}
     if isinstance(X, (list, tuple)):


sakuranew:
skorch is amazing ，I like it

BenjaminBossan:
Hi @sakuranew, thanks for the PR. Could you please explain why you need this feature? What problem are you trying to solve?

sakuranew:
> Hi @sakuranew, thanks for the PR. Could you please explain why you need this feature? What problem are you trying to solve?

@BenjaminBossan Thank you for your reply
it maybe not very necessary...
default pytorch dataloader support to get a batch include tensor and string，
for example ，in seq2seq task，we can easily get text rather than text_id to compute belu or print log...
:sweat_smile: actually，this can be done using a vocab，so it is not really necessary, just for ones need

ottonemo:
@sakuranew It would be nice to have an use-case example for this change. Can you provide one?

sakuranew:
@ottonemo 
a toy example：
sorry  the code format become very strange in github....
a


`
import torch.nn as nn
import torch.utils.data
from skorch import NeuralNet
from skorch.callbacks import Callback
from skorch.dataset import CVSplit

class RandomDataset(torch.utils.data.Dataset):
    def __init__(self):
        self.X = torch.randn(3, 10)
        self.X_id = ["hash str 1", "hash str 2", "hash str 3"]
        self.Y = torch.tensor([0., 1., 0.])
    def __getitem__(self, idx):
        return {
                   "id": self.X_id[idx],
                   "x": self.X[idx]
               }, self.Y[idx]
    def __len__(self):
        return 3
train_ds = RandomDataset()
class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(10, 1)

    def forward(self, id, x):
        return self.layer(x)
class Log(Callback):
    def on_batch_end(self, net, X=None, y=None, training=None, **kwargs):
        for id, logits in zip(X["id"], kwargs["y_pred"]):
            print("Id:", id, "Predicted logits:", logits)
            # or save result when inputs have been shuffed
net = NeuralNet(MyModule,max_epochs=1, train_split=CVSplit(2), criterion=torch.nn.MSELoss, callbacks=[Log()])
net.fit(train_ds)`

sakuranew:
a toy example:

```python
import torch.nn as nn
import torch.utils.data
from skorch import NeuralNet
from skorch.callbacks import Callback
from skorch.dataset import CVSplit
class RandomDataset(torch.utils.data.Dataset):
    def __init__(self):
        self.X = torch.randn(3, 10)
        self.X_id = ["hash str 1", "hash str 2", "hash str 3"]
        self.Y = torch.tensor([0., 1., 0.])
    def __getitem__(self, idx):
        return {
                   "id": self.X_id[idx],
                   "x": self.X[idx]
               }, self.Y[idx]
    def __len__(self):
        return 3
train_ds = RandomDataset()
class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(10, 1)
    def forward(self, id, x):
        return self.layer(x)
class Log(Callback):
    def on_batch_end(self, net, X=None, y=None, training=None, **kwargs):
        for id, logits in zip(X["id"], kwargs["y_pred"]):
            print("Id:", id, "Predicted logits:", logits)
            # or save result when inputs have been shuffed
net = NeuralNet(MyModule,max_epochs=1, train_split=CVSplit(2), criterion=torch.nn.MSELoss, callbacks=[Log()])
net.fit(train_ds)

```

BenjaminBossan:
Thanks for providing an example. One of the reasons we don't support str dtypes is that pytorch doesn't support it. The example you provide could be achieved without strings, as you said yourself. If you have another example where strings are strictly necessary, or when pytorch introduces a str tensor, I think we could accept such a change.

BenjaminBossan:
@sakuranew I close the PR for now. If one of the conditions I mentioned in my last post comes up again, we can discuss about reopening this.

PR #534, opened 2019-09-20
change the base callback class default argument Xi to X

sakuranew:


BenjaminBossan:
Thanks for finding this. @ottonemo Do you think this should be changed in `Callback` to `X` or rather in `NeuralNet` to `Xi`?

BenjaminBossan:
ping @ottonemo

PR #536, opened 2019-09-21
Bugfix/train end cp with load init state 528

### LINKED ISSUE #528: [BUG] TrainEndCheckpoint not working with LoadInitState
I'm trying to use `TrainEndCheckpoint` with `LoadInitState` but it's not working. It doesn't even return any error as you are using `with suppress(Exception)`. The suppressed error is `*** AttributeError: 'TrainEndCheckpoint' object has no attribute 'f_history_'`, yet the file exists and `self.checkpoint.checkpoint_.f_history_==.../train_end_history.json'`.

I think the issue is that `TrainEndCheckpoint` stopped inheriting from `Checkpoint` in #463. I.e. it doesn't have a `f_history_` attribute anymore. This could be simply resolved by redirecting `self.f_history_` to `self.checkpoint_.f_history_`.

BenjaminBossan:
Fixes #528 

@yanndubs Can you verify that this solves your problem?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a bug that caused LoadInitState not to work with TrainEndCheckpoint (#528)
+
 
 ## [0.6.0] - 2019-07-19
 
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -381,7 +381,6 @@ class EarlyStopping(Callback):
             self.sink(text)
 
 
-
 class ParamMapper(Callback):
     """Map arbitrary functions over module parameters filtered by pattern
     matching.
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
I thought about such a solution but wasn't sure if `f_history_` is the only affected attribute. Instead of enumerating all, I went with the broader solution. If that can cause problems, I would, however, prefer the more specific solution.

thomasjpfan:
I guess it is okay as long as we do not add any more parameters to `TrainEndCheckpoint`.

BenjaminBossan:
> I guess it is okay as long as we do not add any more parameters to `TrainEndCheckpoint`.

But even that shouldn't be a problem: as long as we set the parameter as an attribute on `TrainEndCheckpoint`, `__getattr__` shouldn't be triggered.

thomasjpfan:
Thanks for the info. (Just read up on the difference between `__getattribute__` and `__getattr__`)

Issue #539, opened 2019-09-24
Why does it use only one fixed validation in Kfold?


rightx2:
`skorch.dataset.CVSplit` said,

```
However, instead of cycling through all splits, only one fixed split (the first one) is used.
``` 

Is there any reason? Can't understand why it has to use only 1 fold even when 5 is given as cv split.

githubnemo:
I don't think there is a very clever reason. The purpose of `CVSplit` is to provide validation data while re-using sklearn's CV infrastructure and we chose this way of implementing exactly that. Therefore, the number of folds is just a proxy for specifying the train/valid ratio.

If you think that defining a host of validation sets and cycling through them is a good idea, we are happy to hear you out!

BenjaminBossan:
See my response here: https://github.com/skorch-dev/skorch/pull/540#issuecomment-535132966

ottonemo:
I think the main issue here is that it is unclear how to do early stopping while doing cross-validation. This will be handled in https://github.com/skorch-dev/skorch/issues/544.

PR #540, opened 2019-09-24
Draft for complete cv looping over all folds instead of fixed 1 fold

rightx2:
This pull request is related with #539 

I used `yield` structure to use all folds given when doing cross-validation.

I checked it runs but problem is, in `fit_loop`, need your helps on extending `self.history`, `self.notify` in this new structure. 

I think averaging validation scores after validation is also needed.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -322,10 +322,10 @@ class CVSplit:
         if self._is_stratified(cv):
             args = args + (to_numpy(y),)
 
-        idx_train, idx_valid = next(iter(cv.split(*args, groups=groups)))
-        dataset_train = torch.utils.data.Subset(dataset, idx_train)
-        dataset_valid = torch.utils.data.Subset(dataset, idx_valid)
-        return dataset_train, dataset_valid
+        for idx_train, idx_valid in cv.split(*args, groups=groups):
+            dataset_train = torch.utils.data.Subset(dataset, idx_train)
+            dataset_valid = torch.utils.data.Subset(dataset, idx_valid)
+            yield dataset_train, dataset_valid
 
     def __repr__(self):
         # pylint: disable=useless-super-delegation
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -709,50 +709,49 @@ class NeuralNet:
 
         """
         self.check_data(X, y)
-        epochs = epochs if epochs is not None else self.max_epochs
 
-        dataset_train, dataset_valid = self.get_split_datasets(
-            X, y, **fit_params)
-        on_epoch_kwargs = {
-            'dataset_train': dataset_train,
-            'dataset_valid': dataset_valid,
-        }
+        for dataset_train, dataset_valid in self.get_split_datasets(X, y, **fit_params):
+            epochs = epochs if epochs is not None else self.max_epochs
+            on_epoch_kwargs = {
+                'dataset_train': dataset_train,
+                'dataset_valid': dataset_valid,
+            }
 
-        y_train_is_ph = uses_placeholder_y(dataset_train)
-        y_valid_is_ph = uses_placeholder_y(dataset_valid)
+            y_train_is_ph = uses_placeholder_y(dataset_train)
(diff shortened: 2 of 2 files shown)


BenjaminBossan:
I think there is a fundamental misunderstanding here. The skorch-internal train/valid split is not meant to cycle through all possible split. If you do this, you would train the neural net on the whole dataset, which results in the validation scores not having any value anymore. You wouldn't be able to use the validation scores for gauging your model performance, use it for early stopping, etc.

For what you are trying to achieve, there are other tools. If you want to get the model score on all splits, use sklearn's [`cross_validate`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html) and turn off skorch-internal cross validation (`train_split=False`). If you want all out of fold predictions, use sklearn's [`cross_val_predict`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_predict.html). With these functions, you train independent models for each fold, therefore there is no danger of leakage.

rightx2:
@BenjaminBossan    You said "The skorch-internal train/valid split is not meant to cycle through all possible split. If you do this, you would train the neural net on the whole dataset, which results in the validation scores not having any value anymore" -> It can be easily solved by fitting only 80% dataset instead of whole one (e.g. net.fit(X_80_percent, y_80_percent)).   If I used `cross_validate` api of sklearn and set `train_split=None`,  I cannot use `skorch.callbacks.EarlyStopping` as a callback in NeurlaNetRegressor cuz it requires me `valid_loss`!

githubnemo:
If you don't reset the network between switching folds you will essentially leak validation data from one of the previous folds into your training data, essentially ruining your validation scores. If you do a `cross_validate` and there would be a way to let skorch know what the current validation set is to do early stopping, you would essentially optimize your parameters on your "test" set, also not a good idea. Therefore the best way would be to have a separate portion of validation data extracted from your training data and give it to the neural net to do early stopping/checkpointing. This way you can use cross validation via sklearn and still use early stopping without leaks.

Example:

```python
from skorch.helper import predefined_split
X_train, y_train, X_valid, y_valid = train_test_split(...)
net = NeuralNetClassifier(module, train_split=predefined_split(Dataset(X_valid, y_valid)))
result = cross_validate(net, X_train, y_train, ...)
```

rightx2:
@githubnemo I agree with your opinion about leakage and what I was mean to say that skorch has to reset model on every each fold is set to be valid set, of course!
And your code would prevent data leakage but not complete cross-validation because this will fix the validation set and let the model fit all other folds and early stopping the validation set...which could occur overfitting on specific data fold...

BenjaminBossan:
You don't need to split the data used for early stopping beforehand. Instead, you can just leave skorch-internal cross validation on and use `cross_validate`. That way, for each CV fold, there will be another internal validation set used for early stopping as well. It really depends on what you want to achieve.

ivanpmartell:
@BenjaminBossan are there any complete 'dummy proof' examples or tutorials on how to do both cross validation and early stopping? 
I have a skorch dataset and I want to be able to use cross validation for it, but I haven't found much information out there to do this, or it gets into a complicated mess with the dataset.

``` python
import numpy as np
import torch
from ICNN.module import ICNNModule
from ICNN.dataset import ICNNDataset
from skorch import NeuralNetBinaryClassifier
from skorch.dataset import CVSplit
from skorch.callbacks import EarlyStopping
from sklearn.model_selection import cross_val_score

net = NeuralNetBinaryClassifier(ICNNModule,
                          max_epochs=200,
                          lr=0.01,
                          callbacks=[EarlyStopping(patience=25)],
                          batch_size=16,
                          optimizer=torch.optim.SGD,
                          optimizer__momentum=0.90,
                          train_split=CVSplit(cv=10,stratified=True,random_state=0))


ds = ICNNDataset("data/icnn.csv")
y_train = np.array([y for X, y in iter(ds)])
net.fit(ds, y_train)
```
This is what I have so far, but I would like to have all out of fold scores. I'm new to skorch and any help would be greatly appreciated. Thanks!

PR #541, opened 2019-09-25
MAINT Calls scorer directly

thomasjpfan:
This PR removes the use of the private `_score` function and calls the scorer directly.

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -7,7 +7,6 @@ from functools import partial
 import numpy as np
 from sklearn.metrics.scorer import (
     check_scoring, _BaseScorer, make_scorer)
-from sklearn.model_selection._validation import _score
 
 from skorch.utils import data_from_dataset
 from skorch.utils import is_skorch_dataset
@@ -116,14 +115,7 @@ class ScoringBase(Callback):
         """Resolve scoring and apply it to data. Use cached prediction
         instead of running inference again, if available."""
         scorer = check_scoring(net, self.scoring_)
-        scores = _score(
-            estimator=net,
-            X_test=X_test,
-            y_test=y_test,
-            scorer=scorer,
-            is_multimetric=False,
-        )
-        return scores
+        return scorer(net, X_test, y_test)
 
     def _is_best_score(self, current_score):
         if self.lower_is_better is None:


BenjaminBossan:
Oh wow, I wonder why I made it so complicated (I think it was me anyway).

Out of curiosity, sklearn now allows scorers to be dicts. What would be the expected behavior in skorch?

ottonemo:
> Oh wow, I wonder why I made it so complicated (I think it was me anyway).

It was you and I remember that we discussed this for quite some time and I think we ended up with this solution since there was no way to invoke arbitrary scorers directly. This should have been documented in the code but apparently we didn't think of that :/

Anyway, it seems fine and I can't find a reason against doing it like this.

> Out of curiosity, sklearn now allows scorers to be dicts. What would be the expected behavior in skorch?

My imagination fails me in this case, I can't think of a case where a user would want to pass a dictionary of scores to `EpochScoring`, for example.

thomasjpfan:
> Out of curiosity, sklearn now allows scorers to be dicts. What would be the expected behavior in skorch?

sklearn does not supply a public interface for calling a multimetric scorer. In `skorch`, a user can still have multiple `EpochScoring` callbacks for every metric. This would not be the most optimal thing to do. For example, if two scorers need to call `predict`, both `EpochScoring` callbacks would have to call predict.

In sklearn, we recently updated the multimetric scorers to optimize the above situation: https://github.com/scikit-learn/scikit-learn/pull/14593. This interface is currently private, but it may be nice to have publicly. Maybe a `make_multimetric_scorer`, which returns a callable that has the same interface as a scorer but returns a dictionary.

For `skorch`, we can be safe and error if a user passes a dictionary.

BenjaminBossan:
> For example, if two scorers need to call `predict`, both `EpochScoring` callbacks would have to call predict.

That's why we went down the caching route. This can be prone to certain kind of subtle errors, but I believe it was overall a pragmatic choice. (Btw.: What I used to do to avoid multiple calls to `predict*` in case of multimetric in sklearn was to use caching with a hash of the incoming data that resets on `fit` -- though never in prod.)

> My imagination fails me in this case, I can't think of a case where a user would want to pass a dictionary of scores to EpochScoring, for example

In my imagination, users could pass a dict if they want to have multiple metrics and if they know this from sklearn. However, it's going to be hard to support that, since we need more information than just the scorer from the user (all parameters from `EpochScoring` et al.).

I remember that we used to have an interface for scoring that allowed to pass more than one scoring function. It was less ergonomic than the current implementation, so I wouldn't try to force this on skorch again.

> For `skorch`, we can be safe and error if a user passes a dictionary.

Yes, with a suggestion to use multiple scoring callbacks instead. I created an issue for that: #542

Issue #486, opened 2019-06-18
Troubleshooting Pytorch classifier with cleanlab


mattare2:
Thanks for this very helpful library! I'm trying to use my skorch-wrapped NLP classifier with the [cleanlab](https://github.com/cgnorthcutt/cleanlab) library.  My classifier works fine for training, but when I try to use the cleanlab code, I get the following error:

```
from cleanlab.classification import LearningWithNoisyLabels

# Wrap around any classifier. Yup, you can use sklearn/pyTorch/Tensorflow/FastText/etc.
lnl = LearningWithNoisyLabels(clf=net)
lnl.fit(X = np.array([f.input_ids for f in features]), s = np.array([f.label_ids for f in features]))

# Estimate the predictions you would have gotten by training with *no* label errors.
lnl.predict(val_ds.X)
```

    <ipython-input-272-eb1ed08448d3> in <module>()
    ----> 1 lnl.predict(val_ds.X)
    
    ~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/cleanlab/classification.py in predict(self, *args, **kwargs)
        320           The test data as a feature matrix.'''
        321 
    --> 322         return self.clf.predict(*args, **kwargs)
        323 
        324 
    
    ~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/skorch/classifier.py in predict(self, X)
        187         """
        188         y_preds = []
    --> 189         for yp in self.forward_iter(X, training=False):
        190             yp = yp[0] if isinstance(yp, tuple) else yp
        191             y_preds.append(to_numpy(yp.max(-1)[-1]))
    
    ~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/skorch/net.py in forward_iter(self, X, training, device)
        883         iterator = self.get_iterator(dataset, training=training)
        884         for Xi, _ in iterator:
    --> 885             yp = self.evaluation_step(Xi, training=training)
        886             if isinstance(yp, tuple):
        887                 yield tuple(n.to(device) for n in yp)
    
    ~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/skorch/net.py in evaluation_step(self, Xi, training)
        679         """
        680         with torch.set_grad_enabled(training):
    --> 681             self.module_.train(training)
        682             return self.infer(Xi)
        683 
    
    AttributeError: 'NeuralNetClassifier' object has no attribute 'module_'

Any idea where to start?

[...]

Issue #465, opened 2019-05-01
NeuralNetBinaryClassifier with CalibratedClassiferCV


ZaydH:
I am wrapping a `torch` `Module` inside `NeuralNetBinaryClassifier` to perform binary classification.  My algorithm requires that the output of the neural network be *well-calibrated*, and neural networks do not generatively yield well-calibrated outputs.  

For other non-well calibrated classifiers in `sklearn` (e.g., SVM), I use `sklearn`'s wrapper class `CalibratedClassifierCV`.  However, when I try to wrap a `NeuralNetBinaryClassifier` with `CalibratedClassifierCV`, `sklearn` raises a runtime error "`IndexError: too many indices for array`" since `CalibratedClassifierCV` expects the output tensor of `predict_proba` to be (*n* x 2) in shape (where *n* is the batch size) while `NeuralNetBinaryClassifier` returns a 1-D tensor of size *n*.

**`sklearn` module `calibration.py` lines 303-306**
```
        elif hasattr(self.base_estimator, "predict_proba"):
            df = self.base_estimator.predict_proba(X)
            if n_classes == 2:
                df = df[:, 1:]   # <- Error Here
```

Is there a reasonably straightforward way to make these two classes compatible?  Some other `sklearn` binary classifiers (e.g., logistic regression) output 2-D tensors.  Would it be worth it to do the same for `NeuralNetBinaryClassifier`?  For loss calculation, you could just use the second column in the `Tensor` similar to what `

BenjaminBossan:
Hmm, I'm not quite sure what went wrong here. Personally, I always use `NeuralNetClassifier`, even for the binary case. Would that work for you?

ZaydH:
Sure I will try that and update.  Is `NeuralNetBinaryClassifier` not intended to be used?  Do you think it is worth updating the class to return a 2-D tensor?  If so, I can help with contributing.

ZaydH:
One issue that arose is `CalibratedClassifierCV` expects a `classes_` field for the underlying learner.  

I did a quick `grep` and noticed the field in some of the other `sklearn` classes.  I expect this issue would apply for both `NeuralNetBinaryClassifier` and `NeuralNetClassifier`.  Was this left out of `skorch` for a specific reason?

BenjaminBossan:
[...]

PR #546, opened 2019-10-13
Add `classes_` attribute `NeuralNetClassifier`

BenjaminBossan:
This is inferred from y default but can be overridden by passing
`classes` explicitly during initialization.

Also, add more rigorous tests to classifiers and regressor.
That is to catch errors with specific attributes being set on those
that are not set on `NeuralNet` itself. This can potentially mess with
cloning, which is why it should be tested explicitly.

Addresses #465, #486

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -28,18 +28,31 @@ neural_net_clf_doc_start = """NeuralNet for classification tasks
 
 """
 
-neural_net_clf_criterion_text = """
+neural_net_clf_additional_text = """
 
     criterion : torch criterion (class, default=torch.nn.NLLLoss)
       Negative log likelihood loss. Note that the module should return
-      probabilities, the log is applied during ``get_loss``."""
+      probabilities, the log is applied during ``get_loss``.
+
+    classes : None or list (default=None)
+      If None, the ``classes_`` attribute will be inferred from the
+      ``y`` data passed to ``fit``. If a non-empty list is passed,
+      that list will be returned as ``classes_``. If the initial
+      skorch behavior should be restored, i.e. raising an
+      ``AttributeError``, pass an empty list."""
+
+neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
+      A list of class labels known to the classifier.
+
+"""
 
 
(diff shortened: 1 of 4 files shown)


skorch/classifier.py
```diff
@@ -99,6 +123,8 @@ def check_data(self, X, y):
                    "``iterator_train`` and ``iterator_valid`` parameters "
                    "respectively.")
             raise ValueError(msg)
+        if y is not None:
+            self.classes_inferred_ = np.unique(y)
```
thomasjpfan: `classes_inferred_` can be private.

BenjaminBossan: That's the first thing I did, but this has unintended side effects on `clone`, resulting in it failing the consistency check. This might be something we should investigate further but I thought it won't hurt to have this attribute there. I wouldn't document it in case we do want to remove it later.

githubnemo:
> We need to make sure that `predict` will return the classes used in `fit`.

This doesn't feel strictly necessary to me. Can you elaborate?

githubnemo:
I would recommend merging this and, if it turns out to be an common issue, add checks in the inference step for class consistency.

thomasjpfan:
> We need to make sure that predict will return the classes used in fit.

That's if we really want to comply with the API. For example:

```py
from sklearn.datasets import load_iris
from sklearn.linear_model import LogisticRegression
X, y = [[0], [1]], ['dog', 'cat']
clf = LogisticRegression(random_state=0, solver='lbfgs').fit(X, y)

clf.classes_
clf.classes_
# array(['cat', 'dog'], dtype='<U3')

clf.predict(X)
# array(['dog', 'cat'], dtype='<U3')
```

BenjaminBossan:
I forgot this was possible, it fits a `LabelEncoder` internally, right? Is this supported by all classifiers?

Anyway, I'm not 100% convinced we need that, and would thus propose to postpone the decision.

thomasjpfan:
I am suggesting we check that `classes_` are ints if it was passed in `__init__` during `fit`?

BenjaminBossan:
> I am suggesting we check that classes_ are ints if it was passed in __init__ during fit?
Too late :D

PR #555, opened 2019-11-10
Notebooks explicitly cast int64

### LINKED ISSUE #554: Bug: Error in Basic Usage Example
When following the example notebook provided [here](https://nbviewer.jupyter.org/github/skorch-dev/skorch/blob/master/notebooks/Basic_Usage.ipynb), there is an error with the Neural Net Classifier.

RuntimeError: Expected object of scalar type Long but got scalar type Int for argument #2 'target'

BenjaminBossan:
Fixes #554 

Explicitly cast `y` to int64 in notebooks (int32 doesn't work).

Also removes a stray `pdb on` cell.


BenjaminBossan:
Let's leave this open to also address the point raised [here](https://github.com/skorch-dev/skorch/issues/554#issuecomment-552535391).

ottonemo:
If I understand correctly, this can be merged, right?

BenjaminBossan:
Could be merged from my point of view.

Issue #552, opened 2019-11-07
EpochScoring caching does not work in every case


marrrcin:
I wanted to use skorch `EpochScoring` with dataset ~100GB stored as files in GCS (or S3 does not really matter here). Dataset needs to be loaded on the fly due to memory constraints, so in every epoch, data is downloaded again.

**Reproduction steps:**
1. Implement Dataset (PyTorch) that loads the data on the fly from external source (HTTP/GCS/S3)
1. Create Dataset(s) for training and validation (`train_ds`, `val_ds`)
1. Create `NeuralNetClassifier` for your data, with the following args:
```
train_split=predefined_split(val_ds),
```
1. Add `EpochScoring` callbacks (i.e for Precision, Recall and F1) for both training and validation, *with caching*
1. Run `network.fit(train_ds, y=None)`

**Expected result:**
`EpochScoring` runs fast, because data for `y` and `y_pred` should be already cached.

**Actual result:**
`EpochScoring` is slow, although the cache is enabled. The cause of it is that it used `network.predict` under the hood, which does the iteration across the whole validation dataset again. So if I put 6 metrics in callbacks, I had to loop through the whole dataset 6 times.

I think that `EpochScoring` should call scorers directly, when caching is enabled, because the case above shows that overwriting `infer` at the time of running the callback does not cover every case properly and the provided cache is useless in it.

BenjaminBossan:
Thanks for the detailed report. This is true, we hadn't considered the case where the pure iteration over the dataset could be a bottleneck.

The whole caching part is somewhat delicate and we tried a bunch of different solutions, all with some benefits and drawbacks. My first idea for this problem would be to patch `forward_iter` instead of `infer`. It is a bit more involved but could work. @ottonemo, what do you think?

> I think that `EpochScoring` should call scorers directly,

I'm not exactly sure what you mean. I think it's important that we can reuse as much of the existing infrastructure as possible, which is why went with the "pretending to predict" route in the first place. If you have an example that implements your proposal, we can discuss this.

marrrcin:
The nasty workaround I did just for the same of forcing this to work is the following:

```
class FastEpochScoring(EpochScoring):
    def __init__(self, scoring, multi_class=True, lower_is_better=True, on_train=False, name=None,
                 target_extractor=to_numpy,
                 use_caching=True):
        super().__init__(scoring, lower_is_better, on_train, name, target_extractor, use_caching)
        self.multi_class = multi_class

[...]

PR #557, opened 2019-11-11
Feature skip data iteration when caching scoring

BenjaminBossan:
This is the proposed fix for #552.

As discussed there, my proposal is to cache `net.forward_iter`, so that we can skip the iteration over the data.

It would be very helpful if someone could think of any unintended side-effects this change could have. E.g., in theory, some code could rely on iterating over the data even in case of caching, but I can hardly imagine this happening in reality.

Also, I cannot test on GPU for the moment. I don't see how this would affect the outcome, but it could still be nice of someone verifies it works.

I deprecated `cache_net_infer` because it was not "private". However, I made the new `_cache_net_forward_iter` private.


marrrcin:
Can I tests this from some nightly build or do I have to build the package on my own?

BenjaminBossan:
@marrrcin You would need to install from source, but it's not difficult: https://github.com/skorch-dev/skorch#from-source

skorch/net.py
```diff
@@ -911,10 +911,12 @@ def forward_iter(self, X, training=False, device='cpu'):
...
-            if isinstance(yp, tuple):
-                yield tuple(n.to(device) for n in yp)
-            else:
-                yield yp.to(device)
+            yield self._forward_output(yp, device=device)
+
+    def _forward_output(self, yp, device):
+        if isinstance(yp, tuple):
+            return tuple(n.to(device) for n in yp)
+        return yp.to(device)
```
ottonemo: This is structurally very similar to `skorch.utils.to_tensor`. Maybe we should introduce a `skorch.utils.to_device` instead? This might also become handy if we support multiple GPUs in the future.

BenjaminBossan: Good point, I moved this to skorch.utils.to_device

BenjaminBossan:
@ottonemo I addressed your comment, pls review again.

Before merging this, I should we consider making a new release? I wouldn't mind this new feature being only on master for some time in case it creates some trouble down the line.

PR #558, opened 2019-11-12
Fix squeeze inside NeuralNetBinaryClassifier.infer

qtux:
Previously y_infer.squeeze() removed all dimensions with single entries,
which does not correspond to the docstring description. Instead, remove
only the last dimension, if it contains a single entry.

### DIFF
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -235,6 +235,15 @@ class TestNeuralNetBinaryClassifier:
         valid_acc = net.history[-1, 'valid_acc']
         assert valid_acc > 0.65
 
+    def test_batch_size_one(self, net_cls, module_cls, data):
+        X, y = data
+        net = net_cls(
+            module_cls,
+            max_epochs=1,
+            batch_size=1,
+        )
+        net.fit(X, y)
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for the PR. I think this change should be okay.

Could you please add an entry to the CHANGES.md?

qtux:
Sure, I will rebase the commit.

qtux:
I added a test to show the bug and replaced the squeeze by a reshape (was not correct previously).

skorch/tests/test_classifier.py
```diff
@@ -235,6 +235,16 @@ def test_net_learns(self, net_cls, module_cls, data):
         valid_acc = net.history[-1, 'valid_acc']
         assert valid_acc > 0.65
 
+    def test_batch_size_one(self, net_cls, module_cls, data):
+        X, y = data
+        print("test", y.shape)
```
ottonemo:
```suggestion
```

qtux:
I removed the print statement from the test and rebased the commit.

BenjaminBossan:
Thank you very much for the bugfix.

Issue #542, opened 2019-09-29
Raise a helpful error if dict as passed to scoring callback


BenjaminBossan:
sklearn allows to pass a dictionary as scorer for multimetric scoring. For skorch, a user should pass multiple scoring callbacks instead.

Right now, if a user passes a dict, skorch will fail while trying to retrieve a name here:

https://github.com/skorch-dev/skorch/blob/0581bc6062df86a46a7857ffeab0787e524abbe0/skorch/callbacks/scoring.py#L99

Or here if a name was passed:

https://github.com/skorch-dev/skorch/blob/0581bc6062df86a46a7857ffeab0787e524abbe0/skorch/callbacks/scoring.py#L118

Ideally, there should be an exception before that which will inform the user to use multiple instances of scoring callbacks instead.

BenjaminBossan:
Solved via #560

PR #560, opened 2019-11-17
Raise helpful error if dict passed to scoring callback

lofifnc:
Introduces fix for issue #542 informing the user of the incompatibility in regards to multi-metric evaluation with scikit-learn.

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -81,6 +81,7 @@ class ScoringBase(Callback):
     def _get_name(self):
         """Find name of scoring function."""
         if self.name is not None:
+            print("name", self.name)
             return self.name
         if self.scoring_ is None:
             return 'score'
@@ -90,6 +91,8 @@ class ScoringBase(Callback):
             return self.scoring_.func.__name__
         if isinstance(self.scoring_, _BaseScorer):
             return self.scoring_._score_func.__name__
+        if isinstance(self.scoring_, dict):
+            raise ValueError("Dict not supported for scoring. If you wan't to use multiple scores pass multiple callbacks")
         return self.scoring_.__name__
 
     def initialize(self):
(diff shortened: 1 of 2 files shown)


skorch/callbacks/scoring.py
```diff
@@ -81,6 +81,7 @@ def __init__(
     def _get_name(self):
         """Find name of scoring function."""
         if self.name is not None:
+            print("name", self.name)
```
BenjaminBossan: Could you remove this please? This is probably also the reason the tests fail.

BenjaminBossan:
Thanks for the contribution, Alex. Removing the debugging print statement should hopefully fix the tests.

skorch/callbacks/scoring.py
lofifnc: yea my bad

BenjaminBossan:
Could you please also check the error message within the unit test? `ValueError` is very general and could also be triggered by something else.

skorch/tests/callbacks/test_scoring.py
```diff
@@ -457,6 +457,22 @@ def test_multiple_scorings_share_cache(
         for c1, c2 in zip(cbs['a1'].y_trues_, cbs['a2'].y_trues_):
             assert id(c1) == id(c2)
 
+    def test_multiple_scorings_with_dict(
+            self, net_cls, module_cls, train_split, caching_scoring_cls, data):
```
thomasjpfan:
```suggestion
            self, net_cls, module_cls, train_split, scoring_cls, data):
```

skorch/tests/callbacks/test_scoring.py
```diff
@@ -457,6 +457,22 @@ def test_multiple_scorings_share_cache(
...
+        # This test checks if an exception is raised when a dictionary is passed as scorer.
+        net = net_cls(
+            module=module_cls,
+            callbacks=[
+                caching_scoring_cls({'a1': 'accuracy', 'a2': 'accuracy'}),
```
thomasjpfan:
```suggestion
                scoring_cls({'a1': 'accuracy', 'a2': 'accuracy'}),
```

BenjaminBossan:
Thank you very much.

I didn't know about the `match` keyword in `pytest.raises`, we could actually update quite a few of our tests this way.

Issue #561, opened 2019-11-18
PyTroch >= 1.1.0 scheduler.step() change


mshuaibii:
Not sure if this has been mentioned yet. This isn't critical to functionality but would be nice to address - warning message arises when any lr scheduler is turned on:

```UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`.  Failure to do this will result in PyTorch skipping the first value of the learning rate schedule.See more details at https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate
  "https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate", UserWarning)```

ottonemo:
Thanks for highlighting this. 

This was first discussed in https://github.com/skorch-dev/skorch/issues/467 but the step order was not changed then since we didn't bump the minimum PyTorch version for skorch (to avoid problems with pre-1.1.0 releases).

This is something we should address now as our requirement is PyTorch >=1.1.0.

ottonemo:
Nevermind me, this was already addressed in af04b559445a76da33a2750ff630dee2749402d3 and there are only some warnings that are from tests that don't call `optimizer.step` (for good reason).

@mshuaibii Can you tell us which skorch/PyTorch version you are using and how you triggered this behavior?

mshuaibii:
@ottonemo 

I'm on PyTorch version 1.3.0 and skorch version 0.5.0.post0. Just noticed skorch is on version 0.6.0 so haven't tested that out to see if the issue is resolved. But to replicate - I defined a `LRScheduler` callback in the `NeuralNetRegressor`.

ottonemo:
Yes, please do test with skorch 0.6 - this is probably going to resolve your problem.

mshuaibii:
> Yes, please do test with skorch 0.6 - this is probably going to resolve your problem.

Yup - 0.6.0 took care of it. Thanks!

PR #562, opened 2019-11-19
Fix wrongful LR scheduler step order warnings

ottonemo:
With PyTorch 1.1.0 the optimizer step order was fixed (see af04b559445a76da33a2750ff630dee2749402d3) and a warning is issued when the order is violated. Since some LR scheduler tests don't call the optimizer at all this warning is falsely issued. We can fix this by calling `optimizer.step` anyway.

Relates to #467 and #561.


BenjaminBossan:
Thank you very much.

Issue #497, opened 2019-07-21
Should `load_params(cp=Checkpoint)` always initialize when needed?


thomasjpfan:
Should `load_params(cp=Checkpoint)` always initialize when `initialized_` is False?

BenjaminBossan:
I haven't thought of this, is there any harm in doing so?

thomasjpfan:
I don’t see any harm in doing this. One can only load from a checkpoint when the module is initialized.

BenjaminBossan:
Sounds good. Perhaps there are even other spots where we could initialize out of convenience.

PR #563, opened 2019-11-19
ENH Load_params with a checkpoint initializes when needed

thomasjpfan:
Closes https://github.com/skorch-dev/skorch/issues/497

This updates `NeuralNet.load_params` with a checkpoint initializes when needed


BenjaminBossan:
Thanks, LGTM. Let's wait a day to see if @ottonemo has any comments, otherwise this can be squashed and merged.

BenjaminBossan:
Seems we're good to go :)

PR #569, opened 2019-12-01
Fix link for google colab in MNIST torchvision notebook

lofifnc:


BenjaminBossan:
Good catch, thanks for the PR.

Issue #570, opened 2019-12-04
Deperecated check_is_fitted call


fradav:
When calling GridsearchCV I got :
```plain
/home/fradav/.conda/envs/pytorch/lib/python3.7/site-packages/scikit_learn-0.22rc3-py3.7-linux-x86_64.egg/sklearn/utils/validation.py:937: FutureWarning: Passing all_or_any to check_is_fitted is deprecated and will be removed in 0.23. The any_or_all argument is ignored.
```
This is with scikit-learn `0.22rc3` version.

Regards,

thomasjpfan:
Can you provide a reproducible code example where the warning appears?

BenjaminBossan:
We use sklearn's `check_is_fitted` with that parameter here:

https://github.com/skorch-dev/skorch/blob/7d275df28d49de782373174a0882bc8c31fda844/skorch/utils.py#L502

Since the parameter was deprecated [here](https://github.com/scikit-learn/scikit-learn/pull/145459), we should also remove it.

The question now is if we want to deprecate it as well or just flat out remove it. My guess is that nobody uses `skorch.utils.check_is_fitted` directly, which would make it safe for immediate removal. If we deprecate, people will see two deprecation warnings while just using normal skorch code, and they cannot even easily fix it.

thomasjpfan:
It still has a use case in skorch by raising skorch specific exceptions. Since it is not being used, we can change its behavior without much consequences.

A quick fix would be to check the sklearn version and make sure to call `check_is_fitted` with the correct arguments (depending on version).

BenjaminBossan:
I just saw that the changes to `check_is_fitted` make it unusable for use. E.g.:

```
    attributes : deprecated, ignored
        .. deprecated:: 0.22
           `attributes` is deprecated, is currently ignored and will be removed
           in 0.23.
```

Instead, the function now only checks if there are *any* attributes that end on `_`. This is the case for skorch even before fitting.

fradav:
There is also : 
```plain
/data/miniconda3/envs/pytorch/lib/python3.6/site-packages/sklearn/utils/deprecation.py:144: FutureWarning: The sklearn.metrics.scorer module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.metrics. Anything that cannot be imported from sklearn.metrics is now part of the private API.
  warnings.warn(message, FutureWarning)
```

Should I make a separate issue for this?

[...]

PR #571, opened 2019-12-05
Bugfix: sklearn 0.22

BenjaminBossan:
Currently, skorch won't work with sklearn 0.22 because some code was changed that is more geared towards sklearn-internal use, which we, however, relied upon.

A list of failures:

1. `check_is_fitted` was changed (see #570) in a way that we won't raise a `NotInitializedError` anymore despite the net not being initialized. This is now solved by basically porting the old `check_is_fitted` behavior to skorch.

2. `safe_indexing` was changed (probably [here](https://github.com/scikit-learn/scikit-learn/pull/14475)) in a way that it no longer works with tuples as indices; the following code works in sklearn 0.21 but not in 0.22:

```python
from sklearn.utils import safe_indexing  # or _safe_indexing
import numpy as np


X = np.zeros((5, 3))
i = ([2, 4, 1, 0, 3], [1, 0, 2, 2, 1])
safe_indexing(X, i)
```

This is solved by not using `safe_indexing` in case of tuples anymore.

3. Our `convert_sklearn_metric_function` now fails with `make_scorer(...)` because that returns an sklearn `_PredictScorer` object (or similar), which we don't detect. This is now detected.

The changes were tested with scikit-learn=0.21.3 and scikit-learn==0.22 (no GPU).

Furthermore, pylint now complains about imports that are not at top level (basically about all our tests), which is why I blankly removed that warning.

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -46,11 +46,16 @@ def convert_sklearn_metric_function(scoring):
     sklearn scorer and return it. Otherwise, return ``scoring`` unchanged."""
     if callable(scoring):
         module = getattr(scoring, '__module__', None)
+
+        # those are scoring objects returned by make_scorer starting
+        # from sklearn 0.22
+        scorer_names = ('_PredictScorer', '_ProbaScorer', '_ThresholdScorer')
         if (
                 hasattr(module, 'startswith') and
                 module.startswith('sklearn.metrics.') and
                 not module.startswith('sklearn.metrics.scorer') and
-                not module.startswith('sklearn.metrics.tests.')
+                not module.startswith('sklearn.metrics.tests.') and
+                not scoring.__class__.__name__ in scorer_names
         ):
             return make_scorer(scoring)
     return scoring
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
I'm not so happy about these breaking changes being made by sklearn, since they affect public API and, at least from skimming, are not announced in the changelog.

thomasjpfan:
The util module has been in this gray area regarding changing behavior, as stated in https://scikit-learn.org/stable/developers/utilities.html under “warning”. Historically this policy was put into place, because public functions were put into utils without throughly testing its functionality.

I see how these types of changes can be a bit painful. Going forward, if sklearn wants to better support an ecosystem of estimators, I think sklearn needs to consider some of its utils as “first class” (this means deprecations) and not this gray area. I’ll be sure to bring this up with the rest of the team.

BenjaminBossan:
> The util module has been in this gray area regarding changing behavior, as stated in https://scikit-learn.org/stable/developers/utilities.html under “warning”

Interesting. I didn't know that.

sklearn is a huge codebase, so these things are bound to happen. I think it's especially important to communicate these kind of changes in the release notes.

> I’ll be sure to bring this up with the rest of the team.

Great, thanks!

skorch/callbacks/scoring.py
```diff
@@ -46,11 +46,16 @@ def convert_sklearn_metric_function(scoring):
...
+        # those are scoring objects returned by make_scorer starting
+        # from sklearn 0.22
+        scorer_names = ('_PredictScorer', '_ProbaScorer', '_ThresholdScorer')
         if (
                 hasattr(module, 'startswith') and
                 module.startswith('sklearn.metrics.') and
                 not module.startswith('sklearn.metrics.scorer') and
-                not module.startswith('sklearn.metrics.tests.')
+                not module.startswith('sklearn.metrics.tests.') and
+                not scoring.__class__.__name__ in scorer_names
```
thomasjpfan: Is the `scoring.__class__.__name__` check needed? `make_scorer` returned these objects in 0.21 as well.

BenjaminBossan: I don't quite understand your suggestion. We need this new line to to check if we get any of these objects. I tried to remove the new line and get failing tests with sklearn 0.22 but not 0.21.

PR #572, opened 2019-12-05
Unify predict_proba for neural net classifiers

qtux:
I have changed the `predict_proba` method from the `NeuralNetClassifier` similarly to the `predict_proba` method of the `NeuralNetBinaryClassifier`, to allow the use of the `CrossEntropyLoss` instead of the `NLLLoss` (in a similar way to using `BCEWithLogitsLoss` with the `NeuralNetBinaryClassifier` instead of the BCELoss).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `WandbLogger` callback for logging to Weights & Biases (#607)
 - Added `None` option to `device` which leaves the device(s) unmodified (#600)
 - Add `PassthroughScoring`, a scoring callback that just calculates the average score of a metric determined at batch level and then writes it to the epoch level (#595)
+- Unify predict_proba from NeuralNetClassifier and NeuralNetBinaryClassifier and (optionally) apply a transform when using CrossEntropyLoss or BCEWithLogitsLoss (#572)
 
 ### Changed
 
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -141,6 +141,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         # https://github.com/PyCQA/pylint/issues/1085
         return super(NeuralNetClassifier, self).fit(X, y, **fit_params)
 
+    # pylint: disable=missing-docstring
     def predict_proba(self, X):
         """Where applicable, return probability estimates for
         samples.
@@ -174,7 +175,11 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         """
         # Only the docstring changed from parent.
         # pylint: disable=useless-super-delegation
-        return super().predict_proba(X)
+        y_proba = super().predict_proba(X)
+        self.check_is_fitted(attributes=['criterion_'])
+        if isinstance(self.criterion_, torch.nn.CrossEntropyLoss):
+            y_proba = torch.softmax(torch.tensor(y_proba), -1).numpy()
+        return y_proba
 
     def predict(self, X):
         """Where applicable, return class labels for samples in X.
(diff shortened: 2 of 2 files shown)


qtux:
I am not sure why Travis fails, on Python 3.6.9 tests are passing. Additionally, tests for NeuralNetBinaryClassifier are also failing, for which I did not change the code.

BenjaminBossan:
The CI is failing for unrelated reasons, see #571. 

Regarding your solution: I see where you're coming from. Since it's now the second "special" case, I wonder if we shouldn't think about a more general solution. What I have in mind right now is to have something like a `probability_postprocessor` attribute on the net that is called on the output of `predict_proba` (or on the individual batches).

One advantage is that it would give more flexibility while saving a lot of code when compared to having to duplicate almost the entire `predict_proba` method. In it's most simplest form, this solution would, however, require the user to set it the function themselves.

Maybe we could think of dispatching based on the criterion, though I'm not sure how exactly we could make that flexible.

@qtux Would such a solution work for you?

qtux:
That sounds like a good idea, I will try to think of an implementation. I would open up another PR and reference this one. Just note, that I am currently short of time, so it might take a while :).

BenjaminBossan:
Cool, that would be great.

You can also discuss ideas with us or do a draft PR if you're unsure what the best way forward would be.

BenjaminBossan:
@qtux Did you make any progress? If you need help or pointers, just ask.

qtux:
@BenjaminBossan I pushed a simpler variant of the previous commit. I think there is still a possibility to easily add a post processing step like in #580 or do your suggested loss-to-transformation dict approach.

BenjaminBossan:
Thanks @qtux I'm a bit busy right now but I'll try to have a look at this on the weekend.

qtux:
@BenjaminBossan Again sorry for the huge delay, just wanted to mention that in my opinion this pull request is in a finished state and able to replace the unfinished PR #580. My main reasons for this are:

* this PR is quite short and mostly removes duplicated code and does not introduce too much new code
* it applies the nonlinearity on the whole vector instead of in a batch-wise fashion like #580 

I think this PR provides a sensible base for implementing your proposal in #661. E.g. you could replace the following kind of code lines
```
self.check_is_fitted(attributes=['criterion_'])
if isinstance(self.criterion_, torch.nn.CrossEntropyLoss):
    y_proba = torch.softmax(torch.tensor(y_proba), -1).numpy()
```
with something like
```
y_proba = chosen_non_linearity_function(y_proba)
``´
Looking at your proposal in #661, I would close #580 and maybe consider this PR as a first step into the direction proposed in #661. In any case, both PRs are editable (was already enabled by default?), so feel free to add your changes if you like one of the approaches. I will rebase this to branch to resolve the current conflicts.

BenjaminBossan:
Hi @qtux I thought that #580 was supposed to supersede this PR, maybe I misunderstood.

For this PR to be merged, some tests would still be required.

However, I think now that #661 is a bit nicer from the user perspective and would make this PR obsolete, which is why I wouldn't go forward here. I'm sorry for the time you spent with this PR but I think it opened the door to find the right direction.

If you're interested, you could take #661, though I expect it will be much more involved than this PR or #580.

qtux:
Hi @BenjaminBossan Don't worry :), I am happy that my PRs pushed the
creation of #661 as it looks like the right way forward. Though, I think
this PR might be a good intermediate solution for all of us who are
confused by the current behavior right now.
What kind of tests are you expecting for the current PR? Maybe I can try
to add them if it is not too much of work.

Regarding taking the PR: I think I have not enough time to take #661
though I will keep an eye on it and may give some feedback if my time
allows for it.

BenjaminBossan:
@qtux Maybe I find enough time this weekend to work on #661. Then I think there is not really a point for this intermediate solution. If it turns out that #661 is going to be a longer affair, we could come back to this.

> I will keep an eye on it and may give some feedback if my time allows for it.

That's appreciated.

BenjaminBossan:
#662 has been merged, superseding this PR.

Issue #573, opened 2019-12-13
Deprecated `sklearn.metrics.scorer` in sklearn 0.22


fradav:
sklearn 0.22 + latest skorch github (and also PR `bugfix/sklearn-0.22`)
```plain
/data/miniconda3/envs/pytorch/lib/python3.6/site-packages/sklearn/utils/deprecation.py:144: FutureWarning: The sklearn.metrics.scorer module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.metrics. Anything that cannot be imported from sklearn.metrics is now part of the private API.
  warnings.warn(message, FutureWarning)
```

BenjaminBossan:
Should be solved now

PR #575, opened 2019-12-16
Move away from .scoring imports

ottonemo:
As reported in #573.

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -5,8 +5,8 @@ from contextlib import suppress
 from functools import partial
 
 import numpy as np
-from sklearn.metrics.scorer import (
-    check_scoring, _BaseScorer, make_scorer)
+from sklearn.metrics import make_scorer, check_scoring
+from sklearn.metrics.scorer import _BaseScorer
 
 from skorch.utils import data_from_dataset
 from skorch.utils import is_skorch_dataset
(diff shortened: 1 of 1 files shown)


skorch/callbacks/scoring.py
```diff
@@ -5,8 +5,8 @@
 from functools import partial
 
 import numpy as np
-from sklearn.metrics.scorer import (
-    check_scoring, _BaseScorer, make_scorer)
+from sklearn.metrics import make_scorer, check_scoring
+from sklearn.metrics.scorer import _BaseScorer
```
thomasjpfan: Most likely needs a version check to avoid the deprecation warning:

```
if version.parse(sklearn.__version__) >= version.parse('0.22.0'):
    from sklearn.metrics._scorer import _BaseScorer
else:
    from sklearn.metrics.scorer import _BaseScorer
```

BenjaminBossan:
I think we should agree on a solution for #579 first (i.e. whether to use `LooseVersion` or `packaging.version`.

ottonemo:
@BenjaminBossan  Please review again, changed to `LooseVersion`.

skorch/callbacks/scoring.py
```diff
@@ -3,10 +3,16 @@
 from contextlib import contextmanager
 from contextlib import suppress
 from functools import partial
+from distutils.version import LooseVersion
```
BenjaminBossan: move this import one line up :)

ottonemo:
@BenjaminBossan done

Issue #577, opened 2019-12-18
ModuleNotFoundError: No module named 'packaging' after installing skorch


NasGit3:
I recently installed skorch package. I installed it in a conda virtual environment using the installation guidance with conda:

```
    git clone https://github.com/skorch-dev/skorch.git
    cd skorch
    python setup.py install
```
However, when I run *import skorch* it throws me the packaging error:
   ` ModuleNotFoundError: No module named 'packaging'
`
I followed the solution suggested in this [page][2], but I still get the error. I have python 3.7.5 installed under this conda environment and I didn't get this error before installing skorch. 
 

  [1]: https://github.com/skorch-dev/skorch
  [2]: https://gist.github.com/hangtwenty/b9820fe204eebb0cc5b9aba49f3c8b22

BenjaminBossan:
Does `conda install packaging` help?

Otherwise, can you install skorch from pip? We had a very recent release, so installing from source gives you almost no additional features.

BenjaminBossan:
@NasGit3 any progress?

NasGit3:
Thank you for your reply! I used "conda install -c conda-forge skorch" after I posted this issue and it worked.  Unfortunately, I didn't get to try your solution.

BenjaminBossan:
Okay, great that you found a solution.

PR #579, opened 2019-12-28
BUG Fixes version checks

thomasjpfan:
Fixes https://github.com/skorch-dev/skorch/issues/577 by using:

```py
from distutils.version import LooseVersion
```

### DIFF
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -16,7 +16,7 @@ import sys
 from contextlib import ExitStack
 
 import numpy as np
-from packaging import version
+from distutils.version import LooseVersion
 import pytest
 from sklearn.base import clone
 from sklearn.feature_extraction.text import TfidfVectorizer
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
I wonder, is it better than adding the pretty standard `packaging` as a dependency?

thomasjpfan:
I am +0.5 on using the `distutils` since it is a part of python and it is fairly simple to use.

BenjaminBossan:
According to the answers to [this question](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python), `LooseVersion` is not compliant with the most recent PEP regarding versioning. Probably it makes no difference for us though.

I have personally no strong opinion overall. So I guess +0.5 from me as well :)

BenjaminBossan:
@ottonemo Any opinions here? Otherwise, I think we can proceed.

BenjaminBossan:
@thomasjpfan I think it's good to go.

thomasjpfan:
I agree. (I normally do not merge my own PRs lol)

BenjaminBossan:
Thx, Thomas

PR #580, opened 2020-01-06
WIP: Add transform_proba method to NeuralNet

qtux:
@BenjaminBossan I have implemented something similar to your suggestion from #572 as a WIP PR. It already reduced the amount of duplicate code, but I am not sure whether it is better to offer some default probability transformation or let the user define it freely (maybe defaulting to the identify function). Right now I am inclined to favour some sane default transformations. What do you think about it :)?

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -148,6 +148,12 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         # https://github.com/PyCQA/pylint/issues/1085
         return super(NeuralNetClassifier, self).fit(X, y, **fit_params)
 
+    def transform_proba(self, yp):
+        self.check_is_fitted(attributes=['criterion_'])
+        if isinstance(self.criterion_, torch.nn.CrossEntropyLoss):
+            return torch.softmax(yp, -1)
+        return yp
+
     def predict_proba(self, X):
         """Where applicable, return probability estimates for
         samples.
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thx for the PR. I'm a bit short on time right now but will try to have a thorough look at it on the weekend.

I think what definitely needs to be added is some documentation, i.e. docstrings that explain how and why you would override this new method, and probably also something in the normal docs.

BenjaminBossan:
So this looks pretty strong to me, I can imagine this being useful for many people. As mentioned, you should please add some docs - also, don't forget to add this as a new feature to the CHANGES.md.

I think I would also like to have a very simple unit test, where you override the method and check that it is applied when calling `predict_proba`.

One implementation detail: Right now, you call this method on each individual batch. What was the reasoning compared to calling it on the concatenated output? If there is no particular benefit, I would argue that it's better to call it on the total output, which should result in a slight performance boost (e.g. only checking `isinstance(self.criterion_, torch.nn.CrossEntropyLoss)` once per call).

In the long term, I can imagine providing a dict from loss to transformations, but this is not necessary for this PR.

Again, this is a very nice addition.

BenjaminBossan:
@qtux Any progress on this? Don't hesitate to ask questions if you have any.

BenjaminBossan:
@qtux Any updates?

qtux:
@BenjaminBossan Sorry for the long delay! After reading your comments on this PR (suggesting to not transform batch-wise), I pushed another solution to the previous PR (#572 ). I think that one is simpler and easier to expand upon.

BenjaminBossan:
Hey @qtux, thanks for the update. I believe that something like this would be good to have in skorch. For me, there are a few open points still:

1. Would it be better to run the transformation on the whole `y_proba` instead of batch-wise (think performance)?
2. There should definitely be some documentation about this, otherwise users will have a hard time discovering and correctly applying this feature.
3. I think a simple test that this is correctly applied would be nice.

If you want to get support with any of these points, make your PR writable and I can add some changes on top of yours.

BenjaminBossan:
@qtux Any updates?

BenjaminBossan:
@qtux Do you still plan on working on this? If you need some assistance, please say so.

BenjaminBossan:
#662 has been merged, superseding this PR.

PR #586, opened 2020-02-03
Added Neptune logging 

jakubczakon:
Added Neptune logger that:

- logs metrics `on_batch_end`
- logs metrics  `on_epoch_end`
- logs any additional information directly to `neptune_logger.experiment.log_whatever_neptune_allows` if used with `close_after_train=False`


BenjaminBossan:
@jakubczakon Thanks for the addition, I will have a thorough look at it soon.

What would be the easiest way for me to test this? Is the example in the docstring working or do I need to create a neptune account?

jakubczakon:
That's a fair point @BenjaminBossan 

The example in the docstrings (link to experiment) works. 
You can copy and run the code but you would need to create a (free) account.

I think a decent solution to that would be to use the public token (anonymous mode) not to force anyone to register. 
What do you think about that?

BenjaminBossan:
> I think a decent solution to that would be to use the public token (anonymous mode) not to force anyone to register.
> What do you think about that?

Yes, I think everything that reduces the friction to try it out would be helpful.

jakubczakon:
Done, now anyone can run the example without registering and see their experiments in neptune.

BenjaminBossan:
Nice, I will give this a spin soon and come back with detailed feedback. From a first glance, it looks very good though.

skorch/callbacks/logging.py
```diff
@@ -62,6 +61,135 @@ def on_epoch_end(self, net, **kwargs):
...
+    close_after_train : bool (default=True)
+      Whether to close the ``Experiment`` object once training
+      finishes. Set this parameter to False if you want to continue
+      logging to the same Experiment or if you use it as a context
+      manager.
+
+    keys_ignored : str or list of str (default=None)
+      Key or list of keys that should not be logged to
+      Neptune. Note that in addition to the keys provided by the
+      user.
```
BenjaminBossan:
```suggestion
      user, keys such as those starting with 'event_' or ending on
      '_best' are ignored by default.
```

skorch/callbacks/logging.py
```diff
@@ -62,6 +61,135 @@ def on_epoch_end(self, net, **kwargs):
...
+
+    Note
+    ----
+
+    Install psutil to monitor resource consumption
```
BenjaminBossan: Kinda redundant with the comment on line 76.

skorch/callbacks/logging.py
```diff
@@ -62,6 +61,135 @@ def on_epoch_end(self, net, **kwargs):
...
+
+    You can view example experiment logs here:
+    https://ui.neptune.ai/o/shared/org/skorch-integration/e/SKOR-4/logs
+
+    Examples
+    --------
+    >>> # Create a neptune experiment object
+    ... # We are using api token for an anonymous user.
+    ... # For your projects use the token associated with your neptune.ai account
+    >>> neptune.init(api_token='eyJhcGlfYWRkcmVzcyI6Imh0dHBzOi8vdWkubmVwdHVuZS5tbCIsImFwaV9rZXkiOiJiNzA2YmM4Zi03NmY5LTRjMmUtOTM5ZC00YmEwMzZmOTMyZTQifQ==',
```
BenjaminBossan: Maybe you could add the install instruction of neptune, as well as `import neptune` to the code example.

skorch/callbacks/logging.py
```diff
@@ -62,6 +61,135 @@ def on_epoch_end(self, net, **kwargs):
...
+        self.first_batch_ = True
+
+        keys_ignored = self.keys_ignored
+        if isinstance(keys_ignored, str):
+            keys_ignored = [keys_ignored]
+        self.keys_ignored_ = set(keys_ignored or [])
+        self.keys_ignored_.add('batches')
+        return self
+
+    def on_batch_end(self, net, **kwargs):
```
BenjaminBossan: I wonder if we really need batch level logging. Maybe logging at epoch level is sufficient? At least, I think it would make sense to allow to turn off batch level logging through a parameter.

skorch/tests/callbacks/test_logging.py
```diff
@@ -10,7 +10,143 @@
...
+            classifier_module,
+            callbacks=[neptune_logger_cls(neptune_experiment_cls())],
+            max_epochs=5,
+        )
+        net.fit(*data)
+
+    def test_fit_with_dict_input(
+            self,
+            net_cls,
+            classifier_module,
```
BenjaminBossan: argument not used (probably copied from tensorboard test that also doesn't need it). Tbh, I think this whole test can be removed here.

skorch/callbacks/logging.py
jakubczakon: I often find having that valuable but I agree there should be an option to turn it off.
Added it.

skorch/tests/callbacks/test_logging.py
jakubczakon: dropped it

jakubczakon:
I see that I messed up the formatting in some places with my local settings.
That sais, it passes pylint (with warnings)
Should I fix those, or they are not important to you?

jakubczakon:
I've added this test @BenjaminBossan, 
Interestingly this test passes locally, calling my `.log_metric 130 times, yet on github it calls it only 120 times.
I will investigate but seems unexpected.

BenjaminBossan:
@jakubczakon I tried to understand the numbers. For this, I turned off the internal validation, as it makes the whole calculation more difficult (by adding `train_split=False` as argument to the net).

Now we have the following situation:

* on batch level, we have 2 keys x (40 / 4) batches x 5 epochs = 100 calls
* on epoch level, we have 2 keys x 5 epochs = 10 calls

That is 110 in sum, which is exactly what I find (please add this explanation to the comment for future reference).

I think it would also make sense to copy the same test, but this time with `log_on_batch_end=False`. Then we should expect 10 calls.

Why you get different results locally, I don't know (for me, it's the same as in the CI). My first guess would be some kind of version mismatch.

jakubczakon:
Thank you @BenjaminBossan!

I've added the suggestions and it fixed the issue.

BenjaminBossan:
Thanks @jakubczakon, this looks very good to me now.

I will wait a few days to see if @ottonemo or @thomasjpfan have any comments on this, otherwise I will merge it.

jakubczakon:
Awesome, thanks!

skorch/callbacks/logging.py
```diff
@@ -62,6 +61,142 @@ def on_epoch_end(self, net, **kwargs):
...
+            close_after_train=True,
+            keys_ignored=None,
+    ):
+        self.experiment = experiment
+        self.log_on_batch_end = log_on_batch_end
+        self.close_after_train = close_after_train
+        self.keys_ignored = keys_ignored
+
+    def initialize(self):
+        self.first_batch_ = True
```
thomasjpfan: Is `first_batch_` used?

skorch/tests/callbacks/test_logging.py
```diff
@@ -3,14 +3,165 @@
...
+    ):
+        # ignore 'dur' and 'valid_loss', 'unknown' doesn't exist but
+        # this should not cause a problem
+        npt = neptune_logger_cls(
+            mock_experiment, keys_ignored=['dur', 'valid_loss', 'unknown'])
+        net_cls(
+            classifier_module,
+            callbacks=[npt],
+            max_epochs=3,
+        ).fit(*data)
```
thomasjpfan: Can we assert how many times `log_metric` should be called in this case?

skorch/callbacks/logging.py
BenjaminBossan: I think this is for consistency with the `TensorBoard` callback. It is convenient to have so that you can, e.g., log an image of the network graph exactly once. You may not be able to use `on_train_begin` for this because that one gets the input `X`, not the one that is returned by the data loader.

skorch/tests/callbacks/test_logging.py
jakubczakon: Sure can do.

skorch/callbacks/logging.py
jakubczakon: Yeah, I simply copied it from TensorBoard (to be honest I haven't thought about it much).

Also, if I were to use it properly I should have `self.first_batch_ = False` on `on_batch_end` which is missing.

I can either drop it from `initialize` or add `self.first_batch_ = False` to `on_batch_end`.
Which one would you prefer?

jakubczakon: I've added `self.first_batch_ = False` to `on_batch_end` but I can easily drop it from both as they are not used (to my understanding)

What do you think?

BenjaminBossan: The main reason I wanted to have it for TensorBoard was to be able to trace and add a graph of the network to TensorBoard. I think that option doesn't exist for neptune, does it? However, I think consistency is also nice, so I would leave it there.

thomasjpfan: Let's document the attribute in the docstring and add a quick test for `first_batch_`?

BenjaminBossan: Yes, good idea.

jakubczakon:
Done @thomasjpfan

jakubczakon:
Awesome!
Thank you @BenjaminBossan  @thomasjpfan !

PR #590, opened 2020-02-12
Fix a bug with custom attribute name clashes

BenjaminBossan:
When a user adds a new "settable" attribute (i.e. that works with
`set_params`) whose names starts the same as an existing attribute (say
"optimizer_2"), and adds a corresponding argument (say
"optimizer_2__lr"), skorch will erroneously complain about this
argument because it thinks it belongs to the other
attribute ("optimizer" in this example). The added unit test should
illustrate this behavior.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1334,7 +1334,7 @@ class NeuralNet:
         for key in kwargs:
             if key.endswith('_'):
                 continue
-            for prefix in self.prefixes_:
+            for prefix in sorted(self.prefixes_, key=lambda s: (-len(s), s)):
                 if key.startswith(prefix):
                     if not key.startswith(prefix + '__'):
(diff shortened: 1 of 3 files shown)


skorch/net.py
```diff
@@ -1334,7 +1334,7 @@ def _check_kwargs(self, kwargs):
         for key in kwargs:
             if key.endswith('_'):
                 continue
-            for prefix in self.prefixes_:
+            for prefix in sorted(self.prefixes_, key=lambda s: (-len(s), s)):
```
thomasjpfan: For future maintainability, add a comment about why this sorting is needed or at least a link to this issue. (It is not obvious at first glance why this is needed)

BenjaminBossan: I added a comment

PR #592, opened 2020-02-14
Update dynamic input FAQ with callback solution

ottonemo:
From https://stackoverflow.com/a/60170023/1643939


BenjaminBossan:
Just a minor thing, can we think of a better name than `Resize`? By that name alone, I wouldn't be able to guess what it means and I can imagine this code will be copied by a few people.

PR #595, opened 2020-02-21
add PassthroughScoring

BenjaminBossan:
As mentioned [here](https://github.com/skorch-dev/skorch/pull/593#issuecomment-586739727), I factored out the addition of `PassthroughScoring` from the PR #593 

Here is a copy of the reasoning:

----------

Currently, when adding new losses on a batch level, it's very fiddly to log them on an epoch level. You have to define a custom function that picks the right value from history and passes it through without modification, then add a BatchScoring callback that uses this mock scoring function, defines the name of the score again, and pass noop.

Here is the code before and after:

```python
# somewhere in the net code
    ...
    self.history.record_batch('my_loss', my_loss)
    ...

# before
from skorch.callbacks import BatchScoring
from skorch.utils import noop

def my_loss_score(net, X=None, y=None):
    return net.history[-1, 'batches', -1, 'my_loss']

callbacks = [
    BatchScoring(
        my_loss_score,
        name='my_loss',
        target_extractor=noop,
    ),
]

# after
from skorch.callbacks import PassthroughScoring

callbacks = [
    PassthroughScoring(name='my_loss'),
]
```
----------

I replaced `BatchScoring` with `PassthroughScoring` in the existing code for train and valid loss. Even though `skorch.utils.train_loss_score` and `skorch.utils.valid_loss_score` would not be needed anymore with this change, I still left them in place. This is for 2 reasons:

* someone might use them directly (after all, they're public functions)
* if they are removed, old pickled nets cannot be unpickled anymore

The latter could be fixed by introducing a special case in `__getstate__` and replacing the old `BatchScoring` with the new `PassthroughScoring` which doesn't require these functions, but this seems to be far too much work for very little gain. Instead, I propose to leave those functions for now, maybe deprecate them at some future time.

I'm not quite certain what to do with `test_net_initialized_with_initalized_dataset`. I'll think of something, but if anyone has an idea, I'll be glad to hear it.

### DIFF
--- a/docs/user/callbacks.rst
+++ b/docs/user/callbacks.rst
@@ -142,14 +142,14 @@ compare the performance once with and once without the callback.
 Scoring
 -------
 
-skorch provides two scoring callbacks by default,
+skorch provides two callbacks that calculate scores by default,
 :class:`.EpochScoring` and :class:`.BatchScoring`. They work basically
 in the same way, except that :class:`.EpochScoring` calculates scores
 after each epoch and :class:`.BatchScoring` after each batch. Use the
 former if averaging of batch-wise scores is imprecise (say for AUC
 score) and the latter if you are very tight for memory.
 
-In general, the scoring callbacks are useful when the default scores
+In general, these scoring callbacks are useful when the default scores
 determined by the :class:`.NeuralNet` are not enough. They allow you
 to easily add new metrics to be logged during training. For an example
 of how to add a new score to your model, look `at this notebook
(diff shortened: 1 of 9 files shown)


docs/user/callbacks.rst
```diff
@@ -199,6 +199,26 @@ it is passed to the scoring function. The main reason why we need this
 is that sometimes, the target is not of a form expected by sklearn and
 we need to process it before passing it on.
 
+On top of the two described scoring callbacks, skorch also provides
+:class:`.PassthroughScoring`. This callback does not actually calculate
+any new scores. Instead it uses an existing score that is calculated
+for each batch and determines the average of this score, which is
```
ottonemo, line 205:
```suggestion
for each batch (the train loss, for example) and determines the average of this score, which is
```

Maybe this makes it easier to relate how and when this callback is useful?

skorch/tests/test_net.py
```diff
@@ -1465,22 +1464,15 @@ def __init__(self, *args, foo, **kwargs):
...
-            assert side_effect == [123, 123]
-        else:
-            # train/valid split
-            assert side_effect == [123]
-
+    @pytest.mark.skip
     @pytest.mark.xfail
     def test_net_initialized_with_initalized_dataset(
             self, net_cls, module_cls, data, dataset_cls):
+        # TODO: What to do of this test now??
```
ottonemo, line 1475: Do we need an issue to track this?

BenjaminBossan: tbh I don't remember why I put this here but the test xfails with the expected error, so I removed the comment and skip

docs/user/callbacks.rst
BenjaminBossan: fixed

PR #596, opened 2020-02-22
Make get_params_for and get_params_for_optimizer public

BenjaminBossan:
Private methods are kept for compatibility. Functionally, nothing was
changed.

This is part of the effort discussed in #593 to make skorch more easily
customizable. There, we agreed to split the individual parts of that effort
into their own pull requests.

The main reason for this change is to give users confidence in, and better
documentation for, implementing their own `initialize_` methods. Those
methods require the use of `get_params_for`, which until now was
discouraged since it was private.

One example where this is useful is if the user wants to implement a
net that has more than one optimizer. Now they can override
`initialize_optimizer` and use `get_params_for_optimizer('opt_1', ...)`
to initialize all their optimizers.


skorch/net.py
```diff
@@ -1251,19 +1251,48 @@ def get_iterator(self, dataset, training=False):
...
+        """Collect and return init parameters for an attribute.
 
-            optimizer__lr=0.05,
-            optimizer__param_groups=[
-                ('rnn*.period', {'lr': 0.3, 'momentum': 0}),
-                ('rnn0', {'lr': 0.1}),
-            ]
+        Attributes could be, for instance, pytorch modules, criteria,
+        or data loaders (for optimizers, use
+        :meth:`.get_params_for_optimizer` instead). Use the returned
```
thomasjpfan: There is still a use case for using `get_params_for` with an optimizer as shown in the change in `examples/translation/translation.ipynb` in this PR.

BenjaminBossan: The change I made there was just a simple search and replace. Now, after actually looking at this, I believe it should use `_get_params_for_optmizer` instead. Or is there any reason not to?

BenjaminBossan:
> Maybe in the future we can have a "Developers guide"?

Yes, good idea. I guess at the moment, the kind of information that could be in such a guide is sprinkled throughout the docs.

> We are a mixture of using subclasses and callbacks for customization.

It's true and it might not always be obvious what the best approach is. In general though, it's probably okay (though not ideal) if there is more than one way to get to the goal.

skorch/net.py
thomasjpfan: Yea it should use `get_params_for_optimizer` just in case someone uses glob matching for `param_groups`?

BenjaminBossan: I changed the code to use `get_params_for_optimizer`.

BenjaminBossan:
@ottonemo I noticed that `_get_params_for_optimizer` returns `[pgroups], kwargs` and then we call `optimizer(*pgroups, **kwargs)`. AFAICT, `pgroups` will alwasy be a one element list (with the element being a list). Therefore, could we not change the code to `return pgroups, kwargs` and `optimizer(pgroups, **kwargs)`? I think now is the chance to make the change.

(We could, in theory, only make the change for `get_params_for_optimizer` to not break `_get_params_for_optimizer`.)

Also, looking at the code from [`torch.optim.Optimizer`](https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html), `params` (the first argument) is expected to be a list, so even if `pgroups` had several elements, it would not work, since it would consist of a list of lists.

BenjaminBossan:
ping @ottonemo

BenjaminBossan:
> I noticed that `_get_params_for_optimizer` returns `[pgroups], kwargs` and then we call `optimizer(*pgroups, **kwargs)`. AFAICT, `pgroups` will alwasy be a one element list (with the element being a list). Therefore, could we not change the code to `return pgroups, kwargs` and `optimizer(pgroups, **kwargs)`? I think now is the chance to make the change.

@ottonemo Did you have a look at this? Do we want to change this?

ottonemo:
@BenjaminBossan Sorry, I overread your comment.

The naming is definitely misleading and it hides the original intent of the return values. In my opinion we should *at least* change the documentation of `get_params_for_optimizer` to read

        Returns
        -------
        args : list
          List of positional parameters.

The idea at the time was to return (args, kwargs) in case the requirements for positional parameters of `optimizer` change (e.g., parameter `default` needs to be supplied as well or there is an additional parameter). We could also just drop the positional parameter and merge everything in `kwargs`, which would be consistent with `get_params`. What do you think?

BenjaminBossan:
@ottonemo That makes sense. I think just renaming the argument to make it explicit that those should be the `*args` (and changing them to tuple) would convey the intent much better, without actually changing the logic of the code. And since now there are strict positional arguments, I guess it could happen that it might be required one day. So I'm in favor of leaving the signature as is but changing the argument name.

PR #597, opened 2020-02-24
Facilitate adding custom modules, criteria, or optimizers

BenjaminBossan:
This is the third part of the change discussed in #593. For the docs,
it assumes that the changes in #596 have been merged.

### Changes

Now, when creating a custom net that has additional modules, criteria,
or optimizers, skorch will take care of registering them inside
`prefixes_` and `cuda_dependent_attributes_`. If this non-obvious
change is not done, it can result in errors down the line.

Additionally, during `set_params`, there is now re-initialization of
those components if any of the special keys (i.e. those that start
with one of the `prefixes_`) has "module", "criterion", or "optimizer"
as a substring.

This last part is not a perfect solution, but I couldn't find a better
way to determine whether a component needs to be reinitialized. Say we
add a new module called "mymodule" to the custom net. How can we know
if calling `net.set_params(mymodule__foo=123)` should result in module
re-initialization? We can't really, except by trying to guess from the
name. Other ideas that I thought of but rejected:

1. Check if `self.module` is a `nn.Module` subclass. But what if it's
not a module subclass but a function that returns an instantiated
module?

2. Instantiate `self.mymodule()` and see if we get a module back. But
what if the module requires arguments in order to be called
successfully? Okay, we could find those arguments and then
initialize. But do we really want to make a "test initialization"? If
feel like this could have unintended side-effects. Also, this would not
handle any custom code a user might add to `initialize_module`.

If anybody can propose a better idea, that would be welcome. My
personal belief is that the substring matching is not that bad in the
sense that it's very unlikely to trigger a false positive. However, it
might not be easy to discover for a user (I added some documenation
and plan on adding the DCGAN example that shows this in practice).

### Additional changes

1.  Refactor tests that used patched score method

The score method used to be patched to give fixed values for a net
fixture used variously in the tests. This was so that certain scores
could be asserted. However, this had the side effect that some tests
that actually want to calculate the real score had mysterious return
values from score().

Instead of fixing those tests, it's better not to patch the score
method in the fixture and to instead patch it in those tests where
it's necessary. This helps making those tests more explicit (which is
better than implicit).

I stumbled upon this accidentally when I tried to run some tests
in isolation.

2. Fix typo: explicitely -> explicitly

### TODO

In one of the tests, I use `self._get_params_for`, which should be
changed to `self.get_params_for` as soon as #596 has been merged.


skorch/net.py
```diff
@@ -1524,6 +1534,90 @@ def __setstate__(self, state):
...
+        # just setattr as usual
+        is_known = name.endswith('_') or (name in self.prefixes_)
+        is_special_param = '__' in name
+        is_torch_mod_or_opt = any(c in name for c in PYTORCH_COMPONENTS)
+        if is_known or is_special_param or not is_torch_mod_or_opt:
+            super().__setattr__(name, attr)
+            return
+
+        self._register_attribute(name)
+        super().__setattr__(name, attr)
```
ottonemo, line 1614: Wouldn't it be more straight forward to reduce this to

1. register if necessary
2. set attribute

instead of having (check, set) and (register, set)?

Suggestion:

```suggestion
if is_torch_mod_or_opt or not (is_known and is_special_param):
    self._register_attribute(name)
    
super().__setattr__(name, attr)
```

skorch/net.py
```diff
@@ -1524,6 +1534,90 @@ def __setstate__(self, state):
...
         self.__dict__.update(state)
 
+    def _register_attribute(
+            self,
+            name,
+            prefixes=True,
+            cuda_dependent_attributes=True,
+    ):
+        """Add attribute name to prefixes_ and
+        cuda_dependent_attributes_.
```
ottonemo, line 1544: This should include the purpose of registering an attribute.

skorch/net.py
```diff
@@ -1524,6 +1534,90 @@ def __setstate__(self, state):
...
+                self.cuda_dependent_attributes_[:] + [name + '_'])
+
+    def _unregister_attribute(
+            self,
+            name,
+            prefixes=True,
+            cuda_dependent_attributes=True,
+    ):
+        """Remove attribute name from prefixes_ and
+        cuda_dependent_attributes_.
```
ottonemo, line 1572: This should include the consequence of un-registering an attribute.

docs/user/neuralnet.rst
```diff
@@ -467,3 +468,79 @@ total loss:
...
+so:
+
+1. Set this attribute inside the corresponding method. E.g., when
+   setting an optimizer, use :meth:`.initialize_optimizer` for that.
+2. Inside the initialization method, use :meth:`.get_params_for` (or,
+   if dealing with an optimizer, :meth:`.get_params_for_optimizer`) to
+   retrieve the arguments for the constructor.
+3. The attribute name should contain the substring ``"module"`` if
+   it's a module, ``"criterion"`` if a criterion, and ``"optimizer"``
+   if an optimizer. This so so that skorch knows if a change in
```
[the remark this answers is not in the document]
BenjaminBossan: done

skorch/net.py
```diff
@@ -1524,6 +1534,90 @@ def __setstate__(self, state):
...
+
+        """
+        # if it's a
+        # 1. known attribute or
+        # 2. special param like module__num_units or
+        # 3. not a torch module/optimizer instance or class
+        # just setattr as usual
+        is_known = name.endswith('_') or (name in self.prefixes_)
+        is_special_param = '__' in name
+        is_torch_mod_or_opt = any(c in name for c in PYTORCH_COMPONENTS)
```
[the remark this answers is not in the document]
BenjaminBossan: Streamlined the code (though a bit different than proposed)

skorch/net.py
BenjaminBossan: Streamlined the code (though a bit different than proposed)

skorch/net.py
BenjaminBossan: done

skorch/net.py
BenjaminBossan: done

skorch/tests/test_net.py
```diff
@@ -2326,6 +2326,138 @@ def train_step(self, Xi, yi, **fit_params):
...
+        # set, which is what is tested here
+        class MyNet(net_cls):
+            def __init__(self, *args, mymodule=module_cls, **kwargs):
+                self.mymodule = mymodule
+                super().__init__(*args, **kwargs)
+
+            def initialize_module(self, *args, **kwargs):
+                super().initialize_module(*args, **kwargs)
+
+                params = self._get_params_for('mymodule')
```
ottonemo, line 2448:
```suggestion
                params = self.get_params_for('mymodule')
```

skorch/net.py
```diff
@@ -1599,6 +1609,100 @@ def __setstate__(self, state):
 
         self.__dict__.update(state)
 
+    def _register_attribute(
+            self,
+            name,
+            prefixes=True,
```
thomasjpfan, line 1615: Do we call `_register_attribute` with `prefixes=False`?

We also always call this with `cuda_dependent_attributes=True`. I guess there are no torch components that are not also `cuda_dependent_attributes`?

skorch/net.py
```diff
@@ -34,6 +34,16 @@
 from skorch.utils import to_tensor
 
 
+PYTORCH_COMPONENTS = {'criterion', 'module', 'optimizer'}
```
thomasjpfan, line 37: Private for now?

```suggestion
_PYTORCH_COMPONENTS = {'criterion', 'module', 'optimizer'}
```

skorch/net.py
BenjaminBossan: Yes, it could be that this is one of these cases of YAGNI. My idea was to provide a more general "setter" for prefixes and cuda dependent attributes that could be used in other places as well, where it might not make sense to change both at the same time.

skorch/net.py
```diff
@@ -1599,6 +1609,100 @@ def __setstate__(self, state):
...
+        # If it's a
+        # 1. known attribute or
+        # 2. special param like module__num_units or
+        # 3. not a torch module/optimizer instance or class
+        # just setattr as usual.
+        # For a discussion why we chose this implementation, see here:
+        # https://github.com/skorch-dev/skorch/pull/597
+        is_known = name.endswith('_') or (name in self.prefixes_)
+        is_special_param = '__' in name
+        is_torch_component = any(c in name for c in PYTORCH_COMPONENTS)
```
BenjaminBossan, line 1695:
```suggestion
        is_torch_component = any(c in name for c in _PYTORCH_COMPONENTS)
```

skorch/net.py
BenjaminBossan: Done. We can make it public later if there is a need.

skorch/net.py
```diff
@@ -1599,6 +1609,100 @@ def __setstate__(self, state):
...
+          Whether to remove from cuda_dependent_attributes_.
+
+        """
+        # copy the lists to avoid mutation
+        if prefixes:
+            self.prefixes_ = [p for p in self.prefixes_[:] if p != name]
+
+        if cuda_dependent_attributes:
+            self.cuda_dependent_attributes_ = [
+                a for a in self.cuda_dependent_attributes_ if a != name + '_']
```
thomasjpfan, line 1676: hm

```suggestion
                a for a in self.cuda_dependent_attributes_[:] if a != name + '_']
```

skorch/net.py
```diff
@@ -1599,6 +1609,100 @@ def __setstate__(self, state):
...
+
+        Use this to remove PyTorch components that are not needed
+        anymore. This is mostly a clean up job, so as to not leave
+        unnecessary prefixes or cuda-dependent attributes.
+
+        This method takes care of not mutating the lists.
+
+        Parameters
+        ----------
+        prefixes : bool (default=True)
```
thomasjpfan, line 1663: I do not see a use case where one would set `prefixes=False` or `cuda_dependent_attributes=False` for `_unregister_attribute`.

Even if `_register_attribute` was called with `prefixes=False`, `_unregister_attribute` would do the right thing with `prefixes` set either way.

skorch/net.py
BenjaminBossan: I don't think we need to copy here since we create a new list anyway, right?

skorch/net.py
BenjaminBossan: The only use would be a microscopic gain in speed I guess. Overall, I believe that we should keep it consistent between register and unregister, so either both have the option or neither. I don't have a strong opinion on that.

skorch/net.py
thomasjpfan: Yea it's okay here.

skorch/net.py
```diff
@@ -1599,6 +1609,100 @@ def __setstate__(self, state):
...
+          Whether to remove from prefixes_.
+
+        cuda_dependent_attributes : bool (default=True)
+          Whether to remove from cuda_dependent_attributes_.
+
+        """
+        # copy the lists to avoid mutation
+        if prefixes:
+            self.prefixes_ = [p for p in self.prefixes_[:] if p != name]
```
thomasjpfan, line 1672: We do not need it here either?

```suggestion
            self.prefixes_ = [p for p in self.prefixes_ if p != name]
```

skorch/net.py
thomasjpfan: Okay let's keep it

skorch/net.py
BenjaminBossan: Yes, true. Fixed

Issue #610, opened 2020-03-30
Cosine annealing called every epoch instead of every batch


guybuk:
Cosine annealing LR should update the learning rate every batch according to: https://arxiv.org/pdf/1608.03983.pdf

It does not, seeing as the method `skorch.LRScheduler.on_batch_end` is: 
```python
    def on_batch_end(self, net, training, **kwargs):
        if (
                training and
                hasattr(self.lr_scheduler_, 'batch_step') and
                callable(self.lr_scheduler_.batch_step)
        ):
            self.lr_scheduler_.batch_step(self.batch_idx_)

        if TorchCyclicLR and isinstance(self.lr_scheduler_, TorchCyclicLR):
            self.lr_scheduler_.step(self.batch_idx_)

        if training:
            self.batch_idx_ += 1
```
`torch.optim.lr_scheduler.CosineAnnealingLR` is neither `TorchCyclicLR` nor has a `batch_step` call (On that note, I haven't found a single LRScheduler that does have a `batch_step` method).

And on a somewhat unrelated note, it would be ideal for me to log somehow the LR per batch (or epoch, depending on the scheduler). Would it be wise to add a `net.record(...)` / `net.record_batch` call to `skorch.LRScheduler`?

BenjaminBossan:
Thanks for reporting.

I'm not an expert on learning rate schedulers, so please correct me if I'm wrong. From my reading of things, the `CosineAnnealingLR` in pytorch is intended to work on an epoch level. They write:

> Set the learning rate of each parameter group using a cosine annealing schedule, where η\_max​ is set to the initial lr and T_cur​ is the number of epochs since the last restart in SGDR:

[docs](https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.CosineAnnealingLR)

For me, it reads like this is the intended behavior. I cannot judge if this is the correct behavior according to the paper, but if it isn't, and if my interpretation is correct, you would need to open an issue in pytorch and report this error. I wouldn't go so far as to try to correct a wrong implementation in pytorch on the skorch side.

But again, I might be reading things wrong here, so please correct me if that's the case.

[...]

PR #613, opened 2020-04-08
Optionally record LR steps in LRScheduler

guybuk:
- The implementation is inspired by how `Checkpoint` handles its own event. By default, has an `event_name=event_lr`, and can be set to None to avoid recording the event at all.

- Added two tests, for schedulers that update every epoch and for schedulers that update every batch. After running all the tests, it seems that nothing has broken. (Though there were a few skipped/failed tests to start with, in unrelated sections of code)

- A short documentation of the new parameter was added to the constructor's docstring

- Removed the `batch_step` code (https://github.com/skorch-dev/skorch/issues/610#issuecomment-609460462)

- Fixed a minor bug where the CyclicLR Scheduler (As well as any other potential batch-step scheduler) would take a step regardless of whether or not it was in `training` mode, so it would actually take steps during validation.


guybuk:
Looks like it fails because previous Pytorch versions don't have `get_last_lr()`? What should be done in this case?

BenjaminBossan:
> Looks like it fails because previous Pytorch versions don't have `get_last_lr()`? What should be done in this case?

Oh, that's too bad. I guess there is no alternative way to access that information that would be available in pytorch < 1.4?

If not, the best we can do is probably to catch the error and just don't do anything in older pytorch versions. The docstring would need to be adjusted accordingly.

The affected tests would need to be skipped using `pytest.mark.skipif`.

guybuk:
In earlier versions, there's still a `get_lr()` method. However, its implementation is very weird, for example: 
```python
class ExponentialLR(_LRScheduler):
...
    def get_lr(self):
        if self.last_epoch == 0:
            return self.base_lrs
        return [group['lr'] * self.gamma
                for group in self.optimizer.param_groups]
```
For the first epoch, it prints the current learning rate, for any other epoch, it'll print the next learning rate. So if my schedule is `[100,90,80,70]` then calling `get_lr()` every epoch results in `[100,80,70,...]`

This behavior is consistent across most schedulers I saw.

Any suggestions? I'm struggling to see how to make this work for an earlier version.

BenjaminBossan:
Hmm, interesting, that looks like a bug. I agree that we shouldn't support a behavior that is most likely buggy (but only for some pytorch version). I can think of two ways to proceed from here:

1. Don't support the feature at all for pytorch 1.4.
2. Implement a skorch function that calls `get_last_lr` if pytorch >= 1.4 and uses a custom logic similar to `get_lr` but without the inconsistency (should return the same results for all supported pytorch versions).

I don't have a personal preference, since I rarely use learning rate schedulers. 1. is of course the quick and easy solution while 2. is nicer for the user. In fact, you could also start with 1. and you/someone could add 2 later.

skorch/callbacks/lr_scheduler.py
```diff
@@ -53,14 +52,20 @@ class LRScheduler(Callback):
...
+      Pass ``None`` to disable placing events in history.
+      **Note:** This feature works only for pytorch version >=1.4
+
     kwargs
       Additional arguments passed to the lr scheduler.
 
     """
 
-    def __init__(self, policy='WarmRestartLR', monitor='train_loss', **kwargs):
+    def __init__(self, policy='WarmRestartLR', monitor='train_loss', event_name="event_lr", **kwargs):
```
BenjaminBossan, line 65: Line is too long (>88), could you please break it (1 argument per line)?

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -186,6 +186,40 @@ def test_lr_scheduler_set_params(self, classifier_module, classifier_data):
...
+        scheduler = LRScheduler(TorchCyclicLR, base_lr=1, max_lr=5, step_size_up=4)
+        net = NeuralNetClassifier(
+            classifier_module,
+            max_epochs=1,
+            lr=123.,
+            batch_size=batch_size,
+            callbacks=[('scheduler', scheduler)]
+        )
+        net.fit(X, y)
+        new_lrs = scheduler.simulate(net.history[-1, 'train_batch_count'], initial_lr=123.)
```
BenjaminBossan, line 219: line too long

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -186,6 +186,40 @@ def test_lr_scheduler_set_params(self, classifier_module, classifier_data):
         net.fit(*classifier_data)  # we need to trigger on_train_begin
         assert net.callbacks[0][1].lr_scheduler_.base_lrs[0] == 456
 
+    @pytest.mark.parametrize('policy,kwargs', [
+        (StepLR, {'gamma': 0.9, 'step_size': 1})
+    ])
+    def test_lr_scheduler_record_epoch_step(self, classifier_module, classifier_data, policy, kwargs):
```
BenjaminBossan, line 192: line too long

skorch/callbacks/lr_scheduler.py
```diff
@@ -12,6 +12,7 @@
 from torch.optim.lr_scheduler import MultiStepLR
 from torch.optim.lr_scheduler import ReduceLROnPlateau
 from torch.optim.lr_scheduler import StepLR
+
```
BenjaminBossan, line 15: empty line seems unnecessary to me

BenjaminBossan:
@guybuk Thanks for addressing the concerns. Just a nitpick: Could you please use hanging indents where you broke the lines? This is consistent with the majority of the skorch code base.

@thomasjpfan Since you are more experienced with lr schedulers, could you please also take a look?

guybuk:
Sure. Are there formatting specifications somewhere I missed? I don't actually pay attention to my formatting and I use the formatting option in Pycharm without even noticing...

Also, kind of awkward but what exactly are hanging indents?
```python
            if self.event_name is not None and hasattr(self.lr_scheduler_,
                                                       "get_last_lr"):
```
Should become
```python
            if self.event_name is not None \
                    and hasattr(self.lr_scheduler_, "get_last_lr"):
```
?

Or is it 
```python
        new_lrs = scheduler.simulate(net.history[-1, 'train_batch_count'],
                                     initial_lr=123.)
```
Should become
```python
        new_lrs = scheduler.simulate(
            net.history[-1, 'train_batch_count'],
            initial_lr=123.
        )
```

BenjaminBossan:
This would be good:

```python
            if self.event_name is not None and hasattr(
                self.lr_scheduler_, 'get_last_lr'):
```

> Should become
> ```python
>         new_lrs = scheduler.simulate(
>             net.history[-1, 'train_batch_count'],
>             initial_lr=123.
>         )
> ```

Yes, that's good, ideally with a trailing comma on the third line.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -186,6 +186,47 @@ def test_lr_scheduler_set_params(self, classifier_module, classifier_data):
...
+        scheduler = LRScheduler(policy, **kwargs)
+        lrs = scheduler.simulate(epochs, initial_lr=123.)
+        net = NeuralNetClassifier(
+            classifier_module,
+            max_epochs=epochs,
+            lr=123,
+            callbacks=[('scheduler', scheduler)]
+        )
+        net.fit(*classifier_data)
+        if hasattr(scheduler.lr_scheduler_, "get_last_lr"):
```
thomasjpfan, line 207: As stated by @BenjaminBossan , I would prefer a `@pytest.mark.skip` here based on `torch.__version__` then checking for the method here.

BenjaminBossan: ping @guybuk

guybuk: Never actually used `pytest` before. Is this what you mean?
```python
@pytest.mark.skipif(version.parse(torch.__version__)<version.parse("1.4.0")
def test...
```

thomasjpfan:
```py
from distutils.version import LooseVersion

@pytest.mark.skipif(LooseVersion(torch.__version__) < '1.4')
def test...
```

BenjaminBossan:
Thanks a lot

PR #607, opened 2020-03-10
Add logger to Weights & Biases

borisdayma:
Add support for logging through Weights & Biases.

`WandbLogger` automatically logs metrics, model topology, gradients and best trained model.

Sample run: https://app.wandb.ai/borisd13/skorch/runs/1vs6725x?workspace=user-borisd13
You can switch tabs on the left and see computer resources, model graph, logged files, etc

PR #616, opened 2020-04-11
[WIP] Update README and enviroment

BenjaminBossan:
* conda install of requirements-dev.txt currently fails because of a
  missing dependency, use pip instead
* update environment.yml, especially now containing pytorch, since
  there is now a unified installation method

Further sections of the README and possibly docs need to be updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -192,7 +192,6 @@ If you want to help developing, run:
     cd skorch
     conda env create
     source activate skorch
-    pip install -r requirements-dev.txt
     pip install -e .
 
     py.test  # unit tests
--- a/README.rst
+++ b/README.rst
@@ -187,7 +187,7 @@ If you want to help developing, run:
     conda env create
     source activate skorch
     # install pytorch version for your system (see below)
-    conda install -c conda-forge --file requirements-dev.txt
+    pip install -r requirements-dev.txt
     python setup.py develop
 
     py.test  # unit tests
@@ -205,7 +205,7 @@ If you just want to use skorch, use:
     # create and activate a virtual environment
     pip install -r requirements.txt
     # install pytorch version for your system (see below)
-    python setup.py install
+    pip install .
 
 If you want to help developing, run:
 
@@ -217,7 +217,7 @@ If you want to help developing, run:
     pip install -r requirements.txt
     # install pytorch version for your system (see below)
     pip install -r requirements-dev.txt
-    python setup.py develop
+    pip install -e .
 
     py.test  # unit tests
     pylint skorch  # static code checks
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -186,17 +192,16 @@ If you want to help developing, run:
     cd skorch
     conda env create
     source activate skorch
-    # install pytorch version for your system (see below)
-    conda install -c conda-forge --file requirements-dev.txt
-    python setup.py develop
+    pip install -r requirements-dev.txt
```
thomasjpfan, line 195: Should we have a `environmental-dev.yml` for using conda for development?

BenjaminBossan: We could consider this, the only problem is that one of the requirements is not available through conda.

https://github.com/skorch-dev/skorch/pull/607#issuecomment-612365883

thomasjpfan: Its possible to include [pip only packages on the yaml](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually).

Here is an [concrete example that dask-ml uses](https://github.com/dask/dask-ml/blob/5d24b72586b03d6ffe5080f4031c63a5b85954b9/ci/environment-docs.yaml#L46-L49).

BenjaminBossan: Now I get it, you would like to have a conda environment file that contains all the dependencies, including for development (not distinct ones like `requirements.txt` and `requirements-dev.txt` whose libarries don't overlap). This could certainly be done.

The question is, do we even need a non-development `environment.yml` or is that only interesting for developers anyway?

thomasjpfan: > The question is, do we even need a non-development environment.yml or is that only interesting for developers anyway?

I agree, `environment.yml` seems only interesting for developers.

As for this PR, using pip to install `requirements-dev.txt` is nicer from a maintenance point of view. (We would not need to keep a `environment.yml` in sycn with the `requirements-dev.txt`.

BenjaminBossan: :+1:

BenjaminBossan:
If someone could please test with CUDA, that would be great.

docs/user/installation.rst
```diff
@@ -79,24 +76,30 @@ If you want to help developing, run:
...
     pylint skorch  # static code checks
 
 pytorch
 ~~~~~~~
 
-PyTorch is not covered by the dependencies, since the PyTorch
-version you need is dependent on your system. For installation
-instructions for PyTorch, visit the `pytorch website
-<http://pytorch.org/>`__.
+PyTorch is not covered by the dependencies, since the PyTorch version
```
ottonemo, line 87: That's not true anymore as the `environment.yml` includes pytorch now?

BenjaminBossan: I think it's valid this way. Most people will install skorch through pip and `pip install skorch` will not automatically install pytorch. In the section `From source`, it only says to install pytorch in the subsection `Using pip`, not in `Using conda`. For these reasons, I would leave this part as is.

PR #620, opened 2020-04-19
CVSplit raise warning when random_state is not used

### LINKED ISSUE #619: CVSplit should raise error when random_state is not used
When CVSplit is created like this:

```python
CVSplit(5, random_state=123)
```

some devs (including me! :wink:) might think, that `random_state` will cause data to be shuffled.
But it isn't.

I suggest to throw error when `random_state` is not used for given `CVSplit` (i.e. when `cv` parameter is not `float`).

It is consistent with scikit-learn:

```python
from sklearn.model_selection import KFold
a = KFold(5, shuffle=False, random_state=123)
```

gives a `FutureWarning`:

```
.../lib/python3.7/site-packages/sklearn/model_selection/_split.py:296:
FutureWarning: Setting a random_state has no effect since shuffle is False.
This will raise an error in 0.24.
You should leave random_state to its default (None), or set shuffle=True.
  FutureWarning
```

mslapek:
Fixes #619

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Raise `FutureWarning` in `CVSplit` when `random_state` is not used. Will raise an exception in a future (#620)
+
 ### Fixed
 
 ## [0.8.0] - 2019-04-11
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -1,5 +1,5 @@
 """Contains custom skorch Dataset and CVSplit."""
-
+import warnings
 from functools import partial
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
Thanks for the quick PR, this looks very good to me, nice tests.

Two comments:

* what is the meaning of `ctor` in `test_ctor_random_state_warning`?
* could you please add an entry to CHANGES.md, which points out that this will be an error in the future?

mslapek:
Thanks for review.

> what is the meaning of ctor in test_ctor_random_state_warning?

Sometimes *ctor* is used as an shortcut for *constructor*.
I've made test name more descriptive in recent commit.

> could you please add an entry to CHANGES.md, which points out that this will be an error in the future?

Done.

Issue #482, opened 2019-06-09
Adding NeuralNetTranformer


YannDubs:
Hi, 

To be really sklearn-ish I think it would be very nice to have a class `NeuralNetTransformer` which inherits from `TransformerMixin` and implements a `.transform` method.

**Usecases: representation learning.** 

Example:
use a VAE for dimensionality reduction then work in this new representation space (no joint learning). This is often done in semi supervised learning (e.g. M1-M2 model), reinforcement learning (e.g. world models), ... but could also enable the use of all the predictors in sklearn by first transforming using a VAE / pretrained CNN without last layers and then predict using an svm or perform some clustering all in a single Pipeline.

**How**:
- the quickest is simple to have a transform function which does the same as predict (i.e. concatenate the first outputs of forward). But the representation is typically a temporary output of the model (i.e. latent in VAE or one of the middle layer in a CNN), so it would be better not to compute the whole forward pass if only the first few steps are needed. One could maybe have a `.transform` function in the model that only returns what is needed ? Or maybe a `is_transform` flag to the `forward` ?

PS: thanks for the great library :)

githubnemo:
Just so that I understand the intent correctly, you propose something that enables the user to easily use pretrained models or transforming/embedding models in sklearn pipelines? Generally I like this idea!

Basically we could implement `transform` as a call to `forward` and expect that most users will use the class for pretrained models that directly outputs e.g. embeddings. Users are able to override the `transform` method if they want to, for example in cases where models output features in `transform` and do classifications via `predict`.

YannDubs:
Yes this is exactly what I'm saying. Once in this new embedded space, one could use any algorithm in sklearn for classification / regression / clustering with much less parameters due to the low dimensionality of the embedded space. I feel that this would be a major use case of skorch: by taking advantage of the whole sklearn library.

thomasjpfan:
This is a really good idea! I am thinking of a workflow where one uses a `NeuralNetClassifier` for training and then pull out a piece of the neural net, put it into a `NeuralNetTransformer` as a feature extractor.

YannDubs:
@thomasjpfan that's exactly what I'm currently doing with a Semi Supervised VAE (M2), but I think that there are many other applications. E.g. use a resnet and then do for example clustering on the last convolutional output.

BenjaminBossan:
I believe the simplest implementation would be to have `transform` do what `predict_proba` currently does.

[...]

PR #621, opened 2020-04-20
ENH : set train / valid of criterion.

### LINKED ISSUE #565: [Feature Request] Calling `.eval` on criterion
usually criterions in pytorch inherit from nn.Module, meaning that they should have a `self.training` option that says whether it is during train or validation. yet skorch does not update it. This is useful for cases where there are multiple losses and the weights of each loss change during training (but not testing). E.g. beta-vae with annealing, Gans .... It seems that you though about that as `get_loss` takes `training` as argument. It would be useful to update the criterion accordingly :

```
def get_loss(self, y_pred, y_true, X=None, training=False):
    """Return the loss for this batch. [...]"""
    y_true = to_tensor(y_true, device=self.device)

    if isinstance(self.criterion_, nn.Module):
        if training:
            self.criterion_.train()
        else:
            self.criterion_.eval()

    return self.criterion_(y_pred, y_true)
```

YannDubs:
closes #565

Usually criterions in pytorch inherit from nn.Module, meaning that they should have a `self.training` option that says whether it is during train or validation. yet skorch does not update it. This is useful for cases where there are multiple losses and the weights of each loss change during training (but not testing). E.g. beta-vae with annealing, Gans ....

@BenjaminBossan

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1099,6 +1099,10 @@ class NeuralNet:
 
         """
         y_true = to_tensor(y_true, device=self.device)
+
+        if isinstance(self.criterion_, torch.nn.Module):
+            self.criterion_.train(training)
+
         return self.criterion_(y_pred, y_true)
 
     def get_dataset(self, X, y=None):


BenjaminBossan:
Thanks for taking this one.

Do you think there is a good way of testing this feature, so as to prevent regressions in the future?

BenjaminBossan:
ping @YannDubs

YannDubs:
Sorry I didn't really understand what you asked for. Are you asking for a unittest that checks whether or not the criterion is correctly on train or valid ? Or are you asking for a concrete usecases that could also be tested ?

I see 2 use cases for this (I actually used this for both):
1/ For annealing and changing of hyperparameters. Typically in vea you want to anneal beta (the lagrange multiplier), but it doesn't really make sense to annealy beta during validation
2/ More and more people run optimization procedures / have networks in the loss. For example the discriminator of GAN is often used as a loss for the generator (and not really as a module in itseld, as we discard it afterwards).

The second use case is more tricky as it requires to make other changes to skorch. For example I had to change the checkpointing procedures so as to also checkpoint the criterion. That's something I wanted to open as an issue to see what was you thought about this ...

ottonemo:
I think what Benjamin meant was going for was unit testing this feature :)

> The second use case is more tricky as it requires to make other changes to skorch. For example I had to change the checkpointing procedures so as to also checkpoint the criterion. That's something I wanted to open as an issue to see what was you thought about this ...

That's a good point. With the addition from this PR the criterion is left out from saving and with the changes in #597 it is easier to add custom modules to the neural net instances which are then *also* exempt from checkpointing. I feel that we need a checkpointing approach that scales better with 'auxiliary' objects that arise in more complex scenarios because I don't see adding yet another `f_`-parameter to the checkpoint callback as a good idea.

Let's discuss our options in #635.

BenjaminBossan:
> I think what Benjamin meant was going for was unit testing this feature :)

Yes, but ideally it's close to a real use case so that we have more confidence that the right thing happens.

BenjaminBossan:
I have a suggestion for two unit tests:

```python
    def test_criterion_training_set_correctly(self, net_cls, module_cls, data):
        # check that criterion's training attribute is set correctly

        X, y = data[0][:50], data[1][:50]  # don't need all the data
        side_effect = []

        class MyCriterion(nn.NLLLoss):
            """Criterion that records its training attribute"""
            def forward(self, *args, **kwargs):
                side_effect.append(self.training)
                return super().forward(*args, **kwargs)

        net = net_cls(module_cls, criterion=MyCriterion, max_epochs=1)
        net.fit(X, y)

        # called once with training=True for train step, once with
        # training=False for validation step
        assert side_effect == [True, False]

        net.partial_fit(X, y)
        # same logic as before
        assert side_effect == [True, False, True, False]

    def test_criterion_is_not_a_torch_module(self, net_cls, module_cls, data):
        X, y = data[0][:50], data[1][:50]  # don't need all the data

        def my_criterion():
            return torch.nn.functional.nll_loss

        net = net_cls(module_cls, criterion=my_criterion, max_epochs=1)
        net.fit(X, y)  # does not raise
```

They could be added to the end of `TestNeuralNet` in `test_net.py`. They are a bit artificial but should do the job. The first one fails with the provided fix, the second one is just a safeguard for potential regressions.

What do you think @YannDubs?

YannDubs:
I'm putting this on hold until Neurips deadline (10  days ). I will then send you a long proposal I have that will need this PR (and other things including #635 ). Happy to write tests and example notebooks (I'm thinking VAE and GAN) for that if you like the proposal. This will also close #482 . I prefer waiting util then as this PR is key and I will want to test many (real) usecases.
In short I think VAE's should be trained as `NeuralNet(module=Encoder,criterion=Decoder)` and GAN as `NeuralNet(module=Generator,criterion=Discriminator)`.  I'll be more precise then

BenjaminBossan:
> In short I think VAE's should be trained as `NeuralNet(module=Encoder,criterion=Decoder)` and GAN as `NeuralNet(module=Generator,criterion=Discriminator)`. I'll be more precise then

That sounds very intriguing, I haven't thought of that possibility. Looking forward to your proposal.

As to this PR, I think we can also just merge it for now and add the tests later (I think the ones I pasted above should be enough for now). If that makes your work easier, please say so and I'll merge.

YannDubs:
Sounds good to me ! thanks

BenjaminBossan:
> I will then send you a long proposal I have that will need this PR (and other things including #635 )

Hi @YannDubs this PR has been merged and there is a PR to address #635 (namely #652). Are you making progress with your proposal?

BenjaminBossan:
ping @YannDubs any news?

PR #623, opened 2020-04-20
[RFC] Refactor fit loop

BenjaminBossan:
# Introduction

I've been working on an idea I had some time ago
regarding (ostensibly) making developer experience nicer in
skorch when it comes to customizing the net.

So one of the pain points is that when I override certain certain
methods, I have to make sure that I don't forget to run `notify`
with the right arguments inside the methods. If I forget that,
the code will probably still run but produce some strange errors.

My goal was thus to automate away the `notify` calls. Now, if for
instance a user implements their own fit loop, they no longer
need to manually notfiy `on_train_begin` and `on_train_end`. To
achieve this, however, I had to resort to metaclasses, so there
is definitely a price for the convenience.

(Note: The problem still remains that if, say, someone completely
overrides `fit_loop` without any call -- directly or indirectly
-- to `train_step_single`, `on_grad_computed` still won't ever be
called; that's why you should usually call super() at some point.)

# Side effects

A side effect of my changes is that the exact X and y sent to
`on_batch_end` no longer depends on whether we use caching or not
in `EpochScoring`. I would consider this a positive, but it could
potentially break some code.

Another side effect of the same underlying change is that it's
not really possible anymore to differentiate between y being a
placeholder or not. That's a bit unfortunate but a minor
sacrifice overall.

Apart from that, I had to unifiy the signatures of all the `on_*`
methods, we have been somewhat sloppy in that regard but it
didn't cause trouble until now. The arguments passed to these
methods can also contain more elements now then they did before.

# Conclusion

I'm now putting this here as an RFC because I'm unsure if the
potential benefit is worth the tradeoff of increased code
complexity. IMO, if we can make a tradeoff that leads to
more complex code on the skorch side but to less complex code on
the user side, it's potentially worth it. What do you think?

Of course, with such a drastic change, we could break some
existing code. This should only affect heavily modified code, so
presumably only very few users. Still it's worth mentioning. I
think that since we're still below v1.0, we could afford to make
such a change.

One of my initial goals was to also automate the history record
calls. I didn't do it here (also haven't thought about how to do
it) because the effort to implement this is only worth if there
is an agreement to continue with this.

# Missing

* unit tests
* docstrings
* subclassing guideline in docs
* entry the CHANGES (this could break existing customized code)
* remove unused placeholder code
* also automate history logging??
* could add checks inside of `notfiy_decorator` that the right
  arguments were supplied to notify


BenjaminBossan:
I notice now that this will break every custom callback whose `on_*` methods don't have `*args` in the signature or are not exhaustively listing all `*args`. That's not so nice. If we still want to proceed with this, we would need to at least introduce a deprecation period beforehand.

BenjaminBossan:
After discussion with @ottonemo this is probably not the way to go, we will think about a better solution.

PR #626, opened 2020-04-27
[MRG] LRScheduler batch option

### LINKED ISSUE #610: Cosine annealing called every epoch instead of every batch

guybuk:
Closes #610 

I could use some assistance with schedulers that update once an epoch:

`optim._LRScheduler` initializes with `self.last_epoch=0`. This means that if we're training for `0,1,2,3,4` epochs, the respective values for `last_epoch` at the end of every epoch (where we update LR) will be `1,2,3,4,5`. This requires changes I made in the tests:

`assert lr_policy.lr_scheduler_.last_epoch == max_epochs-1` => `assert lr_policy.lr_scheduler_.last_epoch == max_epochs`

The reason this seems to have worked until now is in fact because of what I think is a bug:

It seems like until now, inside `on_epoch_end`, `epoch=len(net.history)-1=>epoch=0`. This means  that when `self.lr_scheduler_.step(epoch)` is called for the first time, the scheduler does not actually take a step because `self.last_step==epoch==0`, and `last_epoch` would stay at 0 for one extra epoch, which is why the test above passes (the one with `max_epochs - 1`).

I changed this code so that `epoch=len(history)` so the step would make sense, and replaced the order of `history.record` with `lr_scheduler_.step` so that from now on `history.record` will record **the LR with which the step was performed rather than the LR with which the next step will be performed** (which was this way until now).

Finally, I feel like as far as I can tell, all is in order. Except for `ReduceLROnPlateau`, which I'm struggling to understand how it works in general, and how the unique code for it inside `on_epoch_end` works.

Thanks!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed support for schedulers with a `batch_step()` method in `LRScheduler`. 
 - Raise `FutureWarning` in `CVSplit` when `random_state` is not used. Will raise an exception in a future (#620)
 - The behavior of method `net.get_params` changed to make it more consistent with sklearn: it will no longer return "learned" attributes like `module_`; therefore, functions like `sklearn.base.clone`, when called with a fitted net, will no longer return a fitted net but instead an uninitialized net; if you want a copy of a fitted net, use `copy.deepcopy` instead;`net.get_params` is used under the hood by many sklearn functions and classes, such as `GridSearchCV`, whose behavior may thus be affected by the change. (#521, #527)
-- Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch.
+- Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch. (#626)
 
 ### Fixed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,11 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added the `event_name` argument for `LRScheduler` for optional recording of LR changes inside `net.history`. NOTE: Supported only in Pytorch>=1.4
+- Added the `step_every` argument for `LRScheduler` to set whether the scheduler step should be taken on every epoch or on every batch.
 
 ### Changed
 
 - Removed support for schedulers with a `batch_step()` method in `LRScheduler`. 
 - Raise `FutureWarning` in `CVSplit` when `random_state` is not used. Will raise an exception in a future (#620)
+- Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch.
 
 ### Fixed
 
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -162,7 +162,7 @@ class LRScheduler(Callback):
                     score = -np.inf
                 elif self.lr_scheduler_.mode == 'min':
                     score = np.inf
-                elif epoch:
+                else:
                     score = net.history[-1, self.monitor]
 
             self.lr_scheduler_.step(score, epoch)
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -57,6 +57,10 @@ class LRScheduler(Callback):
       Pass ``None`` to disable placing events in history.
       **Note:** This feature works only for pytorch version >=1.4
 
+    step_every: str, (default='epoch'_
+      Value for when to apply the learning scheduler step. Can be either 'batch' or
+      'epoch'.
+
     kwargs
       Additional arguments passed to the lr scheduler.
 
@@ -66,10 +70,12 @@ class LRScheduler(Callback):
                  policy='WarmRestartLR',
                  monitor='train_loss',
                  event_name="event_lr",
+                 step_every='epoch',
                  **kwargs):
         self.policy = policy
         self.monitor = monitor
         self.event_name = event_name
+        self.step_every = step_every
         vars(self).update(kwargs)
 
     def simulate(self, steps, initial_lr):
@@ -119,7 +125,7 @@ class LRScheduler(Callback):
         # These are the parameters that are passed to the
         # scheduler. Parameters that don't belong there must be
         # excluded.
-        excluded = ('policy', 'monitor', 'event_name')
+        excluded = ('policy', 'monitor', 'event_name', 'step_every')
         kwargs = {key: val for key, val in vars(self).items()
                   if not (key in excluded or key.endswith('_'))}
         return kwargs
@@ -135,6 +141,8 @@ class LRScheduler(Callback):
         )
 
     def on_epoch_end(self, net, **kwargs):
+        if not self.step_every == 'epoch':
+            return
         epoch = len(net.history) - 1
         if isinstance(self.lr_scheduler_, ReduceLROnPlateau):
             if callable(self.monitor):
@@ -150,20 +158,19 @@ class LRScheduler(Callback):
             self.lr_scheduler_.step(score, epoch)
             # ReduceLROnPlateau does not expose the current lr so it can't be recorded
         else:
-            self.lr_scheduler_.step(epoch)
             if self.event_name is not None and hasattr(
                     self.lr_scheduler_, "get_last_lr"):
                 net.history.record(self.event_name, self.lr_scheduler_.get_last_lr()[0])
+            self.lr_scheduler_.step()
 
     def on_batch_end(self, net, training, **kwargs):
-        if not training:
+        if not training or not self.step_every == 'batch':
             return
-        if TorchCyclicLR and isinstance(self.lr_scheduler_, TorchCyclicLR):
-            self.lr_scheduler_.step(self.batch_idx_)
-            if self.event_name is not None and hasattr(
-                    self.lr_scheduler_, "get_last_lr"):
-                net.history.record_batch(self.event_name,
-                                         self.lr_scheduler_.get_last_lr()[0])
+        if self.event_name is not None and hasattr(
+                self.lr_scheduler_, "get_last_lr"):
+            net.history.record_batch(self.event_name,
+                                     self.lr_scheduler_.get_last_lr()[0])
+        self.lr_scheduler_.step()
         self.batch_idx_ += 1
 
     def _get_scheduler(self, net, policy, **scheduler_kwargs):
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -116,8 +116,8 @@ class LRScheduler(Callback):
         self.lr_scheduler_ = None
         self.batch_idx_ = 0
         # TODO: Remove this warning on 0.10 release
-        if self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR" \
-                and self.step_every == 'epoch':
+        if (self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR"
+                and self.step_every == 'epoch'):
             warnings.warn(
                 "The LRScheduler now makes a step every epoch by default. "
                 "To have the cyclic lr scheduler update "
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -3,6 +3,8 @@
 import sys
 
 # pylint: disable=unused-import
+import warnings
+
 import numpy as np
 import torch
 from torch.optim.lr_scheduler import _LRScheduler
@@ -57,9 +59,9 @@ class LRScheduler(Callback):
       Pass ``None`` to disable placing events in history.
       **Note:** This feature works only for pytorch version >=1.4
 
-    step_every: str, (default='epoch'_
-      Value for when to apply the learning scheduler step. Can be either 'batch' or
-      'epoch'.
+    step_every: str, (default='epoch')
+      Value for when to apply the learning scheduler step. Can be either 'batch'
+       or 'epoch'.
 
     kwargs
       Additional arguments passed to the lr scheduler.
@@ -77,6 +79,14 @@ class LRScheduler(Callback):
         self.event_name = event_name
         self.step_every = step_every
         vars(self).update(kwargs)
+        if policy == TorchCyclicLR or policy == "TorchCyclicLR":
+            warnings.warn(
+                "The LRScheduler callback makes a step "
+                "every epoch by default from now on. To have the cyclic lr "
+                "scheduler update every batch, "
+                "set step_every='batch'",
+                DeprecationWarning,
+            )
 
     def simulate(self, steps, initial_lr):
         """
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
Thanks again for working on this @guybuk. I'll take a closer look when I have more time. 

@thomasjpfan is the export for LR schedulers, hopefully he has time to answer your questions.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -94,10 +94,10 @@ def test_lr_callback_steps_correctly(
         )
         net.fit(X, y)
         # pylint: disable=protected-access
-        assert lr_policy.lr_scheduler_.last_epoch == max_epochs - 1
+        assert lr_policy.lr_scheduler_.last_epoch == max_epochs
 
     @pytest.mark.parametrize('policy, kwargs', [
-        (TorchCyclicLR, {'base_lr': 1e-3, 'max_lr': 6e-3}),
+        (TorchCyclicLR, {'base_lr': 1e-3, 'max_lr': 6e-3, 'step_every': 'batch'}),
```
thomasjpfan, line 100: This would not be backwards compatible with `TorchCyclicLR` with default arguments. As in `TorchCyclicLR` would step every epoch if a user does not pass in `step_every='batch'`.

skorch/callbacks/lr_scheduler.py
```diff
@@ -57,6 +57,10 @@ class LRScheduler(Callback):
       Pass ``None`` to disable placing events in history.
       **Note:** This feature works only for pytorch version >=1.4
 
+    step_every: str, (default='epoch'_
```
thomasjpfan, line 60:
```suggestion
    step_every: str, (default='epoch')
```

BenjaminBossan:
> For skorch, since we are support torch > 1.1, this change should be okay.

We actually do support 1.1 to 1.4. However, I was thinking of adding 1.5, maybe we should drop 1.1 then? My impression is that since PyTorch introduces so few breaking changes, most people use the more recent versions, but I have no data to back that up.

thomasjpfan:
Sorry I mistyped. I meant to say that we support `pytorch >= 1.1` (with the equality), we can depend on the new behavior. So we do not need to drop `1.1`.

skorch/tests/callbacks/test_lr_scheduler.py
guybuk: So the implementation is fine and we just need to restore the ` if TorchCyclicLR and isinstance(self.lr_scheduler_, TorchCyclicLR):...` segment?

thomasjpfan: Now that `step_every` is a parameter to `LRScheduler`, if we keep the `isinstance(self.lr_scheduler_, TorchCyclicLR)` piece, we will not be strictly following the `step_every` parameter.

We can have a `step_every='auto'` option that special cases the `CosineAnnealingLR` and `TorchCyclicLR`. Specifically, if the class is `CosineAnnealingLR` or `TorchCyclicLR` we use batch steps, otherwise we use epoch step. We will still allow the `batch` and `epoch` options for custom lr schdulers.

guybuk: That would mean introducing the entire concept of an 'auto' option ,and having to be backwards compatible for it in the future. However all it'll do is special case cyclic (for the record, cosine annealing isn't hard coded to work with batches, only cyclic). 

This basically means that if we don't want 'auto'  to just be a cover up for the cyclic case and instead be meaningful, we will have to add reasonable recommendations for every scheduler we introduce (which will be the default auto options). That will most likely involve having to read the paper from which it was introduced, or it's documentation.

BenjaminBossan: This is a bit of a conundrum. Do you know why some schedulers prefer batch and other epoch? Is it some derived property or more arbitrary?

> we will have to add reasonable recommendations for every scheduler we introduce

This is certainly not the way to go. Could we not just use whatever the default is?

guybuk: > This is a bit of a conundrum. Do you know why some schedulers prefer batch and other epoch? Is it some derived property or more arbitrary?

Not for certain. Naturally if you wanted your lr to change smoothly you'd do it per batch rather than epoch. I'm just plugging in a scheduler as is written in its documentation.

> This is certainly not the way to go. Could we not just use whatever the default is?

A scheduler doesn't have a default. It has a step function which the user can use whenever and wherever they'd like. Setting mode `epoch` or `batch` as defaults will break backwards compatibility for anything that isn't intended to use the default mode we choose.

Since the Cyclic scheduler is a special case for which skorch hacked a solution, I suggested leaving that solution as is (so that it won't conflict with this feature or have compatibility issues) until skorch makes an update that breaks compatibility, and just get rid of that hack then.

BenjaminBossan: Good points. Maybe it would be worth it to actually break backwards compatibility here for the sake of getting a clean solution. In fact, we could even detect when such a breakage happens and issue a warning to the user, with instructions how to revert to the old behavior. Would that be a compromise?

thomasjpfan: > Would that be a compromise?

I would be happy with a deprecation warning with a suggestion of using `step_every='batch'` in the future.

skorch/callbacks/lr_scheduler.py
guybuk: Fixed

skorch/tests/callbacks/test_lr_scheduler.py
guybuk: Done.

guybuk:
Looked at the `ReduceLROnPlateau` tests, and shifted the `if` statements in `on_epoch_end` around so the tests pass. I probably need someone who knows the feature to make sure it's still ok.

skorch/callbacks/lr_scheduler.py
```diff
@@ -158,19 +158,20 @@ def on_epoch_end(self, net, **kwargs):
...
-                    if self.lr_scheduler_.mode == 'max':
-                        score = -np.inf
-                    else:
-                        score = np.inf
+                if self.lr_scheduler_.mode == 'max':
+                    score = -np.inf
+                elif self.lr_scheduler_.mode == 'min':
+                    score = np.inf
+                elif epoch:
+                    score = net.history[-1, self.monitor]
```
BenjaminBossan, line 166: I'm no expert on lr schedulers, so can't comment on the exact logic that should be applied here. However, what I'm seeing is that the logic might have a "hole", where score wouldn't be defined at all if none of the `if` or `elif` matches. In the logic before, score would always be defined.

Could it be that we don't actually need `elif epoch`? Previously, we accessed `net.history[-2, self.monitor]`, i.e. the second to last row in the `history`. Therefore, we had to make sure that we're not in the first epoch (at least that's my interpretation of things). Now that we're accessing `net.history[-1, self.monitor]`, we can probably never get an index error here, hence don't need the guard. Would that make sense?

guybuk: Makes sense. Pushed a change.

skorch/callbacks/lr_scheduler.py
```diff
@@ -66,11 +72,21 @@ def __init__(self,
...
                  **kwargs):
         self.policy = policy
         self.monitor = monitor
         self.event_name = event_name
+        self.step_every = step_every
         vars(self).update(kwargs)
+        if policy == TorchCyclicLR or policy == "TorchCyclicLR":
+            warnings.warn(
+                "The LRScheduler callback makes a step "
+                "every epoch by default from now on. To have the cyclic lr "
```
thomasjpfan, line 85: For deprecation, we must keep the current behavior and state that the behavior will change in a Future version:

> Starting in 0.10, the LRScheduler will make a step every epoch by default. To have the cyclic lr schedule update every batch set step_every='batch'".

I am thinking of moving this logic into `initialize`:

```py
def initialize(self):
	self.step_every_ = self.step_every
    if policy == TorchCyclicLR or policy == "TorchCyclicLR" and self.step_every=='epoch':
		self.step_every_ = 'batch'
```

And then use `self.step_every_` in the other methods.

@BenjaminBossan How do you feel about this approach to deprecating the behavior? The simpler approach would be to update in `__init__`.

skorch/callbacks/lr_scheduler.py
```diff
@@ -66,11 +72,21 @@ def __init__(self,
...
+                "scheduler update every batch, "
+                "set step_every='batch'",
+                DeprecationWarning,
```
thomasjpfan, line 88: FutureWarning shows by default to users.

```suggestion
                FutureWarning,
```

skorch/callbacks/lr_scheduler.py
guybuk: Could you clarify the advantage of having both `self.step_every` and `self.step_every_`? It looks to me like having a second variable is unnecessary.

Once we decide whether to place this code in `initialize` or `init`, I'll make the changes.

thomasjpfan: Traditionally the sckit-learn API does not allow for changing the `__init__` parameters during construction. But since this is a callback and not an estimator, we may not need to strictly follow this API contract.

BenjaminBossan: Please try to stick with the sklearn convention here and don't change the `__init__` parameters. This means making the change during `initialize` (which is equivalent to `fit` in sklearn) and use the trailing underscore. The reason why we need this is because it makes it possible to use `set_params` on callbacks.

> How do you feel about this approach to deprecating the behavior?

When it comes to deprecation: I don't have a strong opinion. From my point of view, it would also be okay to change the behavior to the new behavior, raise a warning and pass a message that explains how to get back to the old behavior. But deprecation is also totally fine with me.

Regardless of how, could you please add a `# TODO: ...` comment, @guybuk, so that we don't forget to eventually change the behavior and remove the warning?

thomasjpfan: > When it comes to deprecation: I don't have a strong opinion. From my point of view, it would also be okay to change the behavior to the new behavior, raise a warning and pass a message that explains how to get back to the old behavior. But deprecation is also totally fine with me.

In this case, I am +0.5 with moving fast and go with the behavior change.

Let's change the behavior with a FutureWarning, we can remove the warning in `0.10`. With the behavior change, we will not need to have a `self.step_every_`.

guybuk: Done. Let me know if there's anything else.

Also, thanks for holding my hand throughout the process. I hope it was worth you time :smiley:

skorch/callbacks/lr_scheduler.py
```diff
@@ -123,6 +115,15 @@ def initialize(self):
         self.policy_ = self._get_policy_cls()
         self.lr_scheduler_ = None
         self.batch_idx_ = 0
+        # TODO: Remove this warning on 0.10 release
+        if self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR":
```
thomasjpfan, line 119: Only need to warn if `step_every` is `'epoch'`.

```suggestion
        if self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR" and self.step_every == 'epoch':
```

We should also have a test to make sure the warning message is raised.

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -241,6 +241,32 @@ def test_lr_scheduler_record_batch_step(self, classifier_module, classifier_data
...
+        (TorchCyclicLR, {'base_lr': 123, 'max_lr': 999})
+    ])
+    def test_cyclic_lr_with_epoch_step_warning(self,
+                                               classifier_module,
+                                               classifier_data,
+                                               policy,
+                                               kwargs):
+        with pytest.warns(None) as record:
+            scheduler = LRScheduler(
+                policy, **kwargs)
```
thomasjpfan, line 254: We do not need to parametrize

```suggestion
                TorchCyclicLR, base_lr=123, max_lr=999)
```

skorch/callbacks/lr_scheduler.py
```diff
@@ -116,7 +116,8 @@ def initialize(self):
         self.lr_scheduler_ = None
         self.batch_idx_ = 0
         # TODO: Remove this warning on 0.10 release
-        if self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR":
+        if self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR" \
+                and self.step_every == 'epoch':
```
thomasjpfan, line 120: Nit:

```suggestion
        if (self.policy_ == TorchCyclicLR or self.policy_ == "TorchCyclicLR" 
                and self.step_every == 'epoch'):
```

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -241,6 +241,32 @@ def test_lr_scheduler_record_batch_step(self, classifier_module, classifier_data
...
 
+    @pytest.mark.parametrize('policy, kwargs', [
+        (TorchCyclicLR, {'base_lr': 123, 'max_lr': 999})
+    ])
+    def test_cyclic_lr_with_epoch_step_warning(self,
+                                               classifier_module,
+                                               classifier_data,
+                                               policy,
+                                               kwargs):
+        with pytest.warns(None) as record:
```
thomasjpfan, line 252:
```suggestion
        msg = "The LRScheduler now makes a step every epoch by default. "
        with pytest.warns(FutureWarning, match=msg) as record:
```

After this change, we can remove the lines that come after `assert len(record) == 1`

CHANGES.md
```diff
@@ -11,12 +11,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added the `step_every` argument for `LRScheduler` to set whether the scheduler step should be taken on every epoch or on every batch.
 
 ### Changed
 
 - Removed support for schedulers with a `batch_step()` method in `LRScheduler`. 
 - Raise `FutureWarning` in `CVSplit` when `random_state` is not used. Will raise an exception in a future (#620)
 - The behavior of method `net.get_params` changed to make it more consistent with sklearn: it will no longer return "learned" attributes like `module_`; therefore, functions like `sklearn.base.clone`, when called with a fitted net, will no longer return a fitted net but instead an uninitialized net; if you want a copy of a fitted net, use `copy.deepcopy` instead;`net.get_params` is used under the hood by many sklearn functions and classes, such as `GridSearchCV`, whose behavior may thus be affected by the change. (#521, #527)
+- Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch.
```
thomasjpfan, line 21:
```suggestion
- Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch. (#626)
```

PR #634, opened 2020-05-11
Support pytorch 1.5

BenjaminBossan:
I did not actually test this locally, will just see if CI passes. But from the changelog of PyTorch, I didn't spot anything that we would need to change on the skorch side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: xenial
 python:
   - "3.7.5"
   - "3.6"
-  - "3.5"
+  - "3.5"  # TODO: drop support in 0.10.0
 env:
   - PYTORCH_VERSION="1.1.0"
   - PYTORCH_VERSION="1.2.0"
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
Everything seems to be fine, as expected. Again, I didn't run this on GPU, so if someone could test out PyTorch 1.5 on GPU, that would be nice.

Only downside is that CI is taking really long by now. It will get worse once we add Python 3.8. I wonder how important it is for us to support Python 3.5?

thomasjpfan:
> I wonder how important it is for us to support Python 3.5?

The ecosystem has dropped Python 3.5 by following https://numpy.org/neps/nep-0029-deprecation_policy.html

>  if someone could test out PyTorch 1.5 on GPU, that would be nice.

I'll run it.

thomasjpfan:
Confirmed tests pass on python3.7 and python3.8 with cuda 10.2

BenjaminBossan:
> Confirmed tests pass on python3.7 and python3.8 with cuda 10.2

Thanks a lot.

> The ecosystem has dropped Python 3.5 by following https://numpy.org/neps/nep-0029-deprecation_policy.html

I didn't know about this, thanks.

I added a future change warning to the CHANGES.md to give some adjustment time. The idea is to drop official support of Python 3.5 with the skorch 0.10.0 release. After that point, we could allow the use of f-strings and other more recent language additions.

PR #638, opened 2020-05-16
Make intent of output of get_params_for_optimizer more explicit

BenjaminBossan:
This is a follow up of the discussion in #596

The idea is for `get_params_for_optimizer` to to return 
`*args` and `**kwargs`, but that wasn't quite clear. This
is now clarified by the use of a tuple for `args`, names of
variables, and adjusted docstring.


BenjaminBossan:
ping @ottonemo

Issue #636, opened 2020-05-13
Issue with LBGFS optimizer: `optimizer_.zero_grad` no longer called inside of `train_step_single`


jwdink:
It looks like in the following commit, the call to `optimizer_.zero_grad` was removed from `train_step_single`, and instead put inside of `train_step`:

https://github.com/skorch-dev/skorch/commit/a62e4190d38bf698a314c4283ca7a9dd7cfbe428#diff-5d1f8e13ab8b822b72e9b295cf7301c5L596

As far as I can tell, this isn't appropriate if you're using the LBFGS optimizer -- for that optimizer you need to zero the gradient on every call to the closure.

I ran into this because nets that had previously worked on skorch 0.60 now quickly hit `nan` loss on skorch 0.80. I reverted versions and hunted down this commit for the cause (see [here](https://github.com/strongio/strong-glm/blob/master/examples/ceiling_weibull.py) for an example).

BenjaminBossan:
@jwdink Thank you a lot for reporting this error.

You are right, the commit you identified is the reason for the trouble.

This is actually a difficult problem to solve. The idea of moving the `zero_grad` out of each individual call of `optimizer.step` was to allow the user to have more control over it, e.g. to implement gradient accumulation. This is in conflict with optimizers that want to call the closure multiple times and need `zero_grad` after each of it.

As a quick fix for you, I would propose to either restore the previous state or move the call to `zero_grad` inside the `step_fn` closure here:

https://github.com/skorch-dev/skorch/blob/1f6b5428a0d58b9d0d2ef08384aeb45c1c14df55/skorch/net.py#L655-L660

I believe it should fix your issue.

Now what could be done in the long run? Hard to tell. Reverting the change removes the mentioned benefit. We could, hypothetically, add a parameter to `NeuralNet` that allows to switch between the two. But that still requires the user of LBFGS or similar optimizers to actively remember that. It's thus not a very nice solution.

If we could somehow know what optimizers actually need to call `zero_grad`, we could control the behavior dynamically. But I don't think we can know that. So this is also not a solution.

I never use optimizers that require that, thus have almost no experience. Do you have any good idea?

jwdink:
> Now what could be done in the long run? Hard to tell. Reverting the change removes the mentioned benefit. We could, hypothetically, add a parameter to NeuralNet that allows to switch between the two. But that still requires the user of LBFGS or similar optimizers to actively remember that. It's thus not a very nice solution.

[...]

PR #639, opened 2020-05-16
Call zero grad on each train step

BenjaminBossan:
This fixes a bug introduced by moving the `optimizer.zero_grad()` call
outside of the train step function, making it incompatible with LBFGS
and other optimizers that call the train step several times per
batch and expect the gradient to be reset after each call.

I extended the existing test for LBFGS to actually check the
training and validation scores after the fit call. With the
faulty behavior, the scores are actually diverging, and with the
bugfix, we get reasonable scores.

In #636, I wrote

> So we could introduce a parameter like perform_zero_grad_each_step=True or so (not a very nice name) and if a user wants to do their custom stuff with gradients, they need to turn it off.

After some thinking, I believe it's not actually necessary to
introduce an argument to turn off the `zero_grad` call during
each train step. The reason is that if a user wants to do some
custom stuff like gradient accumulation, they'll have to override
`train_step` anyway, which means they'll control how exactly
`zero_grad` will be called.

### DIFF
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -2138,13 +2138,21 @@ class TestNeuralNet:
         assert net.history[:, "train_batch_count"] == [train_batch_count]
         assert net.history[:, "valid_batch_count"] == [valid_batch_count]
 
+    @flaky(max_runs=3)
     def test_fit_lbfgs_optimizer(self, net_cls, module_cls, data):
         X, y = data
         net = net_cls(
             module_cls,
             optimizer=torch.optim.LBFGS,
-            batch_size=len(X))
+            lr=1.0,
+            batch_size=-1,
+        )
         net.fit(X, y)
+
+        last_epoch = net.history[-1]
+        assert last_epoch['train_loss'] < 1.0
+        assert last_epoch['valid_loss'] < 1.0
+        assert last_epoch['valid_acc'] > 0.75
 
     def test_accumulator_that_returns_last_value(
             self, net_cls, module_cls, data):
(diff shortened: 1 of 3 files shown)


skorch/tests/test_net.py
```diff
@@ -2138,14 +2138,22 @@ def test_batch_count(self, net_cls, module_cls, data, batch_size):
         assert net.history[:, "train_batch_count"] == [train_batch_count]
         assert net.history[:, "valid_batch_count"] == [valid_batch_count]
 
+    @flaky(max_runs=3)
```
thomasjpfan, line 2141: Does increasing `max_epochs` make this less flaky?

BenjaminBossan: The test is not actually flaky. I just added it because I thought it to be cautious for any test where we expect the model to reach a certain score.

My general impression is that tests have become less flaky, perhaps because of the `seeds_fixed` auto-fixture. Do you believe we should remove the flaky decorators?

thomasjpfan: It would be nice to have a `seeds_fixed`. Although this is not a blocker for this PR.

BenjaminBossan: We have it already here:

https://github.com/skorch-dev/skorch/blob/9fafede929891e1c5c98ba4e7764035b75a25d26/skorch/tests/conftest.py#L23-L27

thomasjpfan: I was just about to implement this and saw it. You added it not too long ago. :) This should mean our tests are deterministic. (hopefully)

BenjaminBossan:
ping @ottonemo

PR #643, opened 2020-05-29
Add tests for criterion being set to train/valid

BenjaminBossan:
As discussed in #621.

Also: Add entry to CHANGES.md

ping @YannDubs


BenjaminBossan:
Thanks for reviewing

Issue #605, opened 2020-03-06
skorch.fit can't handle lists of lists with variable length


econti:
I'm having a hard time figuring out how to pass a list of lists (with variable length) to skorch's `fit` method.

Specifically, I have a feature that is a list of ID's (e.g. `[[1, 12, 3], [6, 22]...]`) which are converted to a dense representation using an embedding table in my PyTorch module's `forward` method:

```
def forward(self, X_float, X_id_list):
    ...
```

When I call `net.fit()` on my data set (e.g. `{"X_float": ..., "X_id_list": ...}` I get the following error caused by the list of lists:

```
ValueError: Dataset does not have consistent lengths.
```

I've also tried converting the list of lists to a pandas dataframe and numpy array (of objects) and neither works. How do you handle variable length lists of lists in `skorch.fit`?

cgarciae:
Don't know about the specifics of this in skorch but generally you need to add padding / perform slicing so every sample has the same length. The only exception of this are Tensorflow's Ragged Tensors, but even then you have to specify a default value to pad with when converting to regular tensors (Pytorch doesn't have Ragged Tensors yet).

BenjaminBossan:
@econti Could you check whether [`PackedSequence`](https://pytorch.org/docs/stable/nn.html#packedsequence) solves your issue?

Otherwise, we have an example [here](https://nbviewer.jupyter.org/github/skorch-dev/skorch/blob/master/examples/rnn_classifer/RNN_sentiment_classification.ipynb) that shows how to potentially deal with variable length sequences.

econti:
Thanks @BenjaminBossan, that did the trick for me. Leaving a code snippet here for anyone else who encounters a similar issue:

```
# data["X_id_list"] looks like:
# [[1, 3], [0, 40, 16], ...]

X_id_list = {}

for series_name, series in data["X_id_list"].iteritems():
    pre_pad = [torch.tensor(i) for i in series]
    X_id_list[series_name] = pad_sequence(
        pre_pad, batch_first=True, padding_value=0
    )
```

BenjaminBossan:
@econti Great that you found a solution and thanks for the snippet.

ToddMorrill:
[...]

PR #646, opened 2020-06-03
Don't pass `y=None` to the `train_split` function

kqf:
By default `skorch` requires the `train_split` function to have two positional arguments `X` and `y`. This leads to unexpected behavior when working on unsupervised tasks:

```python
def unsupervised_split(dataset, split_ratio=0.7):
    return # some splits

...

def build_model():
    model = skorch.NeuralNetClassifier(
        module=SimpleModule,
        train_split=unsupervised_split,
    )
    ...

# This causes an error as `y=None` will be passed as a positional 
# argument to `unsupervised_split` function.
model = build_model().fit(df)
```

As suggested by @BenjaminBossan, It is possible to define a wrapper function that accepts a dummy positional argument, to solve the problem, but the same can be achieved with the minimum changes on the `skorch` side.  This update also eases the integration with `torchtext` iterator/split methods (see the discussion [here](https://github.com/skorch-dev/skorch/issues/605#issuecomment-637042846)).


BenjaminBossan:
I had a quick look at this and it looks good so far. However, I'd like to think a bit longer on this since this change might break some existing code. I'd like to make sure that nothing severe breaks because of the change.

Meanwhile, could you please add an entry to the CHANGES.md, in the `Changed` section.

kqf:
Thanks for reviewing the PR, sure, take your time.

BenjaminBossan:
Thanks for updating the CHANGES.md.

I would like to see an explicit test for the new behavior. As of now, it's only tested coincidentally.

I'm a tiny bit nervous since this change could break existing code (as witnessed by the tests that needed changing). It's probably not necessary to raise a warning, but I could see adding a comment in line 1215 (before the `if y is None`) along these lines:

```
# After a change in #646, `y` is no longer passed to `self.train_split` if it is `None`. To revert to the previous behavior, remove the following two lines:
```

(of course with the proper indentations and line breaks)

The rest looks good to me.

skorch/tests/test_net.py
```diff
@@ -2490,6 +2490,36 @@ def initialize_module(self, *args, **kwargs):
...
         assert hidden_units == 99
 
+    @pytest.mark.parametrize("needs_y, train_split, raises", [
+        (False, None, ExitStack()),  # ExitStack = does not raise
+        (True, None, ExitStack()),
+        (False, lambda x: (x, x), ExitStack()),
+        (True, lambda x, y: (x, x), ExitStack()),
+        (True, lambda x: (x, x), pytest.raises(TypeError)),  # Raises an error
+    ])
+    def test_train_split_with_nans(self, needs_y, train_split, raises):
```
BenjaminBossan, line 2500: I'm a bit confused about the name `test_train_split_with_nans`, where do nans come into play? Is the test not about `y` being passed to `train_split` only when it is not None?

skorch/tests/test_net.py
```diff
@@ -2490,6 +2490,36 @@ def initialize_module(self, *args, **kwargs):
...
+        net = NeuralNet(
+            MLPModule,
+            module__input_units=n_features,
+            max_epochs=2,
+            criterion=UnsupervisedLoss,
+            train_split=train_split,
+        )
+        with raises:
+            net.fit(X, y)
+            assert net.predict(X) is not None
```
BenjaminBossan, line 2521: Could you explain why this assertion is necessary?

kqf: It's not, I think. I wanted to be more explicit: even though the `y=None` the predictions still make sense. I can change this 🤷‍♂️

skorch/tests/test_net.py
kqf: General remark: indeed, I didn't think well about the name of the test case. 

> where do nans come into play? 

The `NaN`s here come into play as `y=None` and `train_split=None`

>  Is the test not about y being passed to train_split only when it is not None?

I am writing a more verbose answer to this question as the PR comment, so we can move the discussion there

kqf:
Thanks again for the suggestions. 

Some explanations about the test case. The tests are applied to the `NeuralNet` as it's the only class that works for `y=None`. This means that neither classifier nor regressor should pass through `if y is not None` line. Then for each test scenario, we check if the model can fit and make predictions. There is a dummy criterion defined that ignores `y_pred` to make things easy to read, and random data are generated. 

The test cases, these two lines reproduce an old behaviour:
```python
        (False, None, ExitStack()),  # ExitStack = does not raise
        (True, None, ExitStack()),
        (True, lambda x, y: (x, x), ExitStack()),
```
Where the first two lines correspond to the default `train_split` method.

Then we need to check if it works when `train_split` does not require a positional `y`:
```python
        (False, lambda x: (x, x), ExitStack()),
        (True, lambda x: (x, x), pytest.raises(TypeError)),  # Raises an error
```
These cases forced me to change the test data in `test_scoring`. One can see this as a safety check that says: "you are trying to apply unsupervised split for supervised data". 

Thanks again for the review

skorch/tests/test_net.py
kqf: So, can we call this function `test_passes_y_to_train_split_when_not_nan`? Frankly, I can't invent a better name 😅

skorch/tests/test_net.py
BenjaminBossan: Yes, I think it's better to remove it, since it detracts from what is really tested. I actually believe it's impossible for `net.predict` to ever return None, so this assertion will never fail.

skorch/tests/test_net.py
BenjaminBossan: > The `NaN`s here come into play as `y=None` and `train_split=None`

Can you elaborate on that? Do you mean `NaN` in the sense of "not a number"? I don't see how that can happen.

> So, can we call this function `test_passes_y_to_train_split_when_not_nan`? Frankly, I can't invent a better name

With tests, it's always better to use a lengthy name that is more descriptive than a short, less descriptive name. Nobody is going to import your test anyway. If in doubt, tests should also rather contain too many than too few explanatory comments (some skorch tests are hard to understand because we didn't do that right from the start).

Regarding the name itself, I would prefer `test_passes_y_to_train_split_when_not_none` (last word changed), since that is what `y` actually is.

BenjaminBossan:
Thanks for adding the explanation. I actually believe that a condensed form of it could be added to the test (as I mentioned above, better to add too many than too few comments in tests).

> Where the first two lines correspond to the default `train_split` method.

This is not quite true, since the default `train_split` is `CVSplit(5)`, not `None`. Maybe you could actually even add the former?

> Thanks again for the review

Thanks for the PR ;)

kqf:
> This is not quite true, since the default train_split is CVSplit(5)

😮  I overlooked that. For some reason I thought, that it goes like `self.train_split = train_split or CVSplit(5)`. Not sure if `CVSplit` invoked with `y=None` should fail or not. I have to think. 

I agree to the comments, I'll try to fix the issues tomorrow  (now I am way from my workstation).

skorch/tests/test_net.py
```diff
@@ -2493,32 +2493,45 @@ def initialize_module(self, *args, **kwargs):
...
         (True, lambda x: (x, x), pytest.raises(TypeError)),  # Raises an error
     ])
-    def test_train_split_with_nans(self, needs_y, train_split, raises):
+    def test_passes_y_to_train_split_when_not_none(
+            self, needs_y, train_split, raises):
         from skorch.net import NeuralNet
         from skorch.toy import MLPModule
 
+        # By default, `train_split=CVSplit(5)` in the `NeuralNet` definition
+        if train_split == "default":
```
kqf, line 2508: This looks a bit ugly, but I didn't want to introduce a new function/import to this namespace.

skorch/tests/test_net.py
```diff
@@ -2493,32 +2493,45 @@ def initialize_module(self, *args, **kwargs):
...
+            from skorch.dataset import CVSplit
+            train_split = CVSplit(5)
```
kqf, line 2510: This one is an explicit statement, however it can be replaced by the implicit one for sake of extensibility:

```python
train_split = NeuralNet(None, None).train_split
```
So, this way you don't have to modify the test when the default value of the parameter is changed. 
@BenjaminBossan , what do you think?

skorch/tests/test_net.py
kqf: Agreed, I think this can be resolved

skorch/tests/test_net.py
kqf: > Can you elaborate on that?

I think this is covered in the PR comments. Yes, I just confused `None` and `NaN` in the conversation, sorry. 

I agree to the changes, see the new code.

skorch/tests/test_net.py
BenjaminBossan: I think it would be even better to solve it like this:

```python
kwargs = {} if train_split == 'default' else {'train_split': train_split}
```

and then below:

```python
net = NeuralNet(
   ...
   **kwargs)
```

skorch/tests/test_net.py
```diff
@@ -2493,32 +2493,45 @@ def initialize_module(self, *args, **kwargs):
...
             def forward(self, y_pred, _):
                 return y_pred.mean()
 
+        # Generate the dummy dataset
         n_samples, n_features = 128, 10
         X = np.random.rand(n_samples, n_features).astype(np.float32)
         y = np.random.binomial(n=1, p=0.5, size=n_samples) if needs_y else None
 
+        # The `NeuralNetClassifier` or `NeuralNetRegressor` always require `y`
+        # Only `NeuralNet`can transfer `y=None` to `tran_split` method.
```
BenjaminBossan, line 2523:
```suggestion
        # Only `NeuralNet` can transfer `y=None` to `train_split` method.
```

PR #648, opened 2020-06-07
Raise a FutureWarning if Python version is < 3.6

BenjaminBossan:
In skorch 0.10.0, this Python version will no longer be officially supported.

Also, use correct import order in utils.py.

### DIFF
--- a/skorch/__init__.py
+++ b/skorch/__init__.py
@@ -1,5 +1,8 @@
 """skorch base imports"""
 
+import sys
+import warnings
+
 import pkg_resources
 from pkg_resources import parse_version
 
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
> I haven't seen a warning for ending support done like this before

I was not sure what the correct way to do this would be. Do you have another proposal? Anyway, I think that even if we reach skorch 0.10, it doesn't mean we immediately add features that are incompatible with Python 3.5 or below.

thomasjpfan:
> it doesn't mean we immediately add features that are incompatible with Python 3.5 or below.

The major feature I would want to use is... f-strings

> Do you have another proposal?

I think most just stop support and have it error in `setup.py` when support is dropped. I think this warning approach is interesting and it is worth trying.

BenjaminBossan:
> The major feature I would want to use is... f-strings

I guess me too but I wouldn't necesarily change the existing code just because f-strings might be nicer to use.

> I think most just stop support and have it error in `setup.py` when support is dropped.

I thought so too but it's a bit rude :) Since we know we plan to do this, I thought why not warn users.

Issue #631, opened 2020-05-09
Cannot fit NeuralNetBinaryClassifier when y is None


maifeng:
https://github.com/skorch-dev/skorch/blob/fab7fc951f943539575f5a4426ef2093391a2f04/skorch/classifier.py#L289-L292

Hi! Using the following example will throw the exception `AttributeError: 'NoneType' object has no attribute 'dim'`, because the `check_data()` function inside `NeuralNetBinaryClassifier` checks the dimension of y. This behavior is inconsistent with the documentation about using a Dataset as X. Using `NeuralNetClassifier` works. 

Suggested fix: Adding `if y is not None` be added before the dimension check?
 
```
import numpy as np
import torch
import torch.nn.functional as F
from sklearn.datasets import make_classification
from skorch import NeuralNetBinaryClassifier, NeuralNetClassifier
from skorch.callbacks import EarlyStopping
from skorch.dataset import Dataset
from torch import nn


class ClassifierModule(nn.Module):
    def __init__(self, num_units=10, nonlin=F.relu, dropout=0.5):
        super(ClassifierModule, self).__init__()
        self.num_units = num_units
        self.nonlin = nonlin
        self.dropout = dropout

        self.dense0 = nn.Linear(20, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(dropout)
        self.dense1 = nn.Linear(num_units, 10)
        self.output = nn.Linear(10, 2)

    def forward(self, X, **kwargs):
        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = F.relu(self.dense1(X))
        X = F.softmax(self.output(X), dim=-1)
        return X


class MyDataset(torch.utils.data.Dataset):
    def __init__(self, X, y):
        self.X = X
        self.y = y

        assert len(X) == len(y)

    def __len__(self):
        return len(self.X)

    def __getitem__(self, i):
        return self.X[i], self.y[i]


X, y = make_classification(1000, 20, n_informative=10, random_state=0)
X = X.astype(np.float32)
dataset = MyDataset(X, y)

net = NeuralNetBinaryClassifier(ClassifierModule, batch_size=32, train_split=None)
net.fit(dataset, y=None)
```

BenjaminBossan:
Yes, I think it's reasonable to make that change. Maybe if X is a dataset then y can be None. Do you want to add a PR for this?

asberk:
I have submitted a PR for this issue ( #653 )

[...]

PR #653, opened 2020-06-15
Implements fix to #631

asberk:
Only check `get_dim(y) != 1` if `X` is not a dataset.

### DIFF
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -71,7 +71,8 @@ class TestNeuralNet:
         X, y = data
         accuracy = net_fit.score(X, y)
         assert 0. <= accuracy <= 1.
-    def test_check_data_accepts_dataloader(self, net, data):
+
+    def test_check_data_accepts_skorch_dataset(self, net, data):
         from skorch.dataset import Dataset
 
         X, y = data
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -71,6 +71,12 @@ class TestNeuralNet:
         X, y = data
         accuracy = net_fit.score(X, y)
         assert 0. <= accuracy <= 1.
+    def test_check_data_accepts_dataloader(self, net, data):
+        from skorch.dataset import Dataset
+
+        X, y = data
+        dataset = Dataset(X, y)
+        assert net.check_data(dataset, y=None) is None
 
     # classifier-specific test
     def test_takes_log_with_nllloss(self, net_cls, module_cls, data):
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thanks @asberk do you think you could add a unit test for this case?

asberk:
I'll give it a shot, @BenjaminBossan , but I'm not familiar with `pytest`. Would you be willing to provide some guidance?

Right now, in `test_classifier.py` I've got something like

```python
def test_check_data(self, net, X, y):
    assert net.check_data(X, y)
```
This is not close to the ideal. Presumably, we'd like to test what happens if `X` is a dataset and `y` is/isn't `None`, as well as the other cases (like if `X, y = data`). Is there already a `fixture` set up for a dataset (`torch.utils.data.Dataset` or skorch.dataset.Dataset`)?

BenjaminBossan:
> I'll give it a shot, @BenjaminBossan , but I'm not familiar with pytest. Would you be willing to provide some guidance?

Great. Of course I'll try to help you.

So for this test, you want to check that training `NeuralNetBinaryClassifier` with a dataset but without `y` works. Therefore, your test should use a dataset (e.g. the one from skorch), which you can import inside the test. In addition, you could use the [data fixture](https://github.com/skorch-dev/skorch/blob/c72e1fec496e3d80b9ca0709df556c07c2522966/skorch/tests/test_classifier.py#L158-L161) and the [net fixture](https://github.com/skorch-dev/skorch/blob/c72e1fec496e3d80b9ca0709df556c07c2522966/skorch/tests/test_classifier.py#L179-L185) that are already there. The data can be used to initialize the dataset. After calling fit with the dataset, there should be no error, that's good enough. As a check for yourself, you could undo your code change and check if the test fails (in test driven development, you would actually first implement the test, then the fix).

Also, think about a descriptive name for the test.

I hope this gives you a good direction.

skorch/tests/test_classifier.py
```diff
@@ -72,6 +72,13 @@ def test_score(self, net_fit, data):
         accuracy = net_fit.score(X, y)
         assert 0. <= accuracy <= 1.
 
+    def test_check_data_accepts_skorch_dataset(self, net, data):
+        from skorch.dataset import Dataset
+
+        X, y = data
+        dataset = Dataset(X, y)
+        assert net.check_data(dataset, y=None) is None
```
BenjaminBossan, line 80:
```suggestion
    def test_fit_with_dataset_and_y_none(self, net_cls, module_cls, data):
        from skorch.dataset import Dataset

        # deactivate train split since it requires y
        net = net_cls(module_cls, train_split=False, max_epochs=1)
        X, y = data
        dataset = Dataset(X, y)
        assert net.fit(dataset, y=None)
```

BenjaminBossan:
@asberk Thanks a lot!

PR #658, opened 2020-06-23
to_device: Handle nested lists/tuples recursively

ottonemo:
The previous implementation of `to_device` would break when a
user decided to return a list of tensors in `forward`.

This patch applies `to_device` recursively and adds support
for lists in addition to tuples.

### DIFF
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -242,7 +242,7 @@ class TestToDevice:
         (None, None),
     ])
     def test_nested_data(self, to_device, x_list, device_from, device_to):
-        # Sometimes data is nested because it would need to be padded so its
+        # Sometimes data is nested because it would need to be padded so it's
         # easier to return a list of tensors with different shapes.
         # to_device should honor this.
         if 'cuda' in (device_from, device_to) and not torch.cuda.is_available():
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -155,6 +155,10 @@ class TestToDevice:
         length = torch.as_tensor([2, 2, 1])
         return pack_padded_sequence(value, length)
 
+    @pytest.fixture
+    def x_list(self):
+        return [torch.zeros(3), torch.ones(2, 4)]
+
     def check_device_type(self, tensor, device_input, prev_device):
         """assert expected device type conditioned on the input argument for `to_device`"""
         if None is device_input:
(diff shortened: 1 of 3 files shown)


skorch/tests/test_utils.py
```diff
@@ -230,6 +234,32 @@ def test_check_device_packed_padded_sequence(
...
 
+    @pytest.mark.parametrize('device_from, device_to', [
+        ('cpu', 'cpu'),
+        ('cpu', 'cuda'),
+        ('cuda', 'cpu'),
+        ('cuda', 'cuda'),
+        (None, None),
+    ])
+    def test_nested_data(self, to_device, x_list, device_from, device_to):
+        # Sometimes data is nested because it would need to be padded so its
```
BenjaminBossan, line 245:
```suggestion
        # Sometimes data is nested because it would need to be padded so it's
```

ottonemo:
> Just a heads up, there is also #657 working on `to_device`. Does it make sense to also adjust `to_numpy` with the same logic?

I think it makes sense for `predict_proba` and `transform`. I will add it.

> Also, it seems that when passing a list, `to_device` will return a tuple. Do you think it should be a list instead?

Fixed, thanks.

ottonemo:
@BenjaminBossan please review again.

ottonemo:
@BenjaminBossan I think there's an issue with this when calling `to_numpy([1,2,3])`. Before this returned a numpy array, now it doesn't anymore. Please don't merge this yet.

BenjaminBossan:
> I think there's an issue with this when calling `to_numpy([1,2,3])`. Before this returned a numpy array, now it doesn't anymore. Please don't merge this yet.

Is it correct that this raises a `TypeError` (old and new implementation)? I wonder if that's the desired behavior and if it is, if we should clarify that with a corresponding test.

ottonemo:
> > I think there's an issue with this when calling `to_numpy([1,2,3])`. Before this returned a numpy array, now it doesn't anymore. Please don't merge this yet.
> 
> Is it correct that this raises a `TypeError` (old and new implementation)? I wonder if that's the desired behavior and if it is, if we should clarify that with a corresponding test.

This is a question of scope. `to_numpy` was originally introduced to convert from PyTorch to numpy and included device handling / gradient detaching. Now we are transitioning away from that definition by handling other data types as well but its main purpose is still to convert from PyTorch to numpy (even when we are unpacking other data types). If you agree with this definition we should

1. make this explicit in the docs
2. make this explicit in the tests (check for `TypeError` when supplying non-pytorch data, even if nested)

BenjaminBossan:
That sounds very reasonable, I think `to_numpy` shouldn't grow in scope beyond what it delivers at the moment, it was really only meant to be a convenience function.

skorch/tests/test_utils.py
```diff
@@ -174,6 +178,25 @@ def test_tuple(self, to_numpy, x_tuple):
...
+        self.compare_array_to_tensor(x_numpy['b'][1], x_dict['b'][1])
+
+    @pytest.mark.parametrize('x_invalid', [
+        1,
+        [1,2,3],
+        (1,2,3),
+        {'a': 1},
+    ])
+    def test_invalid_inputs(self, to_numpy, x_invalid):
+        " Inputs that are invalid for the scope of to_numpy. "
```
BenjaminBossan, line 194: I've never seen this style, neither docstring nor comment. I think a comment would be sufficient here?

ottonemo: It is still a doc-string but according to PEP257 this style is disregarded. I converted it to a comment.

Example:
```python
>>> def foo():
...     "bla"
...     return 1
... 
>>> foo.__doc__
'bla'
```

BenjaminBossan: > It is still a doc-string but according to PEP257 this style is disregarded

Interesting, I didn't know, though in hindsight it makes sense.

PR #662, opened 2020-06-27
Add predict nonlinearity

### LINKED ISSUE #637: Add option to apply torch.softmax or disable torch.log internally
From what I've gathered, there are basically 3 options for losses for classification tasks when using Skorch.

1)  Have your network output probabilities from a `torch.softmax` layer and then use `nn.NLLLoss`, where Skorch is applying `orch.log` internally. But this has been shown to be numerically unstable. [Overview](https://discuss.pytorch.org/t/why-there-is-no-log-operator-in-implementation-of-torch-nn-nllloss/16610), [Details](https://timvieira.github.io/blog/post/2014/02/11/exp-normalize-trick/)
- No solution: based on my understanding of the links about, I would advise against doing this.

2) Have your network output from `nn.logsoftmax` and then use `nn.NLLLoss`, which is numerically stable. Currently Skorch is applying `torch.log` to the output of the network when `nn.NLLLoss` is specified. So `torch.log` is applied twice and the model doesn't train.
- Solution: have an option to disable the application of `torch.log` internally before `nn.NLLLoss`.

3) Have your network output raw logits (i.e. don't apply any sort of softmax) and then use `nn.CrossEntropyLoss`, which is numerically stable. Skorch will handle training but then `model.predict_proba` breaks down because you're getting raw logits. Obviously I can just apply the softmax function but it breaks the semantics of the sklearn API.
- Solution: have an option to apply of `torch.softmax` internally before `model.predict_proba`.

### LINKED ISSUE #661: [RFC] Add argument to specify output nonlinearity
Currently, the nonlinearity applied to the module output before returning it from `predict_proba` is not changeable.

For `NeuralNetClassifier`, no nonlinearity is applied, which means that if the module doesn't return probabilities (which might be desired, e.g. for numerical stability), `predict_proba` also doesn't return probabilities.

For `NeuralNetBinaryClassifier`, `sigmoid` is applied if the loss is an instance of `torch.nn.BCEWithLogitsLoss`. Otherwise, it's untouched.

For `NeuralNet` and `NeuralNetRegressor`, no nonlinearity is applied, which is typically what is desired (we still provide `predict_proba`, even though it's just the same as `predict`, but it might still be useful for users to have two methods).

The proposal is to add an argument to `NeuralNet` called `output_nonlinearity` (or similar) that is applied to the module's result. This makes it very easy for the user to change the output nonlinearity.

Where is the right place to apply it? I would argue on a batch level inside `forward_iter`, before calling `to_device`. There are three main reasons:

1. Use GPU acceleration if so desired
2. Use pytorch functions instead of numpy -- it's more likely that pytorch provides an appropriate function
3. Not only affects `predict_proba`, but also `forward` and `forward_iter`, which should be preferable for most

Regarding the implementation, it should be backwards compatibly, i.e. the solution should do exactly the same as is being done at the moment. To achieve this, the default `output_nonlinearity` should be `'auto'`. In that case, a special function should be called that infers the nonlinearity from the `net` instance and reproduces the existing behavior. We could think about extending the inferred nonlinearities compared to what is done right now, e.g. to apply `softmax` is the loss is `CrossEntropyLoss`.

(An interesting design decision would be if the `auto` function "dispatches" only on the `criterion` or also on the `net`. E.g., if the `criterion` is `BCEWithLogitsLoss`, is `sigmoid` always applied, or only in case of `NeuralNetBinaryClassifier`? Regardless of the answer, I would like to have the `auto` function take the whole net as argument, not only the the criterion, to infer the nonlinearity, even if we decide to only use the criterion for now, in case we need/want to extend it further.)

Other options for `output_nonlinearity` should be callables, like `torch.softmax`, or `None`, in which case the output is untouched.

This proposal supersedes #572 and #580 (which seem to be abandoned anyway).

Note that this change _might_ break existing code, where methods like `predict_proba` are overridden and make the assumption that the outputs from `forward_iter` are untouched. Those users should set `output_nonlinearity=None` to make their code work again. I think this is an acceptable tradeoff (document this in the CHANGES.md).

BenjaminBossan:
Resolves #637, resolves #661

Supersedes #572 #580

Added a parameter `predict_nonlinearity` to `NeuralNet` which allows users to control the nonlinearity to be applied to the module output when calling `predict` and `predict_proba`. Also, when using `CrossEntropyLoss`, softmax is now automatically applied to the output.

This PR implements what was discussed in #661. Regarding the implementation, I used the most simple approach at the moment (no dispatching based on the net class, for instance). Also, I added a `get_predict_nonlinearity` method to `NeuralNet`, which handles the case of `'auto'` vs `None` vs `callable`.

Regarding the naming, I decided against `output_nonlinearity`. I believe `predict_nonlinearity` makes it much more obvious that this nonlinearity is **not** applied to the module output directly (which would affect loss etc.) but only when calling `predict` and `predict_proba`.

ping @qtux

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -206,12 +206,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         y_pred : numpy ndarray
 
         """
-        y_preds = []
-        for yp in self.forward_iter(X, training=False):
-            yp = yp[0] if isinstance(yp, tuple) else yp
-            y_preds.append(to_numpy(yp.max(-1)[-1]))
-        y_pred = np.concatenate(y_preds, 0)
-        return y_pred
+        return super().predict_proba(X).argmax(1)
 
 
 neural_net_binary_clf_doc_start = """NeuralNet for binary classification tasks
@@ -361,49 +356,3 @@ class NeuralNetBinaryClassifier(NeuralNet, ClassifierMixin):
         """
         y_proba = self.predict_proba(X)
         return (y_proba[:, 1] > self.threshold).astype('uint8')
-
-    # pylint: disable=missing-docstring
-    def predict_proba(self, X):
-        """Where applicable, return probability estimates for
-        samples.
-
-        If the module's forward method returns multiple outputs as a
-        tuple, it is assumed that the first output contains the
-        relevant information and the other values are ignored. If all
-        values are relevant, consider using
-        :func:`~skorch.NeuralNet.forward` instead.
-
-        Parameters
-        ----------
-        X : input data, compatible with skorch.dataset.Dataset
-          By default, you should be able to pass:
-
-            * numpy arrays
-            * torch tensors
-            * pandas DataFrame or Series
-            * scipy sparse CSR matrices
-            * a dictionary of the former three
-            * a list/tuple of the former three
-            * a Dataset
-
-          If this doesn't work with your data, you have to pass a
-          ``Dataset`` that can deal with the data.
-
-        Returns
-        -------
-        y_proba : numpy ndarray
-
-        """
-        y_probas = []
-        self.check_is_fitted(attributes=['criterion_'])
-        bce_logits_loss = isinstance(
-            self.criterion_, torch.nn.BCEWithLogitsLoss)
-
-        for yp in self.forward_iter(X, training=False):
-            yp = yp[0] if isinstance(yp, tuple) else yp
-            if bce_logits_loss:
-                yp = torch.sigmoid(yp)
-            y_probas.append(to_numpy(yp))
-        y_proba = np.concatenate(y_probas, 0)
-        y_proba = np.stack((1 - y_proba, y_proba), axis=1)
-        return y_proba
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -206,7 +206,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         y_pred : numpy ndarray
 
         """
-        return super().predict_proba(X).argmax(1)
+        return super().predict_proba(X).argmax(axis=1)
 
 
 neural_net_binary_clf_doc_start = """NeuralNet for binary classification tasks
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1041,7 +1041,7 @@ class NeuralNet:
             return self.module_(**x_dict)
         return self.module_(x, **fit_params)
 
-    def get_predict_nonlinearity(self):
+    def _get_predict_nonlinearity(self):
         """Return the nonlinearity to be applied to the prediction
 
         This can be useful, e.g., when
@@ -1066,7 +1066,6 @@ class NeuralNet:
         Returns
         -------
         nonlin : callable
-
           A callable that takes a single argument, which is a PyTorch
           tensor, and returns a PyTorch tensor.
 
@@ -1112,7 +1111,7 @@ class NeuralNet:
         y_proba : numpy ndarray
 
         """
-        nonlin = self.get_predict_nonlinearity()
+        nonlin = self._get_predict_nonlinearity()
         y_probas = []
         for yp in self.forward_iter(X, training=False):
             yp = yp[0] if isinstance(yp, tuple) else yp
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -166,7 +166,7 @@ class NeuralNet:
       PyTorch tensor, and return the transformed PyTorch tensor.
 
       This can be useful, e.g., when
-      :func:`~skorch.classifier.NeuralNetClassifier.predict_proba`
+      :func:`~skorch.NeuralNetClassifier.predict_proba`
       should return probabilities but a criterion is used that does
       not expect probabilities. In that case, the module can return
       whatever is required by the criterion and the
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -152,6 +152,8 @@ class NeuralNet:
       ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
       'train_loss'])``).
 
+    predict_nonlinearity : TODO
+
     warm_start : bool (default=False)
       Whether each fit call should lead to a re-initialization of the
       module (cold start) or whether the module should be trained
@@ -213,6 +215,7 @@ class NeuralNet:
             dataset=Dataset,
             train_split=CVSplit(5),
             callbacks=None,
+            predict_nonlinearity='auto',
             warm_start=False,
             verbose=1,
             device='cpu',
@@ -229,6 +232,7 @@ class NeuralNet:
         self.dataset = dataset
         self.train_split = train_split
         self.callbacks = callbacks
+        self.predict_nonlinearity = predict_nonlinearity
         self.warm_start = warm_start
         self.verbose = verbose
         self.device = device
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -22,6 +22,8 @@ from skorch.dataset import uses_placeholder_y
 from skorch.exceptions import DeviceWarning
 from skorch.history import History
 from skorch.setter import optimizer_setter
+from skorch.utils import _identity
+from skorch.utils import _infer_predict_nonlinearty
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import TeeGenerator
 from skorch.utils import check_is_fitted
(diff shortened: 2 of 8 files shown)


skorch/net.py
```diff
@@ -152,6 +154,31 @@ class NeuralNet:
...
+      (softmax for :class:`~torch.nn.CrossEntropyLoss` and sigmoid for
+      :class:`~torch.nn.BCEWithLogitsLoss`). If it cannot be inferred
+      or if the parameter is None, just use the identity function.
+
+      In case a callable is passed, it should accept the output of the
+      module (the first output if there is more than one), which is a
+      PyTorch tensor, and return the transformed PyTorch tensor.
+
+      This can be useful, e.g., when
+      :func:`~skorch.classifier.NeuralNetClassifier.predict_proba`
```
thomasjpfan, line 169: I think we can do this since `NeuralNetClassifier` is imported in the top level.

```suggestion
      :func:`~skorch.NeuralNetClassifier.predict_proba`
```

skorch/tests/test_utils.py
```diff
@@ -669,3 +669,69 @@ def list_gen():
...
+
+    def test_infer_neural_net_classifier_crossentropy_loss(
+            self, infer_predict_nonlinearity, net_clf_cls, module_cls):
+        # CrossEntropyLoss should return valid probabilities
+        net = net_clf_cls(module_cls, criterion=torch.nn.CrossEntropyLoss).initialize()
+        fn = infer_predict_nonlinearity(net)
+
+        X = torch.rand((20, 5))
+        out = fn(X).numpy()
+        assert np.allclose(out.sum(1), 1.0)
```
thomasjpfan, line 713: Nit:

```suggestion
        assert np.allclose(out.sum(axis=1), 1.0)
```

skorch/tests/test_utils.py
```diff
@@ -669,3 +669,69 @@ def list_gen():
...
+            self, infer_predict_nonlinearity, net_bin_clf_cls, module_cls):
+        # BCEWithLogitsLoss should return valid probabilities
+        net = net_bin_clf_cls(module_cls).initialize()
+        fn = infer_predict_nonlinearity(net)
+
+        X = torch.rand(20)  # binary classifier returns 1-dim output
+        X = 10 * X - 5.0  # random values from -5 to 5
+        out = fn(X).numpy()
+        assert out.shape == (20, 2)  # output should be 2-dim
+        assert np.allclose(out.sum(1), 1.0)
```
thomasjpfan, line 726:
```suggestion
        assert np.allclose(out.sum(axis=1), 1.0)
```

skorch/net.py
```diff
@@ -1010,6 +1039,46 @@ def infer(self, x, **fit_params):
             return self.module_(**x_dict)
         return self.module_(x, **fit_params)
 
+    def get_predict_nonlinearity(self):
```
thomasjpfan, line 1042: With this being public there will be two ways to adjust the nonlinearity:

1. With the `predict_nonlinearity` parameter in init.
2. Subclassing

Are we okay with this?

skorch/classifier.py
```diff
@@ -206,12 +206,7 @@ def predict(self, X):
         y_pred : numpy ndarray
 
         """
-        y_preds = []
-        for yp in self.forward_iter(X, training=False):
-            yp = yp[0] if isinstance(yp, tuple) else yp
-            y_preds.append(to_numpy(yp.max(-1)[-1]))
-        y_pred = np.concatenate(y_preds, 0)
-        return y_pred
+        return super().predict_proba(X).argmax(1)
```
thomasjpfan, line 209:
```suggestion
        return super().predict_proba(X).argmax(axis=1)
```

skorch/tests/test_net.py
```diff
@@ -2540,6 +2540,92 @@ def forward(self, y_pred, _):
...
+        # don't want callbacks to trigger side effects
+        net.partial_fit(X, y)
+        assert not side_effect
+
+        # 2 calls, since batch size == 128 and n == 200
+        y_proba = net.predict_proba(X)
+        assert len(side_effect) == 2
+        assert side_effect[0].shape == (128, 2)
+        assert side_effect[1].shape == (72, 2)
+        assert (y_proba == 0).all()
```
thomasjpfan, line 2590: Nit since `y_proba` is floating point.

```suggestion
        assert np.allclose(y_proba, 0)
```

(Yes `==0` works because this was created with `np.zeros_like`)

skorch/net.py
BenjaminBossan: Good question, I was also unsure about this. The `predict_nonlinearity` parameter specifies _what_ nonlinearity I want, whereas the `get_predict_nonlinearity` specifies _how_ the nonlinearity selection is resolved. I would also be okay with making the latter private (with the possibility to make it public later if the need ever arises).

thomasjpfan: I would go with option 1 for now and having this function be public.

If a user were to override `_get_predict_nonlinearity`, we would not officially support it and it may break in the future.

skorch/net.py
```diff
@@ -152,6 +154,31 @@ class NeuralNet:
       ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
       'train_loss'])``).
 
+    predict_nonlinearity : callable, None, or 'auto' (default='auto')
```
thomasjpfan, line 157: One minor concern: if one places a lambda function in `predict_nonlinearity`, it would not be pickle-able. Do you think its mentioning this in the docstring?

skorch/tests/test_net.py
```diff
@@ -2540,6 +2540,92 @@ def forward(self, y_pred, _):
...
+            module_cls,
+            max_epochs=1,
+            criterion=nn.CrossEntropyLoss,
+            predict_nonlinearity=None,
+        ).initialize()
+
+        rv = np.random.random((20, 5))
+        net.forward_iter = lambda *args, **kwargs: (torch.as_tensor(rv) for _ in range(2))
+
+        # 2 batches, mock return value hs shape 20,5 thus y_proba has
```
thomasjpfan, line 2611:
```suggestion
        # 2 batches, mock return value has shape 20,5 thus y_proba has
```

skorch/net.py
BenjaminBossan: Just to be sure, option 1 for you means making `get_predict_nonlinearity` private?

skorch/tests/test_net.py
```diff
@@ -2540,6 +2540,92 @@ def forward(self, y_pred, _):
...
+            return np.zeros_like(X)
+
+        X, y = data[0][:200], data[1][:200]
+        net = net_cls(
+            module_cls, max_epochs=1, predict_nonlinearity=nonlin).initialize()
+
+        # don't want callbacks to trigger side effects
+        net.callbacks_ = []
+        net.partial_fit(X, y)
+        assert not side_effect
```
ottonemo, line 2557: But wouldn't we expect callbacks such as accuracy scoring to call `predict`?

skorch/tests/test_net.py
```diff
@@ -2540,6 +2540,92 @@ def forward(self, y_pred, _):
...
+        assert side_effect[1].shape == (72, 2)
+        assert np.allclose(y_proba, 0)
+
+        net.predict_proba(X)
+        assert len(side_effect) == 4
+
+    def test_predict_nonlinearity_none(
+            self, net_cls, module_cls, data):
+        # even though we have CrossEntropyLoss, we don't want the
+        # output from predict_proba to be modified, since we set
```
ottonemo, line 2598:
```suggestion
        # output from predict_proba to be modified, thus we set
```

skorch/tests/test_utils.py
```diff
@@ -803,3 +803,69 @@ def list_gen():
...
+        net = net_clf_cls(module_cls).initialize()
+        fn = infer_predict_nonlinearity(net)
+
+        X = np.random.random((20, 5))
+        out = fn(X)
+        assert out is X
+
+    def test_infer_neural_net_classifier_crossentropy_loss(
+            self, infer_predict_nonlinearity, net_clf_cls, module_cls):
+        # CrossEntropyLoss should return valid probabilities
```
ottonemo, line 841:
```suggestion
        # CrossEntropyLoss criteron: nonlinearity should return valid probabilities
```

skorch/tests/test_utils.py
```diff
@@ -803,3 +803,69 @@ def list_gen():
...
+        fn = infer_predict_nonlinearity(net)
+
+        X = torch.rand((20, 5))
+        out = fn(X).numpy()
+        assert np.allclose(out.sum(axis=1), 1.0)
+        assert ((0 <= out) & (out <= 1.0)).all()
+
+    def test_infer_neural_binary_net_classifier_default(
+            self, infer_predict_nonlinearity, net_bin_clf_cls, module_cls):
+        # BCEWithLogitsLoss should return valid probabilities
```
ottonemo, line 852:
```suggestion
        # BCEWithLogitsLoss criterion: nonlinearity should return valid probabilities
```

skorch/utils.py
```diff
@@ -552,6 +554,58 @@ def check_is_fitted(estimator, attributes, msg=None, all_or_any=all):
...
+    """Return input as is, the identity operation"""
+    return x
+
+
+def _sigmoid_then_2d(x):
+    """Transform 1-dim logits to valid y_proba
+
+    Sigmoid is applied to x to transform it to probabilities. Then
+    concatenate the probabilities with 1 - these probabilities to
+    return a correctly formed ``y_proba``.
```
ottonemo, line 567: Maybe specify who expects this format. Something along the lines of `formats outputs for use with BCE loss` or something like that

skorch/tests/test_net.py
BenjaminBossan: That's why I removed all callbacks two lines earlier, otherwise the test becomes very messy.

skorch/tests/test_classifier.py
```diff
@@ -289,7 +289,12 @@ def test_custom_loss_does_not_call_sigmoid(
         mock = Mock(side_effect=lambda x: x)
         monkeypatch.setattr(torch, "sigmoid", mock)
 
-        net = net_cls(module_cls, max_epochs=1, lr=0.1, criterion=nn.MSELoss)
+        # we need to add a custom nonlinearity, otherwise the output won't be 2d
```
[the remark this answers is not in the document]
BenjaminBossan: Fixed

skorch/tests/test_net.py
BenjaminBossan: Fixed

skorch/tests/test_utils.py
BenjaminBossan: Fixed

BenjaminBossan:
@ottonemo I addressed your comments.

Issue #628, opened 2020-05-04
skorch.NeuralNet auto score method given criterion


hsorsky:
For ``torch.nn.Module``s passed to ``skorch.NeuralNet``, I think it would make sense to automatically use the passed criterion in the score function if no score is implemented. E.g.

```python
class MyModule(torch.nn.Module):
    def __init__(self, ...):
        ...

class MyLoss(_Loss):
    def __init__(self, ...):
        ...

net = NeuralNet(
    module=MyModule,
    ...
    criterion=MyLoss,
)

net.fit(X, y) # fits by minimising MyLoss's forward method
net.score(X, y) # scores by evaluating MyLoss's forward method
```

BenjaminBossan:
That's an interesting suggestion, since it's not super straightforward to achieve this result at the moment. I would probably not go so far as to make this the default behavior, since it could have some backwards-incompatible consequences (for instance, [`check_scoring`](https://github.com/scikit-learn/scikit-learn/blob/02309ffbdaae45af75b0e87946aab9aedb6b3634/sklearn/metrics/_scorer.py#L420) checks if an estimator has the `score` method, which our nets don't).

A possibility I could see is through a mixin class. The idea would be that I can opt into this by defining:

```python
class MyNet(NeuralNet, ScoringMixin):
    """Gives NeuralNet a score method"""
```

This mixin could reside in the skorch helpers.

Another option would be through some kind of argument (`net = NeuralNet(..., auto_score=True)`) but I don't like adding methods dynamically if there are other, less "magical" ways.

@hsorsky Do you want to work on this?

asberk:
@BenjaminBossan Could you explain what you mean re backwards-incompatible consequences? 

I'm new to using `skorch`. My team is adopting it, so I'm trying to make the transition from `ignite`. So far, I'm confused that the easiest "skorchy" way to get the loss on a dataset `(X, y)` seems to one of:
```python
# this
net.criterion_(nnbc.evaluation_step(X), skorch.utils.to_tensor(y, torch.device('cpu'))).numpy()
# or this
nnbc.get_loss(nnbc.forward(X), torch.tensor(y).float()).numpy()
```
Am I missing something? This seems like too basic a functionality not to have a shorter expression for (especially given that `net.predict` already exists). 

Anyway, if such functionality does not already exist, would you be able to describe in slightly more detail how the change should be implemented so as to avoid the troubles you alluded to above? I would be happy to work on this issue.

[...]

PR #665, opened 2020-07-06
Implement loss scoring for #628

asberk:
This addresses issue #628 . Tests could probably be more complete. @BenjaminBossan , please let me know if/how they could be fleshed out to be more comprehensive. 

As a side note, I have limited familiarity with `PassThroughScoring`, but it seemed, in a way, like overkill. This could be because I'm not properly understanding some edge cases. Anyway, at the time, I thought it was sensible to write an accumulator class that handles the reduction of the batch-level statistics for `criterion`. In case this isn't ideal, is there an easy way to convert the current code to something that uses Skorch tools (*e.g.,* `PassThroughScoring`)?

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -357,3 +357,50 @@ layer, and that the number of features can be determined by
 ``X.shape[1]``. If those assumptions are not true for your case,
 adjust the code accordingly. A fully working example can be found
 on `stackoverflow <https://stackoverflow.com/a/60170023/1643939>`_.
+
+How do I implement a score method on the net that returns the loss?
+-------------------------------------------------------------------
+
+Sometimes, it is useful to be able to compute the loss of a net from within
+``skorch`` (*e.g.,* when a net is part of an ``sklearn`` pipeline). The function
+:func:`skorch.scoring.loss_scoring` achieves this. Two examples are provided
+below. The first demonstrates how to use :func:`skorch.scoring.loss_scoring` as
+a function on a trained ``net`` object.
+
+.. code:: python
+
+    from skorch.scoring import loss_scoring
+
+    X = np.random.randn(250, 25).astype('float32')
+    y = (X.dot(np.ones(25)) > 0).astype(int)
+
+    module = nn.Sequential(
+        nn.Linear(25, 25),
+        nn.ReLU(),
+        nn.Linear(25, 2),
+        nn.Softmax(dim=1)
+    )
+    net = skorch.NeuralNetClassifier(module).fit(X, y)
+    print(loss_scoring(net, X, y))
+
+The second example shows how to sub-class :class:`skorch.NeuralNetClassifier` to
+implement a ``score`` method. In this example, the ``score`` method returns the
+**negative** of the loss value, because we want
+:class:`sklearn.model_selection.GridSearchCV` to return the run with **least**
+loss and :class:`sklearn.model_selection.GridSearchCV` searches for the run with
+the **greatest** score.
+
+.. code:: python
+
+    class ScoredNet(skorch.NeuralNetClassifier):
+        def score(self, X, y=None):
+            loss_value = loss_scoring(self, X, y)
+            return -loss_value
+    
+    net = ScoredNet(module)
+    grid_searcher = GridSearchCV(
+        net, {'lr': [1e-2, 1e-3], 'batch_size': [8, 16]},
+    )
+    grid_searcher.fit(X, y)
+    best_net = grid_searcher.best_estimator_
+    print(best_net.score(X, y))
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -383,7 +383,7 @@ a function on a trained ``net`` object.
     net = skorch.NeuralNetClassifier(module).fit(X, y)
     print(loss_scoring(net, X, y))
 
-The second example shows how to sub-class :class:`skorch.NeuralNetClassifier` to
+The second example shows how to sub-class :class:`skorch.classifier.NeuralNetClassifier` to
 implement a ``score`` method. In this example, the ``score`` method returns the
 **negative** of the loss value, because we want
 :class:`sklearn.model_selection.GridSearchCV` to return the run with **least**
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -362,7 +362,7 @@ How do I implement a score method on the net that returns the loss?
 -------------------------------------------------------------------
 
 Sometimes, it is useful to be able to compute the loss of a net from within
-``skorch`` (*e.g.,* when a net is part of an ``sklearn`` pipeline). The function
+``skorch`` (e.g. when a net is part of an ``sklearn`` pipeline). The function
 :func:`skorch.scoring.loss_scoring` achieves this. Two examples are provided
 below. The first demonstrates how to use :func:`skorch.scoring.loss_scoring` as
 a function on a trained ``net`` object.
--- a/skorch/scoring.py
+++ b/skorch/scoring.py
@@ -3,71 +3,43 @@ from skorch.net import NeuralNet
 from skorch.dataset import unpack_data
 
 
-class _CriterionAccumulator:
-    def __init__(self, criterion):
-        """
-        _CriterionAccumulator(criterion)
-
-        Parameters
-        ----------
-        criterion: PyTorch loss function
-            e.g., nn.MSELoss or nn.NLLLoss
-        """
-        self.criterion = criterion
-        reduction = self._get_reduction()
-        assert reduction in ["mean", "sum", "none",], (
-            "Expected reduction to be one of 'mean', 'sum' "
-            + "or 'none' but got {reduction}.".format(reduction=reduction)
-        )
-        self.log = {"loss": [], "batch_size": []}
-        self._reduce = {
-            "sum": self._reduce_sum,
-            "mean": self._reduce_mean,
-            "none": self._reduce_none,
-        }
-
-    def __call__(self, input, target):
-        loss = self.criterion(input, target)
-
-        batch_size = target.size(0)
-        self.log["batch_size"].append(batch_size)
-
-        # If criterion.reduction is 'none', loss.item() can fail,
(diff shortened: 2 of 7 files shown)


BenjaminBossan:
Thanks a lot @asberk for the PR.

I haven't taken a detailed look at your code yet but two things came to my mind:

1. The docstring of `loss_scoring` needs some clean up, I think it should stick very close to other similar functions. Also, it would be nice to document how to use that function to add a `score` method to the net, similar to the discussion. That could be added to the docstring of `loss_scoring` as well, or maybe in the FAQ section of the docs.

2. I believe from a first glance that the accumulator class is overkill. Shouldn't it be possible to capture the same functionality without a class? The `loss_scoring` could collect the losses and the batch sizes, then pass them to an accumulation function, along with the criterion. Inside that function, you could call `np.average`, similar to how `PassthroughScoring` does it.

I don't think there is a need to use `PassthroughScoring` or to try to share any code with `PassthroughScoring`, the code duplication should be minimal. What do you think about this?

Overall, this looks very good though, nice unit tests.

asberk:
Thank you for your comments, @BenjaminBossan . 

1. I've fixed up the docstring and added an example. The example is not an exemplar as `GridSearchCV` looks for higher scores I believe. If you suggest an avenue that more simply showcases the behaviour, I'll update it to that.
2. I've removed the accumulator abstraction. There is now also a test to make sure `loss_scoring` computes the same value as `net.criterion_(y_proba, labels)`.

skorch/utils.py
```diff
@@ -544,6 +544,7 @@ class TeeGenerator:
     generator more than once.
 
     """
+
```
BenjaminBossan, line 547: Could you please undo this change? I think it was accidentally checked in.

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
...
+    ... )
+    >>> grid_searcher.fit(X, y)
+    >>> best_net = grid_searcher.best_estimator_
+    >>> print(best_net.score(X, y))
+
+    Parameters
+    ----------
+    net : skorch.NeuralNet
+        A fitted Skorch :class:`.NeuralNet` object.
+    X : input data, compatible with skorch.dataset.Dataset
```
BenjaminBossan, line 45: Could you please add an empty line between parameters?

skorch/tests/test_scoring.py
```diff
@@ -0,0 +1,91 @@
...
+import torch
+
+
+class TestLossScoring:
+    @pytest.fixture(scope="module")
+    def data(self, classifier_data):
+        return classifier_data
+
+    @pytest.fixture(scope="module")
+    def val_data(self, classifier_data):
```
BenjaminBossan, line 12: I think for testing this function, we don't actually need separate validation data, we can just use the training data.

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
+import numpy as np
+from skorch.net import NeuralNet
+from skorch.dataset import unpack_data
+
+
+def loss_scoring(net: NeuralNet, X, y=None):
+    """
```
BenjaminBossan, line 7:
```suggestion
    """Calculate score using the criterion of the net
    
    Use the exact same logic as during model training to calculate the score.
```

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
...
+    loss_scoring(net, X, y=None)
+
+    Computes the loss of ``net`` on data (``X``, ``y``).
```
BenjaminBossan, line 10:
```suggestion
```

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
...
+
+    This function can be used to implement the ``score`` method for a
+    :class:`.NeuralNet` through sub-classing. This is useful, for example, when
+    combining skorch models with sklearn objects that rely on the model's
+    ``score`` method. Below is an example using GridSearchCV.
+
+    >>> class ScoredNet(skorch.NeuralNetClassifier):
```
BenjaminBossan, line 17: Thanks for adding the elaborate example. I think, however, that it could be simplified. 

1) You could just show how to subclass the net here.
2) You could go with the grid search example, but in that case you actually don't need to subclass the net. You can pass the `scoring=loss_scoring` to `GridSearchCV` and it should result in the same outcome.

Assume that you go with the second option: I would then add the subclassing example to the docs, probably to the FAQ ("How to implement a score method on the net that returns the loss?" or something along these lines).

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
...
+    ...     def score(self, X, y=None, lower_is_better=False):
```
BenjaminBossan, line 18:
```suggestion
    ...     def score(self, X, y=None):
```

I would remove this because it doesn't conform to the sklearn API. Also, if `lower_is_better = True`, which is always true for PyTorch criteria, we should return `-loss_value` and not vice versa? Otherwise, something like grid search would choose the model with the _highest_ loss.

skorch/scoring.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Returns
+    -------
+    loss_value : float32 or np.ndarray
+        Return type depends on net.criterion_.reduction, and will be a float if
+        reduction is ``'sum'`` or ``'mean'``. If reduction is ``'none'`` then
+        this function returns a ``np.ndarray`` object.
+
+    """
+    net.check_is_fitted()
```
BenjaminBossan, line 68: Good attention to detail

skorch/scoring.py
asberk: I wound up with a combination of 1. and 2. The docstring now shows the sub-classing. The FAQ now has an example of `loss_scoring(net, X, y)` as well as an example using `GridSearchCV`.

asberk:
I believe I have addressed the changes you've mentioned, @BenjaminBossan . Thank you for the detailed list and comments!

Funny enough, there were tests for `reduction` ∈ {`'mean'` `'sum'`} in that class that no longer exists, and I forgot to transfer that over. The `reduction` fixture uses `'mean'` and `'sum'` only, and `test_scored_net_matches_criterion_value` tests score using the same `reduction` that was used to train the network. This could be modified if you think it is insufficient: so that a trained net's scoring is tested for both `'mean'` and `'sum'`. 

I made a separate test for when the reduction is `'none'` since the behaviour is so different, and since the reduction must be altered after fitting. There is also a test asserting that an unknown reduction raises.

docs/user/FAQ.rst
```diff
@@ -357,3 +357,50 @@ layer, and that the number of features can be determined by
 ``X.shape[1]``. If those assumptions are not true for your case,
 adjust the code accordingly. A fully working example can be found
 on `stackoverflow <https://stackoverflow.com/a/60170023/1643939>`_.
+
+How do I implement a score method on the net that returns the loss?
+-------------------------------------------------------------------
+
+Sometimes, it is useful to be able to compute the loss of a net from within
+``skorch`` (*e.g.,* when a net is part of an ``sklearn`` pipeline). The function
```
BenjaminBossan, line 365:
```suggestion
``skorch`` (e.g. when a net is part of an ``sklearn`` pipeline). The function
```
Or why should "e.g." be in italics?

docs/user/FAQ.rst
```diff
@@ -357,3 +357,50 @@ layer, and that the number of features can be determined by
...
+    module = nn.Sequential(
+        nn.Linear(25, 25),
+        nn.ReLU(),
+        nn.Linear(25, 2),
+        nn.Softmax(dim=1)
+    )
+    net = skorch.NeuralNetClassifier(module).fit(X, y)
+    print(loss_scoring(net, X, y))
+
+The second example shows how to sub-class :class:`skorch.NeuralNetClassifier` to
```
BenjaminBossan, line 386:
```suggestion
The second example shows how to sub-class :class:`skorch.classifier.NeuralNetClassifier` to
```

At least for me that was necessary to make the link work.

skorch/scoring.py
```diff
@@ -0,0 +1,82 @@
...
+            "sample_weight for loss_scoring is not yet supported."
+        )
+
+    net.check_is_fitted()
+
+    dataset = net.get_dataset(X, y)
+    iterator = net.get_iterator(dataset, training=False)
+    history = {"loss": [], "batch_size": []}
+    reduction = net.criterion_.reduction
+    assert reduction in ["mean", "sum", "none",], (
```
BenjaminBossan, line 63: Could you please raise a `ValueError` instead?

skorch/tests/test_scoring.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_scored_net_with_reduction_none(
+        self, scored_net_cls, module_cls, reduction, data
+    ):
+        X, y = data
+        net = scored_net_cls(
+            module_cls, lr=0.01, criterion__reduction=reduction
+        ).fit(X, y)
+        net.set_params(criterion__reduction="none")
+        output = net.score(X, y)
+        assert output.shape[0] == X.shape[0]
```
BenjaminBossan, line 106: I believe this test can be made a little bit more sophisticated than just checking the shape. If you first calculate the loss with `reduction='sum'` and then with `'none'`, the sum of the latter should equal the former.

docs/user/FAQ.rst
asberk: I'd italicized it because it's Latin, not English. But it does not matter on the whole, and I admit it is better to opt for consistency. I suppose better readability of the plaintext is a bonus of no italics.

skorch/tests/test_scoring.py
asberk: Good idea. I kept the `assert` for `shape` equality, and added the extra lines for asserting value equality.

docs/user/FAQ.rst
BenjaminBossan: Okay, I didn't know that. As you said, for consistency (and since it's easier to type), let's do without.

BenjaminBossan:
Thanks @asberk for this feature, you did excellent work. I'm sure many users will find it useful.

asberk:
Thank you @BenjaminBossan for the learning opportunities. Happy to contribute.

PR #675, opened 2020-07-30
Add PyTorch version 1.6.0 to .travis.yml

BenjaminBossan:
Also document the support of that version.

@ottonemo tested PyTorch 1.6.0 with CUDA and it worked, so this should also work without further modification.

I wonder if we should start dropping support for older PyTorch versions. The main reason would be to accelerate CI. What are your thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 dist: xenial
 python:
-  - "3.7.5"
+  - "3.7"
   - "3.6"
   - "3.5"  # TODO: drop support in 0.10.0
 env:
@@ -11,6 +11,10 @@ env:
   - PYTORCH_VERSION="1.4.0"
   - PYTORCH_VERSION="1.5.0"
   - PYTORCH_VERSION="1.6.0"
+jobs:
+  exclude:
+  - python: "3.5"
+    env: PYTORCH_VERSION="1.6.0"
 
 cache:
   apt: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,9 @@ python:
   - "3.6"
   - "3.5"  # TODO: drop support in 0.10.0
 env:
-  - PYTORCH_VERSION="1.1.0"
-  - PYTORCH_VERSION="1.2.0"
   - PYTORCH_VERSION="1.3.1"
   - PYTORCH_VERSION="1.4.0"
-  - PYTORCH_VERSION="1.5.0"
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
I like the idea @thomasjpfan 

It would be helpful to know how many people still rely on old PyTorch versions. My impression is that the developers did a great job in keeping breaking changes to a minimum, which makes me think that most people transition fast. But in the end only some PyTorch devs might know for sure. Perhaps @zou3519 can weigh in?

zou3519:
PyTorch does around 4 major releases a year (e.g., 1.3, 1.4, 1.5, 1.6). In between the major releases there are minor bugfix releases (e.g 1.5.1, 1.4.1).

I don't have full historical stats, but just looking at some [publicly available PyPI download stats](https://packaging.python.org/guides/analyzing-pypi-package-downloads/), it looks like 90% of downloads in the past month were for PyTorch 1.1 and newer.

BenjaminBossan:
Thanks a lot for the reply. 10% of downloads being for versions < 1.1 seems to be a significant amount. I would, however, assume that a lot of those are automatic crawlers for pypi mirrors.

Therefore, I could see us going to the last 4 releases. That doesn't mean skorch would not work with older PyTorch versions, since we rely on stable parts of the API, just that we don't test them.

Another idea could be to check 1.1 + the last 3 or 4 releases, leaving a gap in between, with the assumption that if 1.1 works and 1.3 works, 1.2 probably works as well.

ottonemo:
As far as I know there is no 'special' stable release, therefore to me it would not make much sense to explicitly test for 1.1 (it just seems arbitrary to me). I would favor the last 4 releases instead.

BenjaminBossan:
I agree that 1.1 is arbitrary (I don't even remember why we don't include 1.0). So last 4 releases it is? I.e. 1.3.1, 1.4, 1.5, 1.6. We could also go with 1.4.1 and 1.5.1 instead of 1.4 and 1.5.

ottonemo:
The last 4 minor versions (latest bugfix release) I think. So 1.3.1, 1.4.1, 1.5.1, 1.6.0.

thomasjpfan:
+1 on the last four versions

BenjaminBossan:
Okay, I updated the .travis.yml (and docs), we now support the latest 4 minor versions of PyTorch. There was no 1.4.1 in the conda channel, so that one is still 1.4.0.

BenjaminBossan:
Are we good to go here?

PR #676, opened 2020-08-01
Improve documentation of verbose argument

### LINKED ISSUE #660: Improve documentation of verbose parameter
Currently, the docstring for `verbose` simply says:

```
verbose : int (default=1)
    Control the verbosity level.
```

It's very incomplete.

BenjaminBossan:
Fixes #660


skorch/net.py
```diff
@@ -188,7 +188,11 @@ class NeuralNet:
       further (warm start).
 
     verbose : int (default=1)
-      Control the verbosity level.
+      This parameter controls how much print output is generated by
+      the net and its callbacks. By setting this value to 0, e.g. the
+      summary scores at the end of each epoch are no longer printed
+      (though they are still logged in the ``history`` attribute).
+      This can be useful when running a hyperparameter search.
```
thomasjpfan, line 195: I think "This" in this context means "setting this value to 0" (and does not mean the logging to history attribute) Maybe:

```
summary scores at the end of each epoch are no longer printed.
This can be useful when running a hyperparameter search. The
summary scores are still logged in the history attribute
regardless of the verbose setting.
```

BenjaminBossan: Thanks for the suggestion. I changed the text accordingly, with only a minor tweak.

PR #677, opened 2020-08-01
Better document possible input data

### LINKED ISSUE #672: Is it possible to use multiple inputs with skorch?
Hello
skorch is very useful and it saves me the time of writing training code.

Recently, I need to build a model with multiple inputs, e.g. model(X1, X2, X3). I would like to know there is any way to train this kind of model with skorch.

I read the document and it seems that skorch currently only supports (X, y) .

Thank you.

BenjaminBossan:
Fixes #672 

Until now, there was no one place where the different types of
possible input data were documented. Now there is an additional
section in the docs that covers that. Specifically, it talks about
regular data, Datasets, and multiple inputs.

### DIFF
--- a/docs/user/neuralnet.rst
+++ b/docs/user/neuralnet.rst
@@ -388,6 +388,83 @@ e.g. :class:`.Checkpoint`. Those should be used if you need to have
 your model saved or loaded at specific times, e.g. at the start or end
 of the training process.
 
+Input data
+----------
+
+Regular data
+^^^^^^^^^^^^
+
+skorch supports numerous input types for data. Regular input types
+that should just work are numpy arrays, torch tensors, scipy sparse
+CSR matrices, and pandas DataFrames (see also
+:class:`~skorch.helper.DataFrameTransformer`).
+
+Typically, your task should involve an ``X`` and a ``y``. If you're
+dealing with a task that doesn't require a target (say, training an
+autoencoder), you can just pass ``y=None``. Make sure your loss
+function deals with this appropriately.
+
+Datasets
+^^^^^^^^
+
+:class:`~torch.utils.data.Dataset`\s are also supported, with the
+requirement that they should return exactly two items (``X`` and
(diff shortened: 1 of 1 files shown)


docs/user/neuralnet.rst
```diff
@@ -388,6 +388,83 @@ e.g. :class:`.Checkpoint`. Those should be used if you need to have
...
 
+Input data
+----------
+
+Regular data
+^^^^^^^^^^^^
+
+skorch supports numerous input types for data. Regular input types
+that should just work are numpy arrays, torch tensors, scipy sparse
+CSR matrices, and pandas DataFrames (see also
```
thomasjpfan, line 399: Wow, didn't know we had support for sparse matrices.

BenjaminBossan: Yes :) But note that on the batch level, we cast to sparse. The reason is that at the time of implementation, fundamental operations on sparse matrices were not supported by PyTorch, making it all but impossible to use for typical neural net stuff. I checked just now and there seems to be pretty extensive support for [sparse COO tensors](https://pytorch.org/docs/stable/sparse.html), so maybe we can revisit this again.

docs/user/neuralnet.rst
```diff
@@ -388,6 +388,83 @@ e.g. :class:`.Checkpoint`. Those should be used if you need to have
...
+Multiple input arguments
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In some cases, the input actually consists of multiple inputs. E.g.,
+in a text classification task, you might have an array that contains
+the integers representing the tokens for each sample, and another
+array containing the number of tokens of each sample. skorch has you
+covered here as well.
+
+You could a list or tuple with all your inputs (``net.fit([tokens,
```
ottonemo, line 432:
```suggestion
You could supply a list or tuple with all your inputs (``net.fit([tokens,
```

Issue #681, opened 2020-08-13
Predefined split broken (TypeError: _make_split() got multiple values for argument 'valid_ds')


Dewald928:
I am currently working on the 0.8.1.dev0 version, because I want the LR scheduler on batch training.
But when I updated from 0.8.0 I get the following error when I use a predefined dataset in the train_split. (I didn't change anything except from the update)

`TypeError: _make_split() got multiple values for argument 'valid_ds'`

I know a few things changed with the unsupervised inputs, but I can't seem to figure out what to do to fix this. 
Is there an extra parameter I need to add somewhere or something I'm missing?

BenjaminBossan:
Could you check if the breakage could have been caused by #646? That's the only recent change to the train/valid splitting that comes to my mind, and as the discussion suggests, it might break existing code.

Dewald928:
Yes it is caused by #646 
I checked out the commit #626 everything was still working there.

The issue happens only when I pass a valid_ds to the trainsplit parameter

BenjaminBossan:
Could you please provide a minimal code example to reproduce the error?

Dewald928:
Even if I just run the simple snippet below, I get the same error.


```
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from skorch import NeuralNetClassifier
from skorch.dataset import Dataset
from skorch.helper import predefined_split
from skorch.toy import make_classifier

X, y = make_classification(n_samples=10000, n_features=2, n_redundant=0,
                           n_informative=1, n_clusters_per_class=1,
                           weights=[0.1, 0.9], class_sep=1)

X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.33,
                                                    random_state=1)

valid_ds = Dataset(X_test, y_test)

model = make_classifier(**{"input_units": 2})

net = NeuralNetClassifier(
    model,
    max_epochs=10,
    lr=0.1,
    train_split=predefined_split(valid_ds),
)

net.fit(X, y)
y_proba = net.predict_proba(X)
```

Dewald928:
I still seem to get the error. 

I tried checking out the branch with

```pip install git+https://github.com/skorch-dev/skorch.git@bugfix/predefined_split-bug-caused-by-646``

since I'm using conda. Don't know if that is good practice.

Perhaps someone could just validate, perhaps another issue on my side.

BenjaminBossan:
Hmm, that's strange. I see two possibilities here:

[...]

PR #683, opened 2020-08-16
Prepare for release of skorch v0.9.0

BenjaminBossan:
Steps 1-8 of the [release list](https://github.com/skorch-dev/skorch/wiki/Release).

No deprecations left for this version.

Before releasing, I would wait for #681 to be merged, as it fixes a bug
introduced in #646, i.e. after the last release.


BenjaminBossan:
FYI, I used the "release text" as the commit message. I propose to always do this in the future. That way, the text is in the git history, in case we ever want to move away from github.

PR #686, opened 2020-08-30
Drop official python 3.5 support

BenjaminBossan:
Comes down to no longer integrating it in CI. Add Python 3.8 though.

Also: Remove a FutureWarning in LRScheduler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: xenial
+dist: focal
 python:
   - "3.8"
   - "3.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,10 @@ python:
   - "3.7"
   - "3.6"
 env:
-  - PYTORCH_VERSION="1.3.1"
   - PYTORCH_VERSION="1.4.0"
   - PYTORCH_VERSION="1.5.1"
   - PYTORCH_VERSION="1.6.0"
-jobs:
-  exclude:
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
NB: Merging this PR is not urgent, since we just had the 0.9.0 release. But it should be done before the 0.10.0 release.

thomasjpfan:
We likely want to keep python 3.5 in the CI around for bug fixes in `0.9.x` and merge right before `0.10.0` gets released.

BenjaminBossan:
> We likely want to keep python 3.5 in the CI around for bug fixes in `0.9.x` and merge right before `0.10.0` gets released.

I wouldn't merge too late either because we want to catch bugs with Python 3.8 (or add it in a separate PR, but then CI will be very slow).

ottonemo:
We should also think about updating the base distribution from xenial to focal (or at least bionic).
I don't think xenial as a base distribution is representative for the developer machines out there.

BenjaminBossan:
> We should also think about updating the base distribution from xenial to focal

Good point, updated to focal.

BenjaminBossan:
I added PyTorch 1.7.0 to travis while removing 1.3.1.

BenjaminBossan:
@ottonemo This is pretty much obsolete after moving to github actions. Still some of the changes here should be moved to a new PR, right?

BenjaminBossan:
Superseded by move to github actions and #733

Issue #696, opened 2020-09-23
Question: Custom optimizer for NeuralNetClassifier


supetronix:
Hi,

I looked through documentation, code samples and StackOverflow but I didn't find any easy/quick solution (besides subclassing NuralNet) to use a custom optimizer. 

**Case:**
Instead of using a PyTorch optimizer class name, e.g.

`optimizer=torch.optim.Adam,`

I want to use a custom "Lookahead" optimizer that will take "Adam" as input.

Code sample for PyTorch:
```
        adam_optimizer = torch.optim.Adam(self.parameters(), lr = self.cfg.lr, 
                                     weight_decay=self.cfg.weight_decay)
        optimizer = Lookahead(optimizer=adam_optimizer,k=5,alpha=0.5)
```

code for lookahead optimizer is from:
[Lookahead github](https://github.com/lonePatient/lookahead_pytorch)

any help with easy solution will be appreciated. Or it could be added as a future enhancement to allow passing custom optimizer.

BenjaminBossan:
The problem here is not that you want to pass a custom optimizer, it is that your custom optimizer has a different signature than the other PyTorch optimizers. The solution is hence to take control of how the arguments are routed. Could you try if this works:

```python
def make_lookahead(parameters, optimizer_cls, k, alpha, **kwargs):
    optimizer = optimizer_cls(parameters, **kwargs)
    return Lookahead(optimizer=optimizer, k=k, alpha=alpha)

net = NeuralNet(
    ...,
    optimizer=make_lookahead,
    optimizer__optimizer_cls=torch.optim.Adam,
    optimizer__weight_decay=cfg.weight_decay,
    lr=cfg.lr,
)
```

The idea is basically to have a function that takes care of passing some parameters to `Lookahead` and the others to the optimizer class (i.e. `Adam`).

The alternatives would be to 1) subclass `NeuralNet` and override `initialize_optimizer` (as you mentioned) or 2) refactor `Lookahead` so that instead of requiring the initialized optimizer, it accepts the optimizer class and initializes it inside of its `__init__`.

supetronix:
Thanks @BenjaminBossan , your suggested solution worked perfectly (i.e. create a method to encapsulate Lookahead optimizer). 

Is it documented somewhere on Skorch website? I tried to look through documentation before asking this question so a link may be helpful for me and anyone else that may come across this questions.

[...]

PR #703, opened 2020-10-01
Adding code example to use python function for optimizer property

supetronix:
**Issue reference**: 
This is in reference to documentation issue https://github.com/skorch-dev/skorch/issues/696:

**What does it implement/fix**
I have added an example of using Python function for optimizer property where a user may need to use non-default PyTorch optimizer

### DIFF
--- a/docs/user/neuralnet.rst
+++ b/docs/user/neuralnet.rst
@@ -135,6 +135,26 @@ support for wildcards (globbing):
         ('linear0.bias', {'lr': 1}),
     ]
 
+If your use case requires you to use a non-default PyTorch optimizer then
+you can create a method and assign it to optimizer property:
+
+.. code:: python
+
+    # custom optimizer to encapsulate Adam
+    def make_lookahead(parameters, optimizer_cls, k, alpha, **kwargs):
+        optimizer = optimizer_cls(parameters, **kwargs)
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for the PR.

> If your use case requires you to use a non-default PyTorch optimizer then you can create a method and assign it to optimizer property:

I would phrase this differently. As mentioned in the issue, it's not really about using a PyTorch optimizer, but about using the same signature. Here is a proposal how to rephrase it:

`Your use case may require an optimizer whose signature differs from a default PyTorch optimizer's signature. In that case, you can define a custom function that reroutes the arguments as needed and pass it to the ``optimizer`` parameter:`

docs/user/neuralnet.rst
```diff
@@ -135,6 +135,28 @@ support for wildcards (globbing):
...
+
+
+    net = NeuralNetClassifier(
+            ...,
+            optimizer=make_lookahead,
+            optimizer__optimizer_cls=torch.optim.Adam,
+            optimizer__weight_decay=1e-2,
+            optimizer__k=5,
+            optimizer__alpha=0.5,
+            lr=1e-3
```
BenjaminBossan, line 158:
```suggestion
            lr=1e-3)
```

BenjaminBossan:
Thanks!

Issue #601, opened 2020-02-27
ADABoosting with Skorch


QuantumChamploo:
I would like to use skorch to implement Sklearn's ADABoosting methods on torch based models. When implemented (naively), an error saying "NeuralNetClassifier doesn't support sample_weight". I kind of expected this, but is there a way to extend the models to support sample weights?

I apologize if this is an inappropriate question or if has been answered, but I could not find anyone else trying this

BenjaminBossan:
Hi, you don't give much detail, so I'm not totally sure how you're trying to pass sample weights. But please have a look at our [docs](https://skorch.readthedocs.io/en/stable/user/FAQ.html#i-want-to-use-sample-weight-how-can-i-do-this) to see if this solves your problem. If it does, feel free to close the issue.

BenjaminBossan:
@QuantumChamploo Any updates on this?

QuantumChamploo:
Hey, thanks for checking in, been distracted by the world collapsing. Thank you for pointing me to the docs, but I could not get the example to work. Are there examples of this being used? The docs were a bit short on implementation details and could not find a more verbose example

As for ADABoosting, I'm a little more familiar with I am not sure if the dictionary implementation will work with ADA. Has ADABoosting with Skorch been implemented before?

BenjaminBossan:
Do you maybe have a code example, ideally something that I could reproduce 100%. It's a bit difficult for me to help you in the abstract. I'm fairly certain this can be made to work with skorch.

Stay safe!

QuantumChamploo:
''''from sklearn.datasets import fetch_openml
from sklearn.model_selection import train_test_split
import numpy as np
import matplotlib.pyplot as plt

import torch
from torch import nn
import torch.nn.functional as F

from skorch import NeuralNetClassifier

from sklearn.ensemble import AdaBoostClassifier


mnist = fetch_openml('mnist_784', cache=False)

X = mnist.data.astype('float32')
y = mnist.target.astype('int64')

X /= 255.0

X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.25, random_state=42)

device = 'cuda' if torch.cuda.is_available() else 'cpu'


mnist_dim = X.shape[1]
hidden_dim = int(mnist_dim/8)
output_dim = len(np.unique(mnist.target))


X = {'data': X_train}
X['sample_weight'] = np.ones(len(X_train), dtype=float)

class ClassifierModule(nn.Module):
    def __init__(
            self,
[...]

PR #710, opened 2020-10-17
Warn if valid data is shuffled

BenjaminBossan:
Check for iterator_train__shuffle=True and raise a UserWarning if found.

This is just a minor convenience feature for the user, to prevent unwanted consequences, as e.g. witnessed [here](https://github.com/skorch-dev/skorch/issues/601#issuecomment-653237672).

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1516,6 +1516,16 @@ class NeuralNet:
         did you mean iterator_train__shuffle?
 
         """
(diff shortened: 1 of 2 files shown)


thomasjpfan:
Is the test failing randomly?

BenjaminBossan:
> Is the test failing randomly?

It seems so, I re-ran the CI and now it passed.

thomasjpfan (approved):
LGTM

PR #715, opened 2020-11-01
Remove iid argument from GridSearchCV

BenjaminBossan:
The argument is deprecated and will be removed in sklearn 0.24.


thomasjpfan (approved):
LGTM

PR #716, opened 2020-11-01
DOC: simplify sample_weight example

stsievert:
It simplifies the sample_weight example in the docs. In the previous example, `MyModule.forward` signature changed but  arguments were changed, but there was a comment saying "sample_weight can be ignored." I used that as motivation to do the following:

* Remove need to change signature of `MyModule.forward` to accept `sample_weight`
* Ensure `criterion__reduce` is always `False` (also, one less definition).

I've also added some comments explaining each step of the reweighting.

**Reference issues/PRs**
None.

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -104,23 +104,19 @@ dictionary. Below, there is example code on how to achieve this:
     # add sample_weight to the X dict
     X['sample_weight'] = sample_weight
 
-    class MyModule(nn.Module):
-        ...  # normal PyTorch module definition
-
     class MyNet(NeuralNet):
-        def __init__(self, *args, criterion__reduce=False, **kwargs):
+        def __init__(self, *args, **kwargs):
             # make sure to set reduce=False in your criterion, since we need the loss
             # for each sample so that it can be weighted
-            super().__init__(*args, criterion__reduce=criterion__reduce, **kwargs)
+            super().__init__(*args, criterion__reduce=False, **kwargs)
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for the PR, Scott. In general, I like the direction, but I wonder if this change would actually work:

> In the previous example, `MyModule.forward` signature changed but arguments were changed, but there was a comment saying "sample_weight can be ignored."

The thing is that skorch will pass `sample_weight` as a keyword argument to the `forward` method. If the signature is not adjusted to account for that, there will be a `TypeError`. Or do I miss something?

Maybe the wording could be improved to make that more clear?

BenjaminBossan:
@stsievert Did you take a look at this?

stsievert:
Yes. That allowed me to fix a bug in my code.

I tried to rework this PR to add an optional `sample_weight` argument to `fit`/`partial_fit`. It looks possible but has some barriers.

BenjaminBossan:
> I tried to rework this PR to add an optional `sample_weight` argument to `fit`/`partial_fit`. It looks possible but has some barriers.

I think that won't be that easy. E.g. when you add shuffling to the dataset, the sample weights would need to be shuffled accordingly. This is not really possibly, except by attaching the sample weights to `X`. That's why the solution in the docs was proposed.

BenjaminBossan:
@stsievert any news on this?

stsievert:
I still think it's possible, but don't have time to look into it.

BenjaminBossan:
I think at this point, we can close the PR :)

Issue #576, opened 2019-12-16
device='auto' should select the fastest possible device


ottonemo:
I see the code 

    device = ‘cuda’ if torch.cuda.is_available() else ‘cpu’

repeated often in user code. Maybe we should introduce `device='auto'` exactly for this case?

BenjaminBossan:
I think this is from the offical pytorch documentation, maybe it's okay to leave it as is?

BenAjayiObe:
Hi @ottonemo & @BenjaminBossan , I would be quite interested in creating a PR for this issue if this is something that you decide you want to have.

BenjaminBossan:
I don't see the feature as strictly necessary but would be okay with it being added. @ottonemo @thomasjpfan any opinions on that?

If this change is made, I would suggest to keep the default as is. Also, @BenAjayiObe, please have a look at this function we introduced:

https://github.com/skorch-dev/skorch/blob/09be626e74512124eb74c76b4cbabad4d3b1f274/skorch/utils.py#L128-L136

The functionality should be added there. However, this function is not used consistently in our codebase each time we cast, this would need to be fixed.

What I did wonder: Should we also allow the option to set device to `None` (or `'none'`?) so as to prevent any kind of casting?

BenAjayiObe:
Thanks for the info @BenjaminBossan , that's very helpful. 👍

What would be the use case for preventing casting?

BenjaminBossan:
> What would be the use case for preventing casting?

I can imagine that some people would like to have complete control over it, say run part of the model on cpu and part of it on gpu. If we force the casting, this could be annoying to them.

ottonemo:
> What I did wonder: Should we also allow the option to set device to None (or 'none'?) so as to prevent any kind of casting?

Definitely a good idea.

> @ottonemo @thomasjpfan any opinions on that?

Well I made the suggestion, so I'm in favor (although this is strictly a small convenience feature). Leaving the default as is sounds good to me. As long as the feature doesn't blow up complexity too much it is fine with me.

adit24dhaya:
Opened a draft PR in #1141 that adds device='auto' via _check_device_auto in skorch.utils (to_tensor/to_device/get_map_location + net load/pickle paths). Default remains cpu. Happy to adjust if you'd prefer net.device to be resolved to cuda/cpu after initialize.

PR #717, opened 2020-11-07
Feature/auto device

briankosw:
Implements auto device detection as described in #576.

### DIFF
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -242,14 +242,11 @@ class TestToDevice:
         ('cpu', 'cuda'),
         ('cuda', 'cpu'),
         ('cuda', 'cuda'),
-        ('auto', None),
         (None, None),
     ])
     def test_check_device_torch_tensor(self, to_device, x, device_from, device_to):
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Hey, thanks for the PR. Before proceeding, did you take a look at the discussion in #600? The arguments presented there are still relevant I think.

briankosw:
Ah I didn't see that discussion. So I guess the main gist of the arguments presented in #600 is:
1. Reduce the mental overhead of configuring devices every single time.
2. Potentially support more devices other than 'cuda' and 'cpu'.
3. Make the user explicitly specify the device choice.

It wasn't immediately obvious to me what else needs to be done then? I think a really helpful feature would be to set the device in the beginning of the workflow and then skorch will automatically use that device for future moving/creating of tensors.

BenjaminBossan:
What I took away from the discussion is that having an `'auto'` option just for choosing between `'cpu'` and '`cuda`' is not really worth it, since it just replaces `net = NeuralNet(..., device='cuda' if torch.cuda.is_available() else 'cpu')`. There would be more of an advantage if other options than just those two were supported.

PR #719, opened 2020-11-16
API: allow disabling all callbacks

stsievert:
It allows disabling all callbacks with `callbacks=False`.

I ran into a use case that uselessly spent 97% of the training time in various callbacks with the default `callbacks=None`. After this PR with `callbacks=False`, it only spends 5% of the total training time in various callbacks, a rather drastic speedup.

### DIFF
--- a/docs/user/callbacks.rst
+++ b/docs/user/callbacks.rst
@@ -138,7 +138,7 @@ use a callback for learning rate scheduling (e.g. via
 :class:`.LRScheduler`) and want to test its usefulness, you can
 compare the performance once with and once without the callback.
 
-To completely disable callbacks, set ``callbacks=False``.
+To completely disable all callbacks, set ``callbacks="diable"``.
 
 
 Scoring
--- a/docs/user/callbacks.rst
+++ b/docs/user/callbacks.rst
@@ -138,7 +138,8 @@ use a callback for learning rate scheduling (e.g. via
 :class:`.LRScheduler`) and want to test its usefulness, you can
 compare the performance once with and once without the callback.
 
-To completely disable all callbacks, set ``callbacks="diable"``.
+To completely disable all callbacks, including default callbacks, 
+set ``callbacks="disable"``.
 
 
 Scoring
--- a/docs/user/callbacks.rst
+++ b/docs/user/callbacks.rst
@@ -138,6 +138,8 @@ use a callback for learning rate scheduling (e.g. via
 :class:`.LRScheduler`) and want to test its usefulness, you can
 compare the performance once with and once without the callback.
 
+To completely disable callbacks, set ``callbacks=False``.
+
 
 Scoring
 -------
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -465,6 +465,9 @@ class NeuralNet:
             callbacks_.append((name, cb))
 
         self.callbacks_ = callbacks_
+        if self.callbacks is False:
+            self.callbacks_ = []
+
         return self
 
     def initialize_criterion(self):
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -141,11 +141,20 @@ class NeuralNet:
       data and should return the tuple ``dataset_train, dataset_valid``.
       The validation data may be None.
 
-    callbacks : None, False, or list of Callback instances (default=None)
-      More callbacks, in addition to those returned by
-      ``get_default_callbacks``. Each callback should inherit from
-      :class:`.Callback`. If not ``None``, a list of callbacks is
-      expected where the callback names are inferred from the class
+    callbacks : None, str, or list of Callback instances (default=None)
+      Which callbacks to enable. There are three possible values:
+
+      If ``callbacks=None``, only use default callbacks,
+      those returned by ``get_default_callbacks``.
+
+      If ``callbacks="disable"``, disable all callbacks, or, do not run
+      any of the callbacks.
+
+      If ``callbacks`` is a list of callbacks, use those callbacks in
+      addition to the default callbacks. Each callback should be an
+      instance of :class:`.Callback`.
+
+      Callback names are inferred from the class
       name. Name conflicts are resolved by appending a count suffix
       starting with 1, e.g. ``EpochScoring_1``. Alternatively,
       a tuple ``(name, callback)`` can be passed, where ``name``
@@ -154,9 +163,6 @@ class NeuralNet:
       callbacks (e.g., for the callback with name ``'print_log'``, use
       ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
       'train_loss'])``).
-
-      If ``callbacks="disable"``, completely ignore all default
-      callbacks.
 
     predict_nonlinearity : callable, None, or 'auto' (default='auto')
       The nonlinearity to be applied to the prediction. When set to
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -141,7 +141,7 @@ class NeuralNet:
       data and should return the tuple ``dataset_train, dataset_valid``.
       The validation data may be None.
 
-    callbacks : None or list of Callback instances (default=None)
+    callbacks : None, False, or list of Callback instances (default=None)
       More callbacks, in addition to those returned by
       ``get_default_callbacks``. Each callback should inherit from
       :class:`.Callback`. If not ``None``, a list of callbacks is
@@ -154,6 +154,8 @@ class NeuralNet:
       callbacks (e.g., for the callback with name ``'print_log'``, use
       ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
       'train_loss'])``).
+
+      If ``False``, completely disable callbacks.
 
     predict_nonlinearity : callable, None, or 'auto' (default='auto')
       The nonlinearity to be applied to the prediction. When set to
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -147,7 +147,7 @@ class NeuralNet:
       If ``callbacks=None``, only use default callbacks,
       those returned by ``get_default_callbacks``.
 
-      If ``callbacks="disable"``, disable all callbacks, or, do not run
+      If ``callbacks="disable"``, disable all callbacks, i.e. do not run
       any of the callbacks.
 
       If ``callbacks`` is a list of callbacks, use those callbacks in
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -155,7 +155,8 @@ class NeuralNet:
       ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
       'train_loss'])``).
 
-      If ``False``, completely disable callbacks.
+      If ``callbacks="disable"``, completely ignore all default
+      callbacks.
 
     predict_nonlinearity : callable, None, or 'auto' (default='auto')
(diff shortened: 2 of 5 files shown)


BenjaminBossan:
For my better understanding, could you give more details about your use case. I can only imagine that the network and amount of data must have been really small for this to happen. Otherwise, it would indicate that one of our callbacks is much slower than it should be.

FYI, if I wanted to turn off all callbacks (only ever for debugging purposes, I would do this:

```python
net = NeuralNet(...).initialize()
net.callbacks_ = []
```

stsievert:
Yes, my network is very small, and the data are between ~1KB to 1MB.

> FYI, if I wanted to turn off all callbacks (only ever for debugging purposes, I would do this:

Yes, I know. That's basically what this PR does. I'm submitting this PR because I probably spent an hour or two hours debugging the performance setback.

BenjaminBossan:
I see the utility in this. What I'm unsure about is the use of `False` for this. It's really not intuitive why `None` would leave the default callbacks there but `False` removes them. I wonder if there is a better idea.

If that question is answered, I see two further points to improve:

- The documentation should be extended, it's not clear that this would touch the default callbacks.
- The test should be moved to `test_net.py`, because this is really a functionality of the net class itself.

stsievert:
What's skorch's deprecation policy? I only implemented `callbacks=False` to not deprecate the API. I would prefer this API:

* `callbacks=None`: no callbacks whatsoever.
* `callbacks="auto"`: default callbacks.
* `callbacks=callback_list`: use those callbacks.

BenjaminBossan:
For this change, I wouldn't break backwards compatibility. The utility would be quite low but everyone who uses the default callbacks would have to change their code, that's not worth it.

But maybe we can think of something else than `False` to completely remove all callbacks, including the defaults. Maybe `'no-defaults'`, `'empty'`, something like this?

BenjaminBossan:
@stsievert what do you think of my suggestions?

stsievert:
I agree – backwards compatibility should not be broken because unfortunately `callbacks=None` shows up in `NeuralNet(...).get_params()`.

I like `callbacks=False` because to me it says "callback existence is False". Out of the string options, I like `callbacks="ignore"` most because it answers the same question (but I like `callbacks=False` more). I don't like `callbacks in ["no-defaults", "empty"]` because that's asking for `callbacks=None`.

BenjaminBossan:
> Out of the string options, I like `callbacks="ignore"` most

For me, `"ignore"` implies that there are callbacks, but they are ignored, whereas the implementation completely removes the callbacks. But in the end, I'm totally fine with this choice, what matters is that it's well documented.

Regarding your implementation, I would suggest a small change. Right now, you're checking at the end of `initialize_callbacks`, after all other mechanisms have already been performed unnecessarily. I would propose to perform this check at the beginning of the method and `return self` immediately afterwards. This is a bit more efficient and also more in line with the rest of the skorch code.

stsievert:
I've implemented `callbacks="disable"` to allow disabling all callbacks. I've also better formatted the docstring for `callbacks`.

docs/user/callbacks.rst
```diff
@@ -138,6 +138,8 @@ use a callback for learning rate scheduling (e.g. via
 :class:`.LRScheduler`) and want to test its usefulness, you can
 compare the performance once with and once without the callback.
 
+To completely disable all callbacks, set ``callbacks="diable"``.
```
BenjaminBossan, line 141:
```suggestion
To completely disable all callbacks, including default callbacks, 
set ``callbacks="disable"``.
```

skorch/tests/callbacks/test_training.py
```diff
@@ -1083,3 +1083,11 @@ def initialize_module(self, *args, **kwargs):
 
         assert save_params_mock.call_count == 1
         save_params_mock.assert_has_calls([call(f_mymodule='train_end_mymodule.pt')])
+
+
+class TestNoCallbacks(TestCheckpoint):
+    def test_no_callbacks(self, net_cls, data):
+        net = net_cls(callbacks="disable")
+        net.fit(*data)
+        assert net.callbacks == "disable"
```
BenjaminBossan, line 1092:
```suggestion
```
Not necessary.

skorch/tests/callbacks/test_training.py
```diff
@@ -1083,3 +1083,11 @@ def initialize_module(self, *args, **kwargs):
 
         assert save_params_mock.call_count == 1
         save_params_mock.assert_has_calls([call(f_mymodule='train_end_mymodule.pt')])
+
+
+class TestNoCallbacks(TestCheckpoint):
+    def test_no_callbacks(self, net_cls, data):
```
BenjaminBossan, line 1089: Please move this test into class `TestNeuralNet` inside `tests/test_net.py` (just this test, no extra class required).

skorch/tests/callbacks/test_training.py
```diff
@@ -1083,3 +1083,11 @@ def initialize_module(self, *args, **kwargs):
...
+        net = net_cls(callbacks="disable")
+        net.fit(*data)
```
BenjaminBossan, line 1091:
```suggestion
        net.initialize()
```
Let's not call `fit` here, since it's not necessary and it takes a bit of time. Consequently, there is also no more need for the `data` fixture.

BenjaminBossan (changes requested):
"disable" is fine with me.

I have a few minor change requests, please take a look. Also, please let the users know about this new feature by adding an entry to the CHANGES.md. Apart from that, this is good to go.

skorch/net.py
```diff
@@ -141,11 +141,20 @@ class NeuralNet:
...
-      :class:`.Callback`. If not ``None``, a list of callbacks is
-      expected where the callback names are inferred from the class
+    callbacks : None, str, or list of Callback instances (default=None)
+      Which callbacks to enable. There are three possible values:
+
+      If ``callbacks=None``, only use default callbacks,
+      those returned by ``get_default_callbacks``.
+
+      If ``callbacks="disable"``, disable all callbacks, or, do not run
+      any of the callbacks.
```
thomasjpfan, line 151:
```suggestion
      If ``callbacks="disable"``, disable all callbacks, i.e. do not run
      any of the callbacks.
```

skorch/net.py
```diff
@@ -141,11 +141,20 @@ class NeuralNet:
       data and should return the tuple ``dataset_train, dataset_valid``.
       The validation data may be None.
 
-    callbacks : None or list of Callback instances (default=None)
-      More callbacks, in addition to those returned by
-      ``get_default_callbacks``. Each callback should inherit from
-      :class:`.Callback`. If not ``None``, a list of callbacks is
-      expected where the callback names are inferred from the class
+    callbacks : None, str, or list of Callback instances (default=None)
+      Which callbacks to enable. There are three possible values:
```
thomasjpfan, line 145: Since we only accept only one string option, we can be explicit here:

```suggestion
    callbacks : None, "disable", or list of Callback instances (default=None)
      Which callbacks to enable. There are three possible values:
```

thomasjpfan (commented):
Thank you for the PR @stsievert !

It is interesting how much overhead the callbacks have.

We should include this new feature into https://github.com/skorch-dev/skorch/blob/master/CHANGES.md

BenjaminBossan:
Thanks so much @stsievert.

This is almost good to merge, just a tiny improvement: Your test now resides within `class TestNetSparseInput`, where it doesn't belong. I would suggest to put the test into the "section" about initialization: https://github.com/skorch-dev/skorch/blob/82a3da522a877c2bee8a276ddb4fdaebb8e44dbe/skorch/tests/test_net.py#L1422-L1424

That is really our fault for having such a huge test file, we should probably clean it up one day. It just grew organically to 3000 LOC and now it's hard to insert new tests there.

BenjaminBossan:
> It is interesting how much overhead the callbacks have.

Could be worth investigating.

BenjaminBossan:
I tried to run a few benchmarks but I had to try really hard to make disabling default callbacks matter. Here is what I came up with in the end:

```python
import numpy as np
from torch import nn

# Only 10 samples, 5 features
X, y = make_classification(10, 5, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

class MyModule(nn.Module):
    def __init__(self):
        super(MyModule, self).__init__()

        # very small net
        self.lin = nn.Linear(5, 2)

        X = self.softmax(self.lin(X))
        return X

net = NeuralNetClassifier(
    MyModule,
    max_epochs=500,
    lr=0.1,
    iterator_train__shuffle=True,
    train_split=False,
    #verbose=0,  # <= experiment 1
    #callbacks__train_loss=None,  # <= experiment 2 
)
net.fit(X, y)
```

As you can see, both data and the module are really really small, number of epochs are quite high. In these circumstances, disabling default callbacks reduces wall time from ~3.4 to ~1.7 sec.

What I found takes most time:

1. print log: setting `verbose=0` reduces time from 3.4 to 3.0 sec. This is not too surprising, given how it involves IO
2. train loss: disabling the calculation of train loss leads us from 3.0 to 2.2 sec, so pretty close to 1.7 sec without any default callbacks (note that I disabled the validation split)

My suspicion was that train loss calculation is slow because `history` access is slow. Indeed, almost all of the overhead is spent indexing into the batch logs: https://github.com/skorch-dev/skorch/blob/82a3da522a877c2bee8a276ddb4fdaebb8e44dbe/skorch/history.py#L204-L209

After a short time of tinkering, I couldn't find a way of accelerating these lines without making the code a lot more complicated. Thus I wonder if it's worth it to fix the issue or if we can consider this situation to be too marginal, especially since this PR will provide a solution that should satisfy most of these already rare situations.

stsievert:
This more accurately mirrors my use case:

``` python
import numpy as np
from torch import nn

class MyModule(nn.Module):
    def __init__(self, n_targets=85):
        super(MyModule, self).__init__()

        self.lin1 = nn.Linear(3, n_targets)
        self.lin2 = nn.Linear(n_targets, 2)

        X = self.softmax(self.lin1(X))
        X = self.softmax(self.lin2(X))
        return X


X, y = make_classification(n_samples=1000, n_features=3, n_redundant=1, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

est = NeuralNetClassifier(
    module=MyModule,
    max_epochs=500,
    lr=0.1,
    iterator_train__shuffle=True,
    train_split=False)
```

Here are the timings on my machine:

* `net.fit(X, y)`: 12.4s
* `net.set_params(verbose=0)`: 10.0s
* `net.set_params(verbose=0, callbacks__train_loss=None).fit(X, y)`: 6.97s

BenjaminBossan:
Thanks @stsievert, I can reproduce the relative numbers. With `verbose=0`, I get 20 sec, 11 of which are `history.__getitem__`. Unsurprisingly, it gets worse the longer the run is. I will investigate whether there are any worthwhile improvements there.

Meanwhile, could you please move the test as I indicated [above](https://github.com/skorch-dev/skorch/pull/719#issuecomment-748483376) so that we can merge?

stsievert:
The test has been moved as suggested in https://github.com/skorch-dev/skorch/pull/719#issuecomment-748483376.

BenjaminBossan (approved):
Perfect, thank you for implementing this feature.

PR #724, opened 2020-11-21
Add notebook that shows how to use skorch with optuna

### LINKED ISSUE #718: Add an optuna example
It would be nice to add an example of using [optuna](https://github.com/optuna/optuna) with skorch. The main reason is that `GridSearchCV` and `RandomizedSearchCV` can be much too slow for certain neural networks. Optuna has the potential to be much faster.

[This example](https://github.com/optuna/optuna/blob/master/examples/pytorch_simple.py) could be a starting point. As to how, possibilities are a Python script, a Jupyter notebook, or an extension of the docs. Maybe even some of those, like adding a code example _and_ a new chapter in the docs about hyper parameter search.

BenjaminBossan:
Fixes #718 

[This notebook](https://github.com/skorch-dev/skorch/blob/feature/optuna-example-718/notebooks/optuna-example.ipynb) shows how to use skorch in conjunction with optuna, which can be much more efficient than `GridSearchCV` or `RandomizedSearchCV`. I don't have a deep knowledge of optuna, so there may be room for improvement. The whole example is inspired by [this one](https://github.com/optuna/optuna/blob/master/examples/pytorch_simple.py).

Right now, the implementation makes use of the skorch-internal train/valid split. This allows us to use optuna with very few lines of extra code.

Right now, this notebook contains an ad hoc implementation of an `OptunaCallback`, which is responsible for logging and pruning trials. Possibly, this could be extended for more functionality. If so, it could find it's way into the skorch code base.

Eventually, we could consider adding a section to the docs about hyper-parameter optimization using different libraries. WDYT?


BenjaminBossan:
Update: optuna already provides a [skorch callback](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.SkorchPruningCallback.html), which is almost exactly the same that I used. I changed the notebook to use this instead.

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "def define_model(trial):\n",
+    "    \"\"\"Model definition and hyper-parameter setting\n",
+    "\n",
+    "    We set the number of units, the optimizer, the learning rate, and the number of epochs.\n",
+    "    \n",
+    "    To make this work efficiently with optuna, add the\n",
+    "    `SkorchPruningCallback <https://optuna.readthedocs.io/en/stable/reference/generated/optuna.integration.OptunaSearchCV.html?highlight=optuna.integration.OptunaSearchCV>`\n",
+    "    from the optuna library.\n",
+    "\n",
+    "    \"\"\"\n",
```
thomasjpfan, line 182: Given this is a notebook, can we pull this out into an markdown block above this code block?

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,495 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def objective(trial):\n",
+    "    \"\"\"Train the model and return the value to be maximized\"\"\"\n",
```
thomasjpfan, line 208: A markdown description before defining this function?

> Next, we define an objective function that trains a `NeuralNetClassifier` and returns the value to be maximized.

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,495 @@
...
+    "    model = define_model(trial)\n",
+    "    X, y = get_data()\n",
+    "    model.fit(X, y)\n",
+    "    accuracy = model.history[-1, METRIC_NAME]\n",
```
thomasjpfan, line 212: Would it be cleaner to directly place the metric name here?

```suggestion
    "    accuracy = model.history[-1, 'valid_acc']\n",
```

thomasjpfan (commented):
It's nice to see how easy it is to integrate with optuna.

notebooks/optuna-example.ipynb
BenjaminBossan: done

notebooks/optuna-example.ipynb
BenjaminBossan: dome

notebooks/optuna-example.ipynb
BenjaminBossan: I would leave it as is. The idea is to set all the constants at the beginning of the notebook. This way, if I want to change the them (say, optimizing `valid_loss` instead), I don't need to chase through the notebook to replace all instances.

BenjaminBossan:
@thomasjpfan I improved the notebook by adding more descriptions, as you correctly pointed out. Now it should be more accessible for users who don't already know optuna. Furthermore, I increased the image size, which was quite small originally.

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "The PyTorch module is a simple multi-layer perceptron with dropout."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "Tree-structured Parzen Estimator) algorithmclass MyModule(nn.Module):\n",
```
thomasjpfan, line 165: typo?

```suggestion
    "class MyModule(nn.Module):\n",
```

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,554 @@
...
+    "        device=DEVICE,\n",
+    "        callbacks=[SkorchPruningCallback(trial, monitor='valid_acc')],\n",
+    "    )"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In optuna, the objective should be a function that accepts an instance of the `trial` returns the value that we want to optimize."
```
thomasjpfan, line 230:
```suggestion
    "In optuna, the objective should be a function that accepts an instance of the `trial` and returns the value that we want to optimize."
```

notebooks/optuna-example.ipynb
```diff
@@ -0,0 +1,554 @@
...
...Lu3aEIC4Pg5WXrUIksxtnqP1kLTvxsyBnZ+89qq9Roh4aGIiMjA7du3UJlZSWOHTuG8PBwk31KSkpQWVkJADh48CC6desGpVJpVl9qXme+2ocsyQ2elTp0vf0HfMuLUCC44mJWKVwffRQu4eFMssnpOVv9J2vBiZ8NOSN7/1ltlSvacrkc06ZNw4oVKyCKIqKiohASEoL9+/cDAEaOHIm0tDR8+OGHkMlkCA4OxrPPPltvX7Ic19xslCs80LEwHcUuSmR6qqGTu6JreQnXv6YWxRprAVsT1zZu2fjZkLOy55/VVltHu2/fvujbt69J28iRI43/7tKlCz744AOz+5Ll6DV+8Lt+EwKALKUvyhRu8NSXQR/oZ+vQqJFYl0nWwrWN7Rc/GyLrs0rpCDmWXhNHIaiyCDJJhE7hCk99GfyFcvSaOMrWoVEjVNdl7r98G7+mFWP/5duYvSsF6QXltg6NnNAzA9sgyNvVpM2e6iVbMn42RNbHR7BTDQE9u8ClXw/8npyGMKEU+kA/9Jo4Cm16dbV1aNQI9dVl2uuf2shxVddLbjmegZySCmg9+RcUe8HPhsj6mGhTDVJuLlRqHzww9yHIO3a0dTjURKzLJGuz53rJlo6fDZF1sXSEahBvpgKCAFmQfT8EgMzDukwiIiLbYKJNNRjSUiFr7QfBjX9OdAasyyQiIrINlo6QCam0FFJuHuR9+9g6FGomrMskIiKyDSbaZEJMTQUAyIKDbRwJNSfWZRIREVkfS0fIhCEtDYKnEgIfc09ERETUJEy0yUgyGCCmp0MWHAxBEGwdDhEREZFDY6JNRlJWFlBRCVkQy0aIiIiImoqJNhkZUlMBuRyyNgG2DoWIiIjI4THRJiMxNQ2yAH8ILlxfmYiIiKipmGgTAEAsKIBUWAhZcIitQyEiIiJyCky0CQAgpqUBAOTBfBokERERUXNgok0AqspGBB8fCF5etg6FiIiIyCkw0SZIFRUQszL5kBoiIiKiZsREmyBmZAAGkWUjRERERM2IiTZBvJkKuLpAaN3a1qEQEREROQ0m2i2cJEkQ01IhCwyEIJfbOhwiIiIip8FEu4WT8vIglZZBzvpsIiIiombFRLuFE1OrlvWTBbE+m4iIiKg5MdFu4cS0VMi0GggeHrYOhYiIiMipMNFuwSSdDmJ2Dpf1IyIiIrIAJtotmJiWBkgSy0aIiI
```
thomasjpfan, line 526: Unrelated note: In general, I have found parallel coordinate plots more useful when they are interactive. (Something like https://facebookresearch.github.io/hiplot/)

thomasjpfan (approved):
Otherwise LGTM

notebooks/optuna-example.ipynb
BenjaminBossan: True, maybe worth suggesting to the optuna maintainers.

Abdelgha-4: Optuna now have the interactive [parallel coordinate plot].(https://optuna.readthedocs.io/en/stable/reference/visualization/generated/optuna.visualization.plot_parallel_coordinate.html)

BenjaminBossan: Nice, wish come true, thanks for letting us know.

BenjaminBossan:
Closing this as it's super old :)

PR #725, opened 2020-12-01
Add a logging callback for Sacred

timokau:
Similar to the existing Neptune logger, just logging to [Sacred](https://github.com/IDSIA/sacred) instead
of Neptune. The main difference is the name of the logging function and
that an experiment does not have to be closed explicitly. I kept the
documentation and functionality similar, to make it easier to use one as
a drop-in replacement for the other.

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -835,14 +835,6 @@ class SacredLogger(Callback):
         self.log_on_epoch_end = log_on_epoch_end
         self.batch_suffix = batch_suffix
         self.epoch_suffix = epoch_suffix
-        if self.batch_suffix is None:
-            self.batch_suffix = (
-                "_batch" if log_on_batch_end and log_on_epoch_end else ""
-            )
-        if self.epoch_suffix is None:
-            self.epoch_suffix = (
-                "_epoch" if log_on_batch_end and log_on_epoch_end else ""
-            )
         self.keys_ignored = keys_ignored
 
     def initialize(self):
@@ -851,6 +843,17 @@ class SacredLogger(Callback):
             keys_ignored = [keys_ignored]
         self.keys_ignored_ = set(keys_ignored or [])
         self.keys_ignored_.add("batches")
+
+        self.batch_suffix_ = self.batch_suffix
+        self.epoch_suffix_ = self.epoch_suffix
+        if self.batch_suffix__ is None:
+            self.batch_suffix_ = (
+                "_batch" if log_on_batch_end and log_on_epoch_end else ""
+            )
+        if self.epoch_suffix_ is None:
+            self.epoch_suffix_ = (
+                "_epoch" if log_on_batch_end and log_on_epoch_end else ""
+            )
         return self
 
     def on_batch_end(self, net, **kwargs):
@@ -861,7 +864,7 @@ class SacredLogger(Callback):
         for key in filter_log_keys(batch_logs.keys(), self.keys_ignored_):
             # skorch does not keep a batch count, but sacred will
             # automatically associate the results with a counter.
-            self.experiment.log_scalar(key + self.batch_suffix, batch_logs[key])
+            self.experiment.log_scalar(key + self.batch_suffix_, batch_logs[key])
 
     def on_epoch_end(self, net, **kwargs):
         """Automatically log values from the last history step."""
@@ -871,4 +874,4 @@ class SacredLogger(Callback):
         epoch = epoch_logs["epoch"]
 
         for key in filter_log_keys(epoch_logs.keys(), self.keys_ignored_):
-            self.experiment.log_scalar(key + self.epoch_suffix, epoch_logs[key], epoch)
+            self.experiment.log_scalar(key + self.epoch_suffix_, epoch_logs[key], epoch)
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -744,6 +744,40 @@ class SacredLogger(Callback):
     more information. Alternatively you can subclass this callback and extend
     the ``on_*`` methods.
 
+    To use this logger, you first have to install Sacred:
+
+    >>> pip install sacred
+
+    You might also install pymongo to use a mongodb backend. See the upstream_
+    documentation for more details. Once you have installed it, you can set up
+    a simple experiment and pass this Logger as a callback to your skorch
+    estimator:
+
+    >>> from sacred import Experiment
+    >>> from sokrch.callbacks.logging import SacredLogger
+    >>> from sokrch.callbacks.scoring import EpochScoring
+    >>> ex = Experiment()
+
+    >>> @ex.config
+    >>> def my_config():
+    ...     max_epochs = 20
+    ...     lr = 0.01
+
+    >>> @ex.automain
+    >>> def main(_run, max_epochs, lr):
+    ...     # Take care to add additional scoring callbacks *before* the logger.
+    ...     net = NeuralNetClassifier(
+    ...         ClassifierModule,
+    ...         max_epochs=max_epochs,
+    ...         lr=0.01,
+    ...         callbacks=[EpochScoring("f1"), SacredLogger(_run)]
+    ...     )
+    ...     # now fit your estimator to your data
+    ...     net.fit(X, y)
+
+    You can then call this script and optionally pass a backend as a command
+    line parameter.
+
 
     Parameters
     ----------
@@ -757,6 +791,9 @@ class SacredLogger(Callback):
       Key or list of keys that should not be logged to Sacred. Note that in
       addition to the keys provided by the user, keys such as those starting
       with 'event_' or ending on '_best' are ignored by default.
+
+
+    .. _upstream: https://github.com/IDSIA/sacred#installing
     """
 
     def __init__(
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -783,7 +783,7 @@ class SacredLogger(Callback):
     ...     )
     ...     # now fit your estimator to your data
     ...     net.fit(X, y)
-    
+
     Then call this from the command line, e.g. like this:
     ``python sacred-script.py with max_epochs=15``
 
@@ -799,6 +799,19 @@ class SacredLogger(Callback):
     log_on_batch_end : bool (default=False)
       Whether to log loss and other metrics on batch level.
 
+    log_on_epoch_end : bool (default=True)
+      Whether to log loss and other metrics on epoch level.
+
+    batch_suffix : str (default=None)
+      A string that will be appended to all logged keys. By default (if set to
+      ``None``) "_batch" is used if batch and epoch logging are both enabled
+      and no suffix is used otherwise.
+
+    epoch_suffix : str (default=None)
+      A string that will be appended to all logged keys. By default (if set to
+      ``None``) "_epoch" is used if batch and epoch logging are both enabled
+      and no suffix is used otherwise.
+
     keys_ignored : str or list of str (default=None)
       Key or list of keys that should not be logged to Sacred. Note that in
       addition to the keys provided by the user, keys such as those starting
@@ -812,10 +825,24 @@ class SacredLogger(Callback):
         self,
         experiment,
         log_on_batch_end=False,
+        log_on_epoch_end=True,
+        batch_suffix=None,
+        epoch_suffix=None,
         keys_ignored=None,
     ):
         self.experiment = experiment
         self.log_on_batch_end = log_on_batch_end
+        self.log_on_epoch_end = log_on_epoch_end
+        self.batch_suffix = batch_suffix
+        self.epoch_suffix = epoch_suffix
+        if self.batch_suffix is None:
+            self.batch_suffix = (
+                "_batch" if log_on_batch_end and log_on_epoch_end else ""
+            )
+        if self.epoch_suffix is None:
+            self.epoch_suffix = (
+                "_epoch" if log_on_batch_end and log_on_epoch_end else ""
+            )
         self.keys_ignored = keys_ignored
 
     def initialize(self):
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
Thanks a lot, I will take a long look when I have some time.

Some things I noticed from first glance: I think it would be helpful for users if you could document how to install sacred in the docstring, since it doesn't come automatically with skorch. Furthermore, please add an entry to the CHANGES.md.

Also, we should probably add a few tests, similar to the ones for the [other loggers](https://github.com/skorch-dev/skorch/blob/master/skorch/tests/callbacks/test_logging.py). Would you feel up for the task of adding those?

timokau:
Sure, thanks for the pointers :)

I have added some usage & installation documentation, a changelog entry and a test. The test is basically an adaption of the Neptune test, so there is a lot of similar code. I'm not sure what the policy for that is in skorch.

How does this look to you? Do you want me to squash my changes?

timokau:
Sorry for all the notifications. I discovered that the test wasn't actually executed in the CI since I didn't add sacred to requirements-dev. The test then failed, which I hopefully fixed. Unfortunately the CI isn't running anymore for some reason. I'll probably have to set up a local dev environment after all.

BenjaminBossan:
Hey, thanks for adding the changes I asked for. I'll try to give this a good look on the weekend. Then I'll also try to run the tests. I wonder why the CI stopped, maybe it's related to the travis policy changes they announced recently.

> Do you want me to squash my changes?

Please don't, since it makes it harder to review the PR. I will squash before merging.

skorch/callbacks/logging.py
```diff
@@ -728,3 +728,105 @@ def on_epoch_end(self, net, **kwargs):
     def on_train_end(self, net, **kwargs):
         if self.close_after_train:
             self.writer.close()
+
+
+class SacredLogger(Callback):
+    """Logs results from history to Sacred.
+
+    Sacred is a tool to help you configure, organize, log and reproduce
+    experiments developed at IDSIA. See https://github.com/IDSIA/sacred.
```
BenjaminBossan, line 737:
```suggestion
    experiments. Developed at IDSIA. See https://github.com/IDSIA/sacred.
```

Otherwise, it makes you think Sacred only works for "experiments developed at IDSIA" :)

skorch/callbacks/logging.py
```diff
@@ -728,3 +728,105 @@ def on_epoch_end(self, net, **kwargs):
...
+    >>> def main(_run, max_epochs, lr):
+    ...     # Take care to add additional scoring callbacks *before* the logger.
+    ...     net = NeuralNetClassifier(
+    ...         ClassifierModule,
+    ...         max_epochs=max_epochs,
+    ...         lr=0.01,
+    ...         callbacks=[EpochScoring("f1"), SacredLogger(_run)]
+    ...     )
+    ...     # now fit your estimator to your data
+    ...     net.fit(X, y)
```
BenjaminBossan, line 776:
```suggestion
    # contents of sacred-experiment.py
    >>> import numpy as np
    >>> from sacred import Experiment
    >>> from sklearn.datasets import make_classification
    >>> from skorch.callbacks.logging import SacredLogger
    >>> from skorch.callbacks.scoring import EpochScoring
    >>> from skorch import NeuralNetClassifier
    >>> from skorch.toy import make_classifier

    >>> ex = Experiment()

    >>> @ex.config
    >>> def my_config():
    ...     max_epochs = 20
    ...     lr = 0.01

    >>> X, y = make_classification()
    >>> X, y = X.astype(np.float32), y.astype(np.int64)

    >>> @ex.automain
    >>> def main(_run, max_epochs, lr):
    ...     # Take care to add additional scoring callbacks *before* the logger.
    ...     net = NeuralNetClassifier(
    ...         make_classifier(),
    ...         max_epochs=max_epochs,
    ...         lr=0.01,
    ...         callbacks=[EpochScoring("f1"), SacredLogger(_run)]
    ...     )
    ...     # now fit your estimator to your data
    ...     net.fit(X, y)
    
    Then call this from the command line, e.g. like this:
    ``python sacred-script.py with max_epochs=15``
```

I amended the example so that it can run without further additions. Could you maybe add a sentence to summarize what Sacred does for the user?

skorch/tests/callbacks/test_logging.py
```diff
@@ -191,6 +192,139 @@ def test_first_batch_flag(
...
+            self,
+            net_cls,
+            classifier_module,
+            data,
+            sacred_logger_cls,
+            mock_experiment,
+    ):
+        # ignore 'dur' and 'valid_loss', 'unknown' doesn't exist but
+        # this should not cause a problem
+        npt = sacred_logger_cls(
```
BenjaminBossan, line 251:
```suggestion
        logger = sacred_logger_cls(
```
`npt` was for "neptune" ;)

skorch/tests/callbacks/test_logging.py
```diff
@@ -191,6 +192,139 @@ def test_first_batch_flag(
...
+            mock_experiment, keys_ignored=['dur', 'valid_loss', 'unknown'])
+        net_cls(
+            classifier_module,
+            callbacks=[npt],
```
BenjaminBossan, line 255:
```suggestion
            callbacks=[logger],
```

skorch/tests/callbacks/test_logging.py
```diff
@@ -191,6 +192,139 @@ def test_first_batch_flag(
...
+            max_epochs=3,
+        ).fit(*data)
+
+        # 3 epochs x 2 epoch metrics = 6 calls
+        assert mock_experiment.log_scalar.call_count == 6
+        call_args = [args[0][0] for args in mock_experiment.log_scalar.call_args_list]
+        assert 'valid_loss' not in call_args
+
+    def test_keys_ignored_is_string(self, sacred_logger_cls, mock_experiment):
+        npt = sacred_logger_cls(
```
BenjaminBossan, line 265:
```suggestion
        logger = sacred_logger_cls(
```

skorch/tests/callbacks/test_logging.py
```diff
@@ -191,6 +192,139 @@ def test_first_batch_flag(
...
+            mock_experiment, keys_ignored='a-key').initialize()
+        expected = {'a-key', 'batches'}
+        assert npt.keys_ignored_ == expected
```
BenjaminBossan, line 268:
```suggestion
        assert logger.keys_ignored_ == expected
```

BenjaminBossan (changes requested):
Overall looks pretty good, thanks for adding the requested changes. I have a few comments, please take a look.

Additionally, one of the tests is failing. Could you please take a look at the error below?

```
    def test_fit_with_real_experiment(
            self,
            net_cls,
            classifier_module,
            data,
            sacred_logger_cls,
            sacred_experiment_cls,
    ):
        net = net_cls(
            classifier_module,
            callbacks=[sacred_logger_cls(sacred_experiment_cls())],
            max_epochs=5,
        )
>       net.fit(*data)

skorch/tests/callbacks/test_logging.py:283: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
skorch/classifier.py:142: in fit
    return super(NeuralNetClassifier, self).fit(X, y, **fit_params)
skorch/net.py:903: in fit
    self.partial_fit(X, y, **fit_params)
skorch/net.py:862: in partial_fit
    self.fit_loop(X, y, **fit_params)
skorch/net.py:782: in fit_loop
    self.notify("on_epoch_end", **on_epoch_kwargs)
skorch/net.py:325: in notify
    getattr(cb, method_name)(self, **cb_kwargs)
skorch/callbacks/logging.py:832: in on_epoch_end
    self.experiment.log_scalar(key, epoch_logs[key], epoch)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <sacred.experiment.Experiment object at 0x7f9d2c115d10>, name = 'dur'
value = 0.002778291702270508, step = 1

    def log_scalar(self, name: str, value: float, step: Optional[int] = None) -> None:
        """
        Add a new measurement.
    
        The measurement will be processed by the MongoDB* observer
        during a heartbeat event.
        Other observers are not yet supported.
    
    
        Parameters
        ----------
        name
            The name of the metric, e.g. training.loss
        value
            The measured value
        step
            The step number (integer), e.g. the iteration number
            If not specified, an internal counter for each metric
            is used, incremented by one.
        """
        # Method added in change https://github.com/chovanecm/sacred/issues/4
        # The same as Run.log_scalar
>       self.current_run.log_scalar(name, value, step)
E       AttributeError: 'NoneType' object has no attribute 'log_scalar'
```

skorch/callbacks/logging.py
timokau: That is their official tagline, but I see how it might be confusing :smile:

skorch/tests/callbacks/test_logging.py
timokau: Of course! Seems like I took the "inspiration" from the neptune test a little too far ;)

skorch/callbacks/logging.py
timokau: Is "Sacred is a tool to help you configure, organize, log and reproduce experiments" too broad?

timokau:
I think I addressed all of your feedback. The revived CI is also happy. Please have another look.

BenjaminBossan:
Thanks, Timo, I will take a look as soon as I have a bit of time.

skorch/callbacks/logging.py
```diff
@@ -728,3 +728,117 @@ def on_epoch_end(self, net, **kwargs):
...
+    your net's history to Sacred.
+
+    If you want to log additional information, you can simply add it to
+    ``History``. See the documentation on ``Callbacks``, and ``Scoring`` for
+    more information. Alternatively you can subclass this callback and extend
+    the ``on_*`` methods.
+
+    To use this logger, you first have to install Sacred:
+
+    >>> pip install sacred
```
BenjaminBossan, line 749:
```suggestion
    $ pip install sacred
```
Since this is not run in a Python prompt.

BenjaminBossan (commented):
Thanks again for your adjustments, this is almost good to go. I suggested a small amendment of the docstring, the rest looks good.

Before merging, I have a question. Both the batch-level logging and the epoch-level logging use `experiment.log_scalar`. How does Sacred know the difference? Is it because the latter explicitly passes the `epoch`? Or does it not matter to differentiate?

timokau:
> How does Sacred know the difference? Is it because the latter explicitly passes the epoch? Or does it not matter to differentiate?

Good catch! It doesn't. I took this behavior from the neptune logger and implicitly assumed that batch logging and epoch logging were mutually exclusive, but it looks like that isn't the case. I only needed batch logging for my own usecase, so it was kind of an afterthought to match `NeptuneLogger`'s behavior.

I could think of two ways to change this:

1. Add a `_batch` suffix to the key when logging at the batch level.
2. Disable epoch logging when logging at the batch level.

Intuitively I'd expect (2). One issue would be that you can't easily add two loggers (one for batch logging, one for epoch logging), since the keys would conflict. We could resolve that by adding an optional prefix/suffix argument to the logger.

Assuming Neptune doesn't do any magic to infer batch vs epoch logging, it currently behaves the same way though. What would you prefer?

BenjaminBossan:
> I could think of two ways to change this:
> 
>     1. Add a `_batch` suffix to the key when logging at the batch level.
> 
>     2. Disable epoch logging when logging at the batch level.

For option 1, that would mean that we would see 2 graphs in Sacred, one with batch metrics that is updated frequently and another one with epoch metrics. Don't you think that would be quite nice for the user?

> Assuming Neptune doesn't do any magic to infer batch vs epoch logging, it currently behaves the same way though. What would you prefer?

Maybe @jakubczakon can clarify what happens with Neptune.

Anyway, `SacredLogger` should do what makes sense for Sacred. Of course it would be nice if all those loggers were as similar as possible, so that users can swap them easily, but I wouldn't go as far as to emulate behavior that just doesn't fit well with what Sacred offers.

timokau:
> For option 1, that would mean that we would see 2 graphs in Sacred, one with batch metrics that is updated frequently and another one with epoch metrics. Don't you think that would be quite nice for the user?

Yes, that could be nice. It shouldn't hurt much in any case.

I have pushed some changes. The user can now choose batch and/or epoch logging and a suffix if necessary. What do you think? I can also remove the suffix configuration and always enable epoch logging ("option 1") if you think the current implementation is unnecessarily complex.

skorch/callbacks/logging.py
```diff
@@ -728,3 +728,147 @@ def on_epoch_end(self, net, **kwargs):
...
+        self.batch_suffix = batch_suffix
+        self.epoch_suffix = epoch_suffix
+        if self.batch_suffix is None:
+            self.batch_suffix = (
+                "_batch" if log_on_batch_end and log_on_epoch_end else ""
+            )
+        if self.epoch_suffix is None:
+            self.epoch_suffix = (
+                "_epoch" if log_on_batch_end and log_on_epoch_end else ""
+            )
```
BenjaminBossan, line 845: Could you please move this logic to `initialize`? Also, add an underscore to the attribute name: `self.batch_suffix_ = ...` (analog to `keys_ignored_`).

The reason is that all `__init__` parameters should be left untouched. That way, if a user calls `set_params(callbacks__sacred__batch_suffix="foo")`, this extra logic is honored, otherwise, it would be ignored. I know this is a total edge case, but we should be consistent with the general sklearn mechanics here.

BenjaminBossan (changes requested):
Thanks for the changes. I don't think the logger is too complicated now. Sure, there are a handful of arguments, but their meaning is pretty straightforward.

I have a small change request, please look at the comment. Other than that, this looks good to go.

skorch/callbacks/logging.py
timokau: Of course, I didn't realize that the Loggers have to adhere to sklearn's Estimator API as well. Thanks for catching! Please take another look.

BenjaminBossan: Yes, we built it that way to allow to grid search even on callback parameters.

BenjaminBossan (approved):
Thanks a lot, nice addition.

timokau:
Thanks for the review Benjamin!

PR #727, opened 2020-12-08
Remove the deprecated LRScheduler epoch parameter

timokau:
The parameter was deprecated, see [1].

[1] https://github.com/pytorch/pytorch/blob/dee82ef3ea1ed2cae4e891f071adff03cb4a6e0d/torch/optim/lr_scheduler.py#L13

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- We no longer pass teh `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
+- We no longer pass the `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
 
 ### Fixed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- We no longer pass teh `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
+
 ### Fixed
 
 ## [0.9.0] - 2020-08-30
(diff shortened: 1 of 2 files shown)


thomasjpfan (commented):
Thank you for the PR @timokau !

With this change to the API, we have to depend on the scheduler to keep track of the epoch state correctly. I can see this breaking someones code if they did not serialize  `lr_scheduler` correctly. But I think this would be an edge case.

May you add this to the [changelog](https://github.com/skorch-dev/skorch/blob/master/CHANGES.md)?

timokau:
I'm not sure I understand the edge case. The LR scheduler is initialized by the callback, so there isn't much the use can do wrong when initializing it. Or are you concerned about the case where a user adds the callback in the middle of training?

timokau:
Oh, you are talking about serialization. Somehow I missed that.

I added a changelog entry, but I'm not sure if that is what you had in mind. Please have a look, I'm happy to amend it if you think it should say something else. Feel free to squash the changes once they look good to you.

BenjaminBossan:
> I can see this breaking someones code if they did not serialize `lr_scheduler` correctly

Thomas, could you elaborate on this?

CHANGES.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- We no longer pass teh `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
```
BenjaminBossan, line 16:
```suggestion
- We no longer pass the `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
```

thomasjpfan:
> Thomas, could you elaborate on this?

If someone were to use `load_params` and `save_params` and forget to load/save the optimizer. I think this is mostly an edge case tho.

BenjaminBossan:
> If someone were to use `load_params` and `save_params` and forget to load/save the optimizer. I think this is mostly an edge case tho.

Is that because the lr scheduler has different attributes based on whether it's called with or without the `epoch` argument?

Anyway, if you think it's acceptable, please go ahead and merge @thomasjpfan.

thomasjpfan:
> Is that because the lr scheduler has different attributes based on whether it's called with or without the epoch argument?

The epoch parameter effectively allowed the lr scheduler to be more stateless.

timokau:
Thanks @thomasjpfan!

PR #729, opened 2020-12-13
Try to fix failing docs

BenjaminBossan:
Build fails with message that Cython>=0.28.5 needs to be installed to build sklearn. Let's add this dependency.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,3 +2,4 @@ numpy>=1.13.3
 https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
+Cython>=0.28.5
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,4 +2,4 @@ numpy>=1.13.3
 https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
-Cython>=0.28.5
+scikit-learn>=0.19.1


BenjaminBossan:
This seems to fix the problem: https://readthedocs.org/projects/skorch/builds/12551153/

thomasjpfan:
It is weird because readthedocs is:

1. Installing the RC
2. Even with the RC, [the wheels](https://pypi.org/project/scikit-learn/0.24.0rc1/#files) should be used instead of the source.

BenjaminBossan:
Yeah, it's not clear to me why that was necessary. Any idea how else to fix the failing docs build?

docs/requirements.txt
```diff
@@ -2,3 +2,4 @@ numpy>=1.13.3
 https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
+Cython>=0.28.5
```
thomasjpfan, line 5: It is really strange how `setup.py` tries to build the wheel from source. I have seen this before but I have not figured out the root case of it.

I think another way to fix it is to install scikit-learn before `setup.py` gets to run:

```suggestion
scikit-learn>=0.19.1
```

BenjaminBossan: That did it: https://readthedocs.org/projects/skorch/builds/12622129/

thomasjpfan:
I think having scikit-learn in the requirements fixes the issue: https://readthedocs.org/projects/skorch/builds/12622129/

BenjaminBossan:
> I think having scikit-learn in the requirements fixes the issue: https://readthedocs.org/projects/skorch/builds/12622129/

Yes, I mentioned that [here](https://github.com/skorch-dev/skorch/pull/729#discussion_r547569053) but the comment is no longer being shown since it's marked as resolved. From my point of view, this can be merged.

Issue #674, opened 2020-07-28
Parameter max_epochs isn't respected from history


karmus89:
Looking at the code at NeuralNet.fit_loop-function:

https://github.com/skorch-dev/skorch/blob/7a845682be15956a132c3817f0431177a620fee5/skorch/net.py#L714

From what I understand, the max_epochs is actually used in a once-in-fit fashion. With early stopping and continued training however what I'd expect would be that the model is trained only for max_epochs regardless. 

In other words and my opinion, the training should not proceed further than what has been defined by max_epochs (i.e. net.history[-1].epoch should never exceed net.max_epochs).

thomasjpfan:
Thank you for raising this issue!

I agree the current behavior of `max_epochs` is strange. It is not consistent with `sklearn`, for example:

```py
from sklearn.ensemble import RandomForestRegressor
from sklearn.datasets import make_regression
X, y = make_regression(n_features=4, n_informative=2,
                       random_state=0, shuffle=False)
regr = RandomForestRegressor(n_estimators=100,
                             max_depth=2, random_state=0,
                             warm_start=True)
regr.fit(X, y)
len(regr.estimators_)
# 100

regr.fit(X, y)
# UserWarning: Warm-start fitting without increasing n_estimators does not fit new trees.
```

karmus89:
Would changing this behavior break something somewhere else?

https://github.com/skorch-dev/skorch/blob/7a845682be15956a132c3817f0431177a620fee5/skorch/net.py#L714-L724

I'm thinking in the lines of making the iteration check against the `history` attribute for the last epoch number. But then again, it seems to be modified via an event listener and would be more on the side of "implied and expected" rather than clearly written out code. 

With that said, the `fit_loop` calls out the `on_epoch_begin` of the `NeuralNet`, which in turn assigns a new epoch.

What are your thoughts on this?

thomasjpfan:
This change is technically possible.

Since this is a backward incompatible change, we have to decide if this is worth a deprecation cycle.

BenjaminBossan:
Yes, we have a few divergences here from sklearn. On the one hand, as you said, the epoch counter in the fit loop will always start from 0, even if you've already cycled through a few epochs. Relatedly, calling `partial_fit` also goes through `max_epoch` epochs, instead of just 1.

I think in this case, we should think pragmatically. A few design choices in skorch diverge from sklearn because training a neural net often is a different beast than training a "regular" estimator.

[...]

PR #730, opened 2020-12-20
Feature/faster history getitem

BenjaminBossan:
Changes implementation of `History.__getitem__` so that the type of
items is only looked up once, not each time for each element. This
should be fine, as the types should not change between
elements (e.g. the rows in `'batches'`).

This can save a lot of time. In an example using 1000 samples, 500
epochs and a very small net, history access time goes down from 10 sec
to 2.5 sec.

Here is the code, provided by @stsievert in #719:

```python
import numpy as np
from torch import nn

class MyModule(nn.Module):
    def __init__(self, n_targets=85):
        super(MyModule, self).__init__()

        self.lin1 = nn.Linear(3, n_targets)
        self.lin2 = nn.Linear(n_targets, 2)

        X = self.softmax(self.lin1(X))
        X = self.softmax(self.lin2(X))
        return X

X, y = make_classification(n_samples=1000, n_features=3, n_redundant=1, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

net = NeuralNetClassifier(
    module=MyModule,
    max_epochs=500,
    lr=0.1,
    iterator_train__shuffle=True,
    verbose=0,
    train_split=False)

net.fit(X, y)
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed implementation of `net.history` access to make it faster; this should not have any noticeable effects, but if you encounter bugs, please create an issue
+
 ### Fixed
 
 ## [0.9.0] - 2020-08-30
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - We no longer pass the `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
-- Changed implementation of `net.history` access to make it faster; this should not have any noticeable effects, but if you encounter bugs, please create an issue
+- Changed implementation of `net.history` access to make it faster; this should result in a nice speedup when dealing with very small model/data but otherwise not have any noticeable effects; if you encounter bugs, though, please create an issue
 
 ### Fixed
 
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -33,15 +33,14 @@ def _getitem_list_list(items, keys):
     filtered = []
     for item in items:
         row = []
-        do_append = True
         for key in keys:
             try:
                 row.append(item[key])
             except KeyError:
-                do_append = False
                 break
-        if row and do_append:
-            filtered.append(row)
+        else:  # no break
+            if row:
+                filtered.append(row)
     return filtered
 
 
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -108,10 +108,6 @@ def _get_getitem_method(items, key):
     * history[0, 'foo', :, ('bar', 'baz')]: key is list/tuple of str
 
     """
-
-    if isinstance(items, tuple):
-        raise ZeroDivisionError
-
     if isinstance(items, list):
         if isinstance(key, list):
             return _getitem_list_list
@@ -119,7 +115,7 @@ def _get_getitem_method(items, key):
             return _getitem_list_tuple
         if isinstance(key, str):
             return _getitem_list_str
-        raise ZeroDivisionError
+        raise TypeError("History access with given types not supported")
 
     if isinstance(items, dict):
         if isinstance(key, list):
@@ -128,7 +124,7 @@ def _get_getitem_method(items, key):
             return _getitem_dict_tuple
         if isinstance(key, str):
             return _getitem_dict_str
-    raise ZeroDivisionError
+    raise TypeError("History access with given types not supported")
 
 
 def _unpack_index(i):
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -39,7 +39,7 @@ def _getitem_list_list(items, keys):
                 row.append(item[key])
             except KeyError:
                 do_append = False
-                continue
+                break
         if row and do_append:
             filtered.append(row)
     return filtered
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -23,17 +23,112 @@ def _filter_none(items):
     return type_(filter(_not_none, items))
 
 
-def _getitem(item, i):
-    """Extract value or values from dicts.
+def _getitem_list_list(items, keys):
+    """Ugly but efficient extraction of multiple values from a list of
+    items.
 
-    Covers the case of a single key or multiple keys. If not found,
-    return placeholders instead.
+    Keys are contained in a list.
 
     """
-    if not isinstance(i, (tuple, list)):
-        return item.get(i, _none)
-    type_ = list if isinstance(item, list) else tuple
-    return type_(item.get(j, _none) for j in i)
+    filtered = []
+    for item in items:
+        row = []
+        do_append = True
+        for key in keys:
+            try:
+                row.append(item[key])
+            except KeyError:
+                do_append = False
+                continue
+        if row and do_append:
+            filtered.append(row)
+    return filtered
+
+
+def _getitem_list_tuple(items, keys):
+    """Ugly but efficient extraction of multiple values from a list of
+    items.
+
+    Keys are contained in a tuple.
+
+    """
+    filtered = []
+    for item in items:
+        row = ()
+        do_append = True
+        for key in keys:
+            try:
+                row += (item[key],)
+            except KeyError:
+                do_append = False
+                break
+        if row and do_append:
+            filtered.append(row)
+    return filtered
+
+
+def _getitem_list_str(items, key):
+    filtered = []
+    for item in items:
+        try:
+            filtered.append(item[key])
+        except KeyError:
+            continue
+    return filtered
+
+
+def _getitem_dict_list(item, keys):
+    return [item.get(key, _none) for key in keys]
+
+
+def _getitem_dict_tuple(item, keys):
+    return tuple(item.get(key, _none) for key in keys)
+
+
+def _getitem_dict_str(item, key):
+    return item.get(key, _none)
+
+
+def _get_getitem_method(items, key):
+    """Return method to extract values from items.
+
+    For the given type of items and type of keys, find the correct
+    method to extract the values. By calling this only once per items,
+    we can save a lot of type checking, which can be slow if there are
+    a lot of epochs and a lot of items. However, we now make the
+    assumption that the type of items doesn't change (we know that the
+    key doesn't change). This should always be true, except if
+    something really weird happens.
+
+    We are multi-dispatching based on the following possibilities:
+
+    * history[0, 'foo', :10]: get a list of items
+    * history[0, 'foo', 0]: get a dict
+    * history[0, 'foo', :, 'bar']: key is a str
+    * history[0, 'foo', :, ('bar', 'baz')]: key is list/tuple of str
+
+    """
+
+    if isinstance(items, tuple):
+        raise ZeroDivisionError
+
+    if isinstance(items, list):
+        if isinstance(key, list):
+            return _getitem_list_list
+        if isinstance(key, tuple):
+            return _getitem_list_tuple
+        if isinstance(key, str):
+            return _getitem_list_str
+        raise ZeroDivisionError
+
+    if isinstance(items, dict):
+        if isinstance(key, list):
+            return _getitem_dict_list
+        if isinstance(key, tuple):
+            return _getitem_dict_tuple
+        if isinstance(key, str):
+            return _getitem_dict_str
+    raise ZeroDivisionError
 
 
 def _unpack_index(i):
@@ -200,14 +295,9 @@ class History(list):
 
         # extract keys of batches
         # handles: history[..., k_e, i_b][k_b]
-        if k_b is not None:
-            items = [
-                _filter_none([_getitem(b, k_b) for b in batches])
-                if isinstance(batches, (list, tuple))
-                else _getitem(batches, k_b)
-                for batches in items
-            ]
-            # get rid of empty batches
+        if items and (k_b is not None):
+            extract = _get_getitem_method(items[0], k_b)
+            items = [extract(batches, k_b) for batches in items]
             items = [b for b in items if b not in (_none, [], ())]
             if not _filter_none(items):
                 # all rows contained _none or were empty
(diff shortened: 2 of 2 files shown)


stsievert:
Would it be easier to disable Skorch history recording? In my use case, I'm only concerned about performance and don't care at all about history/callbacks/etc.  I'd be okay with this solution:

``` python
est = NeuralNetClassifier(..., history=False)  # disables recording of history
```

At first, I relied on Skorch for entire training process before realizing that Skorch added a lot of overhead, specifically with callbacks and data conversion from NumPy arrays to PyTorch datasets.

[s]:https://jiffyclub.github.io/snakeviz/
[d]:https://distributed.dask.org/en/latest/diagnosing-performance.html#statistical-profiling

This motivated only using Skorch for initialization and writing a new class that overrode `NeuralNet.partial_fit` and called `torch.from_numpy` before feeding through Skorch's `module_`. I encapsulate this implementation of `partial_fit` in a class `NoOverhead`, which expects NumPy arrays to `NoOverhead.partial_fit`:

``` python
class NoOverhead(NeuralNetRegressor):
    def __init__(self, *args, random_state=None, **kwargs):
        self.random_state = random_state

    def partial_fit(self, X: np.ndarray, y: np.ndarray):
        if not self.initialized_:
            self.initialize()
        seed = int(time() * 1000) % (2 ** 30)
        rng = check_random_state(seed)
        X2 = torch.from_numpy(shuffle(X, seed))
        y2 = torch.from_numpy(shuffle(y, seed))

        for k in range(len(X) // self.batch_size):
            lower, upper = k * self.batch_size, (k + 1) * self.batch_size % len(X2)
            y_hat = self.module_.forward(X2[lower:upper])

            loss = self.criterion_(y2[lower:upper], y_hat)

            self.optimizer_.zero_grad()
            loss.backward()
            self.optimizer_.step()
            self.optimizer_.zero_grad()
        return self
```

The complete benchmarking script is a modification of https://github.com/skorch-dev/skorch/pull/730#issue-543098550, and uses similar datasets/models/etc. I use [Dask's statistical profiling tool][d] to profile the timing and see which functions are taking significant time.

<details>
<summary>Complete benchmarking script:</summary>

``` python
import numpy as np
from torch import nn
from dask.distributed import Client
from skorch import NeuralNetClassifier, NeuralNet, NeuralNetRegressor
from sklearn.utils import check_random_state
import numpy as np
import torch
from time import time

class MyModule(nn.Module):
    def __init__(self, n_targets=85):
        super(MyModule, self).__init__()

        self.lin1 = nn.Linear(3, n_targets)
        self.lin2 = nn.Linear(n_targets, 1)

        X = self.softmax(self.lin1(X))
        X = self.softmax(self.lin2(X))
        return X


def shuffle(X: np.ndarray, seed: int) -> np.ndarray:
    Y = X.copy()
    np.random.RandomState(seed).shuffle(Y)
    return Y

class NoOverhead(NeuralNetRegressor):
    def __init__(self, *args, random_state=None, **kwargs):
        self.random_state = random_state

    def partial_fit(self, X: np.ndarray, y: np.ndarray):
        if not self.initialized_:
            self.initialize()
        seed = int(time() * 1000) % (2 ** 30)
        rng = check_random_state(seed)
        X2 = torch.from_numpy(shuffle(X, seed))
        y2 = torch.from_numpy(shuffle(y, seed))

        for k in range(len(X) // self.batch_size):
            lower, upper = k * self.batch_size, (k + 1) * self.batch_size % len(X2)
            y_hat = self.module_.forward(X2[lower:upper])

            loss = self.criterion_(y2[lower:upper], y_hat)

            self.optimizer_.zero_grad()
            loss.backward()
            self.optimizer_.step()
            self.optimizer_.zero_grad()
        return self

def main():
    X, y = make_regression(n_samples=1000, n_features=3, random_state=0)
    X = X.astype(np.float32)
    y = y.astype(np.float32).reshape(-1, 1)

    #  est = NoOverhead(
    est = NeuralNetRegressor(
        module=MyModule,
        max_epochs=100,
        criterion=nn.MSELoss,
        batch_size=128,
        lr=0.1,
        iterator_train__shuffle=False,
        train_split=False)
    est.set_params(verbose=0, callbacks__train_loss=None)
    est.callbacks_ = []
    start = time()
    est.partial_fit(X, y)
    return time() - start

if __name__ == "__main__":
    client = Client("localhost:8786")
    future = client.submit(main)
    t = future.result()
    print(t)
```

</details>

 The results:

* `NeuralNetRegressor`: 1.58s.
* `NoOverhead`: 0.00616s, or 6.16ms.

`NeuralNetRegressor` spend the half (or 52.7%) of the time in PyTorch doing data conversions ([dataloader.py#L521][1]). It spends the other half on the core training in `step_fn`. There must be some other overhead in that method.


[1]:https://github.com/pytorch/pytorch/blob/c0deb231db76dbea8a9d326401417f7d1ce96ed5/torch/utils/data/dataloader.py#L521
[py-spy]:https://github.com/benfred/py-spy

BenjaminBossan:
Interesting analysis. A minor amendment: Your `NoOverhead` code runs only a single epoch, whereas the vanilla skorch code runs 100 epochs, so the difference is not quite as drastic as it may seem.

I replicated your results and made some minor adjustments (e.g. made the model actually learn something on the data to verify that all approaches end in a similar result). For measuring performance, I used snakeviz (average of 3 runs). I worked on this branch (i.e. fast history access). Here is what I found:

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-right" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">condition</th>
<th scope="col" class="org-right">timing</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">vanilla</td>
<td class="org-right">3.0</td>
</tr>

<tr>
<td class="org-left">verbose=0</td>
<td class="org-right">2.1</td>
</tr>

<tr>
<td class="org-left">verbose=0, train_loss=None</td>
<td class="org-right">2.0</td>
</tr>

<tr>
<td class="org-left">verbose=0, history disabled</td>
<td class="org-right">1.9</td>
</tr>

<tr>
<td class="org-left">callbacks disabled</td>
<td class="org-right">1.7</td>
</tr>

<tr>
<td class="org-left">callbacks disabled, fit with tensors*</td>
<td class="org-right">1.6</td>
</tr>

<tr>
<td class="org-left">callbacks disabled, fit with TensorDataset*</td>
<td class="org-right">1.1</td>
</tr>

<tr>
<td class="org-left">no overhead net</td>
<td class="org-right">0.3</td>
</tr>
</tbody>
</table>

\*It's possible to fit skorch nets with torch tensors or torch datasets.

So there is still a pretty big difference, although not nearly as big as initially reported.

What's nice to see is that disabling history is not such a big deal anymore with the proposed changes.

When it comes to the normal PyTorch machinery (forward, backward, etc.), it seems that there is no significant difference between the no-overhead vs normal skorch code. The biggest difference comes through the `DataLoader`, as you also observed. 

Your code completely side-steps PyTorch's `Dataset` and `DataLoader` part. We knew right from the start that these parts would add an overhead. `Dataset` only allows single sample access via `__getitem__`. So effectively, we need to iterate over all individual rows, which are later collated again inside `DataLoader`. Your code iterates over batches, saving a lot of computation.

Could we do away with `Dataset` and `DataLoader` in skorch? I don't think so. They are essential parts of the PyTorch ecosystem, many libraries like torchvision make use of `Dataset`. Once you have `Dataset`s, you need something akin to `DataLoader`. Maybe it would be nice if `Dataset.__getitem__` supported passing slices, not just ints, so that single row access was no longer necessary. But we cannot change that. On top of that, `DataLoader` has some nice properties, like batching of "complex" data (say, a dict of tensors), shuffling, multiple workers etc. If we got rid of it, we would lose this functionality (and might have to re-implement it).

skorch was designed from the start to be PyTorch + extra. Obviously, this extra would have to cost something. We made the assumption that the cost of forward pass, gradient computation, and parameter updates would dominate, and those cannot be avoided.

Still, if we can find a way to squeeze out more performance without making skorch too complex, we should do so (hence this PR). E.g. for this benchmark, I could see that `skorch.dataset.Dataset.__getitem__` is twice as slow as `torch.utils.data.TensorDataset.__getitem__`, adding 0.4 sec to the total. Of course, our `Dataset` has to deal with a lot of different data types, but maybe we can still find room for improvement.

Would it be possible to easily sidestep the whole `Dataset` + `DataLoader` machinery for simple cases? I don't think so without introducing massive changes. This is probably part of the extra that people opt into when using skorch compared to pure PyTorch.

What we should do no matter what is adding some documentation about performance overhead. I think a few of the findings in this thread would fit there nicely. As the table above shows, you can already get from 3 sec to 1.1 sec with a few tweaks.

stsievert:
> A minor amendment: Your NoOverhead code runs only a single epoch, whereas the vanilla skorch code runs 100 epochs, so the difference is not quite as drastic as it may seem.

Thanks for that – I knew the results were too drastic. I can reproduce your numbers, and verified that `partial_fit` runs for `self.max_epochs` epochs (which does not mirror [SGDClassifier][1] as expected).

> Would it be possible to easily sidestep the whole Dataset + DataLoader machinery for simple cases? I don't think so without introducing massive changes. This is probably part of the extra that people opt into when using skorch compared to pure PyTorch.

I have some ideas on how to make that happen (having [get_dataset][2] return an iterable, than catching that in [get_iterator][3]). Hopefully won't require too many compromises (I suspect `tran_split is None` will be necessary). I'll try to submit that PR.

[3]:https://github.com/skorch-dev/skorch/blob/7c76e78238e7edd2bd78afd29a16916438bd4f38/skorch/net.py#L1318
[2]:https://github.com/skorch-dev/skorch/blob/7c76e78238e7edd2bd78afd29a16916438bd4f38/skorch/net.py#L1212
[1]:https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html

BenjaminBossan:
> (which does not mirror [SGDClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html) as expected)

That's [another discussion](https://github.com/skorch-dev/skorch/issues/674).

> I have some ideas on how to make that happen

Yes, I think overriding the mentioned methods could make it feasible to sidestep the whole `Dataset` and `DataLoader` issue by returning an iterator. Depending on requirements, it would need to take care of batching, shuffling, sampling etc. I still think it's probably not feasible to provide this as a general skorch option without causing trouble ("why is `iterator_train__foo=bar` not being used by my iterator?").

> I'll try to submit that PR.

Also take a look at [`get_split_dataset`](https://github.com/skorch-dev/skorch/blob/1c8fc61364d04addebdc89ec4597a1f88e3e50db/skorch/net.py#L1263), which is being called in the fit loop.

ottonemo (approved):

CHANGES.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - We no longer pass the `epoch` parameter to LR schedulers, since that parameter has been deprecated. We now rely on the scheduler to keep track of the epoch itself.
+- Changed implementation of `net.history` access to make it faster; this should not have any noticeable effects, but if you encounter bugs, please create an issue
```
stsievert, line 20:
```suggestion
- Changed implementation of `net.history` access to make it faster, which should be most noticeable for small models/datasets.
```

skorch/history.py
```diff
@@ -23,17 +23,108 @@ def _filter_none(items):
...
+
+def _getitem_dict_tuple(item, keys):
+    return tuple(item.get(key, _none) for key in keys)
+
+
+def _getitem_dict_str(item, key):
+    return item.get(key, _none)
+
+
+def _get_getitem_method(items, key):
```
thomasjpfan, line 92: I think we can take advantage of https://docs.python.org/3/library/functools.html#functools.singledispatch here instead of dispatching ourselves.

skorch/history.py
```diff
@@ -23,17 +23,108 @@ def _filter_none(items):
...
+        row = []
+        do_append = True
+        for key in keys:
+            try:
+                row.append(item[key])
+            except KeyError:
+                do_append = False
+                break
+        if row and do_append:
+            filtered.append(row)
```
thomasjpfan, line 44: No strong opinion:

```py
        for key in keys:
            try:
                row.append(item[key])
            except KeyError:
                break
        else:  # no break
        	if row:
        		filtered.append(row)
```

skorch/history.py
BenjaminBossan: Yes, I actually used implemented it both ways. After running a profiler, I saw that there was no difference (presumably `singledispatch` does more or less the same work under the hood).

The main reasons why I didn't go for `singledispatch`:

- we technically need multiple dispatch (`key` and `items`)
- it can look a bit daunting (having to add `@fun.register` everywhere, no nice syntax as in, say, Julia)
- it's presumably not so well known, so harder for readers to understand what's happening, whereas currently, this suggestion only uses "basic" Python language features

What would be the advantages of using `singledispatch` in your opinion?

skorch/history.py
BenjaminBossan: Yes, absolutely! I will fix it.

CHANGES.md
BenjaminBossan: Yes, thanks, for sure this entry should mention what the benefit is. I included it but didn't use exactly your proposal.

skorch/history.py
stsievert: I haven't looked too closely at the implementation, but I think one benefit of Python's singledispatch is that better documentation is available. Additionally, the concise naming will very likely help new developers like me find relevant examples/SO questions.

BenjaminBossan: > one benefit of Python's singledispatch is that better documentation is available

Do you mean the documentation for `singledispatch` itself? Arguably, the current code doesn't even need consulting documentation as long as you know what `if isinstance(...)` does :) But readability is of course debatable.

However, consider that for this specific case, we dispatch on two arguments, so we would need either nested `singledispatch` or a mix of `singledispatch` and `isinstance`. If we were dispatching only on one argument or if there were a `multidispatch`, I would tend to agree that this approach is superior.

stsievert: Yeah, nevermind. I misunderstood what the comment was referring too. Multiple `isinstance` blocks is a lot clearer.

thomasjpfan: Lets leave it as is. The current implementation is already a net benefit.

thomasjpfan (approved):

PR #732, opened 2020-12-22
Add performance section to docs

BenjaminBossan:
Documents a few was on how to reduce skorch overhead when fitting models.

Based on [this comment](https://github.com/skorch-dev/skorch/pull/730#issuecomment-748932720).

Also, no longer use outdated pytorch version for docs.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 numpy>=1.13.3
-https://download.pytorch.org/whl/torch_stable.html
+https://download.pytorch.org/whl/cpu/torch-1.7.1%2Bcpu-cp38-cp38-linux_x86_64.whl
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 numpy>=1.13.3
-https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
+https://download.pytorch.org/whl/torch_stable.html
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 numpy>=1.13.3
-https://download.pytorch.org/whl/cpu/torch-1.7.1%2Bcpu-cp38-cp38-linux_x86_64.whl
+https://download.pytorch.org/whl/cpu/torch-1.7.1%2Bcpu-cp37-cp37m-linux_x86_64.whl
 sphinx-rtd-theme==0.4.0
 numpydoc==0.8.0
--- a/docs/user/performance.rst
+++ b/docs/user/performance.rst
@@ -2,9 +2,9 @@
 Performance
 ===========
 
-Since skorch provides a few extras on top of a pure PyTorch training code, it is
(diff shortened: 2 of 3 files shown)


docs/user/performance.rst
```diff
@@ -0,0 +1,122 @@
...
+   # skorch version 0.9 or earlier
+   net = NeuralNet(...)
+   net.initialize()
+   net.callbacks_ = []  # manually remove all callbacks
+   net.fit(X, y)
+
+Instead of turning off all callbacks, you can also turn off specific callbacks,
+including default callbacks. This way, you can decide which ones to keep and
+which ones to get rid off. Typically, callbacks that calculate some kind of
+metric tend to be slow.
```
thomasjpfan, line 81: I am thinking this is an performance issue because data has to be moved to the CPU to compute metrics.

docs/user/performance.rst
```diff
@@ -0,0 +1,122 @@
...
+verbose=3)``).
+
+Turning off verbosity
+---------------------
+
+By default, skorch produces a print log of the training progress. This is useful
+for checking the training progress, monitor overfitting, etc. If you don't need
+these diagnostics, you can turn them off via the ``verbose`` parameter. This
+way, printing is deactivated, saving time on i/o. You can still access the
+diagnostics through the ``history`` attribute after training has finished.
```
thomasjpfan, line 46: From https://github.com/skorch-dev/skorch/pull/730#issuecomment-748932720, I am surprised that setting `verbose=0` makes the time go from 3.0 to 2.1. We do not really use `self.verbose` in many places.

thomasjpfan (commented):
We have been receiving performance related issues from time to time. This addition to the documentation would help.

docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.13.3
-https://download.pytorch.org/whl/cpu/torch-1.1.0-cp37-cp37m-linux_x86_64.whl
+https://download.pytorch.org/whl/torch_stable.html
```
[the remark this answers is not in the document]
BenjaminBossan: pytorch-install-no-cpu

Yes, I blundered when copying the install instructions.

docs/user/performance.rst
```diff
@@ -0,0 +1,122 @@
+===========
+Performance
+===========
+
+Since skorch provides a few extras on top of a pure PyTorch training code, it is
```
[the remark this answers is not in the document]
BenjaminBossan: I'll change it to: "Since skorch provides extra functionality on top of a pure PyTorch training ..."

docs/user/performance.rst
BenjaminBossan: Indeed. However, I ran my tests on CPU and there was still a measurable overhead, so it's not solely that.

docs/user/performance.rst
BenjaminBossan: > We do not really use self.verbose in many places.

One idea with `verbose` is that users can easily write their own callbacks with some helpful debugging print calls that they may turn off with a single parameter. Whether many users do, I can't say.

> I am surprised that setting verbose=0 makes the time go from 3.0 to 2.1

My guess with this specific test is that the overhead depends on the speed at which the environment flushes the print. Not sure if it's blocking or not. I.e. if you have a faster terminal, the overhead could be less. But it's just a guess.

thomasjpfan (approved):
LGTM

PR #733, opened 2020-12-22
Maintenance: python and pytorch versions updated

BenjaminBossan:
Supersedes #686 

- use github actions instead of travis
- add CI for PyTorch 1.4.0 (since we want to support 4 versions)
- check PyTorch 1.7.1 instead of 1.7.0
- update docs
- remove warning about Python 3.5 support
- remove warning about change in CyclicLR step being per epoch


thomasjpfan (approved):
LGTM

Issue #734, opened 2021-01-05
ClassWeighted NeuralNetClassifier calculates loss on CPU regardless of device


vlawhern:
Hello,

I'm trying to set up a NeuralNetClassifier with class weights, following the FAQ at (https://skorch.readthedocs.io/en/stable/user/FAQ.html#i-want-to-use-sample-weight-how-can-i-do-this). However, I get the following error when I try to fit my model on a GPU:

``RuntimeError: expected device cpu but got device cuda:0``

Here's a sample code to reproduce the issue (taken primarily from the skorch example):

```python
import numpy as np
from sklearn.datasets import make_classification
from torch import nn

from skorch import NeuralNetClassifier


X, y = make_classification(1000, 20, n_informative=10, random_state=0)
X = X.astype(np.float32)
train = {'X': X}
train['sample_weight'] = np.random.randint(1,3,size=(1000,)) # sets weights random as either 1 or 2
y = y.astype(np.int64)

class MyModule(nn.Module):
    def __init__(self, num_units=10, nonlin=nn.ReLU()):
        super(MyModule, self).__init__()

        self.dense0 = nn.Linear(20, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(0.5)
        self.dense1 = nn.Linear(num_units, num_units)
        self.output = nn.Linear(num_units, 2)
        self.softmax = nn.Softmax(dim=-1)

    def forward(self, X, **kwargs):
        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = self.nonlin(self.dense1(X))
        X = self.softmax(self.output(X))
        return X



class ClassWeighted(NeuralNetClassifier):
    def __init__(self, *args, criterion__reduce=False, **kwargs):
        # make sure to set reduce=False in your criterion, since we need the loss
        # for each sample so that it can be weighted
        super().__init__(*args, criterion__reduce=criterion__reduce, **kwargs)

    def get_loss(self, y_pred, y_true, X, *args, **kwargs):
        # override get_loss to use the sample_weight from X
        loss_unreduced = super().get_loss(y_pred, y_true, X, *args, **kwargs)
        sample_weight = X['sample_weight']
        loss_reduced = (sample_weight * loss_unreduced).mean()
        return loss_reduced


net = ClassWeighted(
    MyModule,
    max_epochs=10,
    lr=0.1,
    device='cuda',
    # Shuffle training data on each epoch
    iterator_train__shuffle=True,
)

[...]

PR #736, opened 2021-01-19
Address #734: move sample weights to self.device

ottonemo:
The code previously shown was just able to run on nets trained
on CPU, not GPUs.


thomasjpfan (approved):
LGTM

Issue #363, opened 2018-10-13
ReduceLRonPlateau does not work as expected


vykhand:
Hi everyone!

I noticed that the LR Scheduler callback with ReduceLRonPlateau does not work as expected. I expect it to reduce LR when the metric has not improved for N=**patience** epochs, but instead it reduces LR just after the fixed N number of epochs

For example:
(config.yaml)
```
scheduler_params: {factor: 0.5, min_lr: 1.0e-06, mode: max, patience: 15, verbose: 2}
```
(NeuralNet callback)
```
scheduler = LRScheduler(policy=scheduler_policy, monitor="val_ioi_best", **model_params["scheduler_params"])
```
or 
```
scheduler = LRScheduler(policy=scheduler_policy, monitor="val_iou", **model_params["scheduler_params"])
```

```
TRAIN PHASE ITER: 3
, global time: 3.29, mem use:0.44GB, GPU_MEM: 0.00GB
  epoch    train_iou    train_loss    val_iou    valid_loss    cp       dur
```
-------  -----------  ------------  ---------  ------------  ----  --------
      1       0.4349        0.2884     0.5459        0.2219     +  223.1961
      2       0.6177        0.1862     0.5850        0.2172     +  222.8284
      3       0.6567        0.1597     0.6741        0.1751     +  222.5643
      4       0.6951        0.1388     0.7109        0.1692     +  222.2332
      5       0.7023        0.1326     0.6809        0.2199        222.3694
      6       0.7277        0.1271     0.6794        0.1622     +  222.0340
      7       0.7418        0.1085     0.6894        0.1746        221.9729
      8       0.7542        0.1038     0.7041        0.1831        221.9098
      9       0.7776        0.0912     0.7253        0.1932        221.9841
     10       0.7814        0.0860     0.7434        0.1550     +  221.7720
     11       0.7841        0.0830     0.7262        0.1764        221.8531
     12       0.7972        0.0771     0.7581        0.1676        222.0086
     13       0.8117        0.0679     0.7438        0.1924        221.5925
     14       0.8099        0.0727     0.7128        0.2350        221.8342
     15       0.8027        0.0736     0.7150        0.2062        222.1696
     16       0.8042        0.0729     0.7447        0.1813        221.6657
Epoch    16: reducing learning rate of group 0 to 5.0000e-05.

Dewald928:
I don't know if this is a user error, but this isn't working as intended. It keeps changing the LR after patience has run out every 10 epochs.
[...]

PR #738, opened 2021-01-29
FIX Makes sure score is passed to ReduceLROnPlateau

thomasjpfan:
Closes https://github.com/skorch-dev/skorch/issues/363

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI helper function now also supports normal (i.e. non-skorch) sklearn estimators
 - Disabling all callbacks is now supported (which allows reducing overhead,
   which is especially relevant for small models).
+- `LRScheduler` now currently passes the score/loss in the history to `ReduceLROnPlateau`. (#738)
 
 ### Changed
 
@@ -35,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Removed support for schedulers with a `batch_step()` method in `LRScheduler`. 
+- Removed support for schedulers with a `batch_step()` method in `LRScheduler`.
 - Raise `FutureWarning` in `CVSplit` when `random_state` is not used. Will raise an exception in a future (#620)
 - The behavior of method `net.get_params` changed to make it more consistent with sklearn: it will no longer return "learned" attributes like `module_`; therefore, functions like `sklearn.base.clone`, when called with a fitted net, will no longer return a fitted net but instead an uninitialized net; if you want a copy of a fitted net, use `copy.deepcopy` instead;`net.get_params` is used under the hood by many sklearn functions and classes, such as `GridSearchCV`, whose behavior may thus be affected by the change. (#521, #527)
 - Raise `FutureWarning` when using `CyclicLR` scheduler, because the default behavior has changed from taking a step every batch to taking a step every epoch. (#626)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI helper function now also supports normal (i.e. non-skorch) sklearn estimators
 - Disabling all callbacks is now supported (which allows reducing overhead,
   which is especially relevant for small models).
-- `LRScheduler` now currently passes the score/loss in the history to `ReduceLROnPlateau`. (#738)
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - CLI helper function now also supports normal (i.e. non-skorch) sklearn estimators
 - Disabling all callbacks is now supported (which allows reducing overhead,
   which is especially relevant for small models).
+- `LRScheduler` now currently passes the score/loss in the history to `ReduceLROnPlateau`. (#738)
```
BenjaminBossan, line 16: Not sure if I understand this sentence correctly, maybe change the wording?

skorch/callbacks/lr_scheduler.py
```diff
@@ -149,12 +149,7 @@ def on_epoch_end(self, net, **kwargs):
             if callable(self.monitor):
                 score = self.monitor(net)
             else:
-                if self.lr_scheduler_.mode == 'max':
-                    score = -np.inf
-                elif self.lr_scheduler_.mode == 'min':
-                    score = np.inf
-                else:
-                    score = net.history[-1, self.monitor]
+                score = net.history[-1, self.monitor]
```
BenjaminBossan, line 152: What do you think: We could catch the `KeyError` and augment the error message by telling the user that they may possibly need to change the order of callbacks if they reference a metric that was not yet computed?

BenjaminBossan (approved):
Thanks for taking this. PR LGTM, just two minor comments.

skorch/callbacks/lr_scheduler.py
thomasjpfan: I think its a good idea. I updated the PR to include this nicer error message.

skorch/callbacks/lr_scheduler.py
```diff
@@ -149,7 +149,14 @@ def on_epoch_end(self, net, **kwargs):
             if callable(self.monitor):
                 score = self.monitor(net)
             else:
-                score = net.history[-1, self.monitor]
+                try:
+                    score = net.history[-1, self.monitor]
+                except KeyError:
+                    raise ValueError(
+                        f"'{self.monitor}' was not found in history. A "
+                        f"PassthroughScoring callback with name='{self.monitor}' "
```
BenjaminBossan, line 157:
```suggestion
                        f"Scoring callback with name='{self.monitor}' "
```

I would suggest to be more agnostic about which kind of scoring is used, since `EpochScoring` would work as well.

skorch/callbacks/lr_scheduler.py
```diff
@@ -149,12 +149,14 @@ def on_epoch_end(self, net, **kwargs):
...
                     score = net.history[-1, self.monitor]
+                except KeyError as e:
+                    raise ValueError(
+                        f"'{self.monitor}' was not found in history. A "
+                        f"PassthroughScoring callback with name='{self.monitor}' "
+                        "should be placed before the LRScheduler callback"
+                    ) from e
 
             self.lr_scheduler_.step(score)
             # ReduceLROnPlateau does not expose the current lr so it can't be recorded
```
jiajiexiao, line 162: Thanks for the fix in this PR, which I found while troubleshooting the same unexpected behavior as mentioned in the issue https://github.com/skorch-dev/skorch/issues/363. 

I noticed this comment but I guess how the learning rate was updated should be tracked in ``net.history`` even for `ReduceLROnPlateau`. As being discussed in https://discuss.pytorch.org/t/how-to-retrieve-learning-rate-from-reducelronplateau-scheduler/54234/3 and https://github.com/pytorch/pytorch/issues/2829, I would suggest replacing the comment line [162](https://github.com/skorch-dev/skorch/blob/cadb46914203cde74fcf08af0565cde8e059b0c0/skorch/callbacks/lr_scheduler.py#L162) with the following:

```
            if self.event_name is not None:
                net.history.record(self.event_name, self.lr_scheduler_._last_lr[0])
```

jiajiexiao (changes requested):

skorch/callbacks/lr_scheduler.py
thomasjpfan: We can discuss this in another PR since this is a feature unrelated to the bug fix.

BenjaminBossan: I created an issue (#739) so that we don't forget. @jiajiexiao does that reflect what you meant?

jiajiexiao: Sure. Thanks, @thomasjpfan @BenjaminBossan

PR #749, opened 2021-03-26
Fix 3 minor issues

BenjaminBossan:
- CHANGES entry of latest merge was under wrong heading
- Changed its link to pointing to latest instead of stable (otherwise
  only visible after next release)
- Replace spurious super class in `NeuralNetClassifier`

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -206,7 +206,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
         y_pred : numpy ndarray
 
         """
-        return super().predict_proba(X).argmax(axis=1)
+        return self.predict_proba(X).argmax(axis=1)
 
 
 neural_net_binary_clf_doc_start = """NeuralNet for binary classification tasks
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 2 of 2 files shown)


skorch/classifier.py
```diff
@@ -206,7 +206,7 @@ def predict(self, X):
         y_pred : numpy ndarray
 
         """
-        return super().predict_proba(X).argmax(axis=1)
+        return self.predict_proba(X).argmax(axis=1)
```
thomasjpfan, line 209: I still find it a little strange that the generic `NeuralNet` has a `predict_proba`, but this has been around for a long time.

Should this be a part of a bug fix release?

BenjaminBossan: Yes, we can remove the method from `NeuralNet` (and `NeuralNetRegressor`), but it should come with a deprecation cycle. Do you want to take this?

To improve compatibility with sklearn even further, we should also look into [tags](https://scikit-learn.org/stable/developers/develop.html#estimator-tags), which partly come from the *mixins we currently don't use.

thomasjpfan (approved):

skorch/classifier.py
BenjaminBossan: > I will take it.

Great, thanks.

Regarding tags: Thanks for the reference. I took a look at the existing tags and to me it seems that some of them we can know in advance (e.g. `requires_fit`, `binary_only` for `NeuralNetBinaryClassifier`) and some we cannot (say, `multioutput`). So we could try to make a fine grained distinction between what estimators should support what tags, but don't even know all the places that they are being used. It might therefore not be worth the effort.

PR #750, opened 2021-03-28
Don't re-initialize uninitialized net after calling set_params

BenjaminBossan:
Previously, when a parameter on, say, the module was changed via
`set_params` (e.g. `net.set_params(module__hidden_units=123)`),
`set_params` would always trigger (re-)initialization of the module. However,
when the net was not initialized in the first place, this is unnecessary. It
is sufficient to set the new attribute and wait for the net to be
initialized later.

Fortunately, this change doesn't seem to have any further impact, i.e.
we didn't implicitly rely on this behavior anywhere. The only exceptions
are 2 tests in `test_cli.py`, but those can easily be adjusted and this
shouldn't have any user impact.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1654,11 +1654,12 @@ class NeuralNet:
             # callbacks need special treatmeant since they are list of tuples
             self.initialize_callbacks()
             self._set_params_callback(**cb_params)
-
-        # if the net is not initialized, just set the attributes on the net and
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
I noticed a few more things I'd like to change when it comes to `set_params` and the initialization behavior. Closing therefore, will open later with these additional changes.

PR #751, opened 2021-04-01
Change: refactor skorch for more consistency when adding custom modules etc.

BenjaminBossan:
This PR is still WIP but as discussed with @ottonemo, I will create a draft PR early to discuss the changes and implementation.

# Motivation

The initial reason why I wanted to work on this is that I'm currently implementing a gpytorch integration (see [this branch](https://github.com/skorch-dev/skorch/tree/feature/gpytorch-integration)). For this, a big part is adding a new custom module called "likelihood". Doing this correctly was actually not trivial and required a lot of more or less duplicated code. Putting such a burden on a user with less experience with skorch would not be possible.

The main reason for this difficulty is that module, criterion and optimizer are treated "special" so far. We assume that they are already there and build everything else around this. If a custom module is added, the user needs to be aware of all the places where this is relevant, which is too error prone.

# Previous work

Some changes to facilitate adding custom modules were already implemented in #597. However, they don't go far enough.

# Changes

## Main changes

With this PR, we remove the special status of module, criterion and optimizer. Instead, all the work that needs to be done when adding any of them to the net is now implemented in a generic manner. This way, custom modules etc. can re-use the same functionality and can therefore expect to be treated the same as these "first class" components.

Here is a list of changed that were added to that effect:

- Until now, custom module parameters were not trained by the optimizer, now they are
- Until now, custom modules/criteria were not automatically moved to the indicated device, now they are
- Until now, custom modules/criteria were not automatically set into train/eval mode, now they are
- Simplified implementation of `initialize_module` et al. - they contained a lot of stuff that was irrelevant for the user, like messaging about why something was re-initialized; now this stuff is done inside the newly added methods `_initialize_module` etc., which are called by `initialize()` and shouldn't be a bother to the user
- Adding a custom module no longer requires the attribute name to contain the substring "module" (which was really not a nice solution), same for criterion and optimizer
- Re-initialization logic was changed: When any module is changed (via `set_params`), this triggers re-initialization of all modules, criteria and optimizers; when any criterion is changed, this triggers re-initialization of all optimizers (but not modules); this is a bit defensive since it could trigger unnecessary inits but it's better than missing any inits

## Additions

- There is now a `get_learnable_params` method on the net to retrieve all learnable parameters (instead of just those of `module_`); it is meant to be overridable by the user (e.g. when they have two optimizers for two modules)
- Added attributes `modules_`, `criteria_` and `optimizers_` to the net to keep track of those; first started as `OrderedDict`s to mirror `nn.Module`s, but that was flaky, as the values in the dict would often get out of sync with the attributes on the net
- If the criterion/criteria have learnable params, those are now passed to the optimizer as well (think GANs)

## Minor changes

- `net.set_params(...)` no longer initializes the net if it's not yet initialized - this was simply unnecessary and could lead to some unexpected behavior
- custom module instances now need to be set inside `initialize_module` (and the name must end on an underscore), else the user will get an appropriate error message; same logic for criterion and optimizer
- added a bunch of unit tests for the custom modules etc. that cover the cases not covered so far
- checking of kwargs is now done during `initialize`, not during `__init__` anymore, since at that point, we don't know yet what custom modules could exist
- run a `_check_kwargs` during `set_params` - previously, this was a loophole that allowed users to set params with typos etc.
- two unconditional print statements are now conditioned on verbosity level

## Notes

I took extra effort to write the code as clearly as possible and add lots of comments, since this touches some complicated parts of the code base. But if something is not obvious, please tell me so that I can improve the code since now it's still fresh in my mind.

You will see that a few of the existing tests have been changed to now call `initialize` on the net when previously they didn't. The reason is that some work like checking kwargs is now moved to `initialize`.

Also, you will see that some tests now use mocked modules to check for device calls. I found this preferable to actually moving to 'cuda' since this will also work without a cuda device (e.g. during CI).

# TODO

- [x] cover a few remaining edge cases
- [x] docstrings
- [x] documentation
- [x] update CHANGES
- [ ] possibly add a more elaborate example


BenjaminBossan:
I updated to documentation to reflect the changes being made. This should help with understanding how this PR will affect the user.

docs/user/customization.rst
```diff
@@ -96,3 +116,141 @@ perform some book keeping, like making sure that callbacks are handled
...
+        def initialize_criterion(self, *args, **kwargs):
+            super().initialize_criterion(*args, **kwargs)
+
+            # add an additional criterion called 'other_criterion_'
+            params = self.get_params_for('other_criterion')
+            self.other_criterion_ = nn.BCELoss(**params)
+            return self
+
+        def initialize_optimizer(self, *args, **kwargs):
+            # first initialize the normal optimizer
```
thomasjpfan, line 215: The example in the docstring for `get_learnable_params` says to override `get_learnable_params` and `initialize_optimizer` to do this. I think what we have here is simpler.

If a user adds their own `module2_`, they can still access them through `self.module2_.named_parameters`. What `get_learnable_params` provides is a nice wrapper to connect the optimizer and the module together, but it require a user to implement this connection.

skorch/net.py
```diff
@@ -230,6 +222,12 @@ class NeuralNet:
       listed attributes are mapped to CPU.  Expand this list if you
       want to add other cuda-dependent attributes.
 
+    modules_ : TODO
+
+    criteria_ : TODO
```
thomasjpfan, line 227: To stay consistent?

```suggestion
    criterias_ : TODO
```

skorch/net.py
```diff
@@ -567,56 +540,214 @@ def _apply_virtual_params(self, virtual_kwargs):
     def initialize_virtual_params(self):
         self.virtual_params_ = {}
 
-    def initialize_optimizer(self, triggered_directly=True):
+    def initialize_optimizer(self):
```
thomasjpfan, line 543: Changes public API, but should be okay?

skorch/net.py
```diff
@@ -1778,29 +1977,44 @@ def _register_attribute(
...
+
+        # Always copy the collections to avoid mutation
         if prefixes:
             self.prefixes_ = self.prefixes_[:] + [name]
 
         if cuda_dependent_attributes:
             self.cuda_dependent_attributes_ = (
                 self.cuda_dependent_attributes_[:] + [name + '_'])
 
+        if self.init_context_ == 'module':
```
thomasjpfan, line 2003: From what I can tell we need the context manager because of how we are using `__setattr__` to update the state of `modules_` and friends. Is this correct?

skorch/net.py
```diff
@@ -567,56 +540,214 @@ def _apply_virtual_params(self, virtual_kwargs):
...
+            self.initialize_optimizer()
+
+            self._register_virtual_param(
+                ['optimizer__param_groups__*__*', 'optimizer__*', 'lr'],
+                optimizer_setter,
+            )
+            return self
+
+    def _initialize_history(self):
+        with self._current_init_context('callbacks'):
```
thomasjpfan, line 737: I do not think we actually explicitly use of the `'callbacks'` or `virtual_params` contexts. Should we still have them?

thomasjpfan (commented):
Some early thoughts of this draft.

docs/user/customization.rst
BenjaminBossan: You are right, it is probably better to encourage users to directly use `module_.named_parameters()`. The `get_learnable_params` method could be relegated to basically being a convenience method to return all named parameters at once. We could therefore remove the `optimizer_name` argument, I'm on the fence what the best design would be here.

skorch/net.py
BenjaminBossan: "criteria" is already plural :)

skorch/net.py
BenjaminBossan: Indeed. In my memory, this argument was just added recently and thus I thought it's unlikely that anyone uses it. But it's actually been there for 2.5 years, so I will add a deprecation.

skorch/net.py
BenjaminBossan: Yes, exactly. This is my proposal how to "know" what kind of attribute we're dealing with, the old one being to infer from the attribute name, which I find inferior. A small disadvantage is that the context is not triggered if a user were to call `initialize_module` et al directly, though I don't believe there is a need for that.

skorch/net.py
BenjaminBossan: Good question. My reasoning was to stay consistent, but I don't see a use case at the moment. I could leave the context managers there but add a comment that they're not used at the moment.

BenjaminBossan:
Since I received positive feedback about the direction of this PR, I completed the remaining TODOs and changed the status from draft PR to PR.

I could add a more elaborate example of using custom modules/criteria/optimizers, say, in the `Advanced_Usage.ipynb`, not sure how necessary that is. If someone has any ideas, that would help.

I locally worked on the Gaussian Processes feature based on the head of this branch and it indeed helps to greatly simplify the code and reduce the number of edge cases one needs to keep track of. Therefore, I believe this change is moving skorch in the right direction. It does add some complexity on the implementation but it reduces the complexity on the user side, especially when it comes to hard to track bugs (e.g. a custom module not being updated).

skorch/net.py
```diff
@@ -492,57 +502,52 @@ class Dummy:
...
+            # criterion already initialized and no params changed:
+            self.criterion_ = self.criterion
+            return
+
+        if is_init:
+            criterion = type(criterion)
+
+        # pylint: disable=attribute-defined-outside-init
+        self.criterion_ = criterion(**kwargs)
+        return self
```
ottonemo, line 529: Can't we refactor the initialization check into its own function?

```python
def initialized_instance(instance_or_cls, kwargs):
    is_init = isinstance(instance_or_cls, torch.nn.Module):
    if is_init and not kwargs:
        return instance_or_cls
    if is_init:
        return type(instance_or_cls)
    return instance_or_cls(**kwargs)
```

`initialize_criterion` might look like this then:
```
kwargs = self.get_params_for('criterion')
self.criterion_ = initialized_instance(self.criterion, kwargs)
```

A user who wants full control may opt to not use it but it would make the code a lot more compact at the cost of removing transparency.

ottonemo (commented):
I skimmed through the PR. I think this looks promising. I am a little worried about the statefulness of `__setattr__` / `initialize_{optimizer,module,criterion}` but the risk seems negligible.

skorch/net.py
```diff
@@ -567,56 +540,223 @@ def _apply_virtual_params(self, virtual_kwargs):
...
+                    msg = self._format_reinit_msg(
+                        "optimizer", kwargs, triggered_directly=False)
+                print(msg)
+
+            self.initialize_optimizer()
+
+            self._register_virtual_param(
+                ['optimizer__param_groups__*__*', 'optimizer__*', 'lr'],
+                optimizer_setter,
+            )
```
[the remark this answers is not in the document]
BenjaminBossan: Yes, good catch, I fixed this. For `lr`, I decided to attach it to `optimizer`.

docs/user/customization.rst
```diff
@@ -96,3 +116,141 @@ perform some book keeping, like making sure that callbacks are handled
...
+The method :func:`~skorch.net.NeuralNet.initialize` is responsible for
+initializing all the components needed by the net, e.g. the module and
+the optimizer. For this, it calls specific initialization methods,
+such as :func:`~skorch.net.NeuralNet.initialize_module` and
+:func:`~skorch.net.NeuralNet.initialize_optimizer`. If you'd like to
+customize the initialization behavior, you should override the
+corresponding methods. Following sklearn conventions, the created
+components should be set as an attribute with a trailing underscore as
+the name, e.g. ``module_`` for the initialized module. Finally, the
+method should return ``self``.
```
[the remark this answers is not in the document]
BenjaminBossan: You are right, I removed the sentence, but left the signature as is. Currently, it doesn't matter what the `initialize_*` methods return.

docs/user/customization.rst
```diff
@@ -5,26 +5,22 @@ Customization
...
+and :class:`.NeuralNetRegressor` for typical classification, binary
+classification, and regressions tasks. They should work as drop-in
+replacements for sklearn classifiers and regressors.
 
-Initialization
-^^^^^^^^^^^^^^
+The :class:`.NeuralNet` class is a little less opinionated about the
+incoming data, e.g. it does not determine a loss function by default.
+Therefore, if you want to write your own subclass for a special use
+case, you would typically subclass from :class:`.NeuralNet`.
```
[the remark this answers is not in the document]
BenjaminBossan: I added a sentence.

docs/user/customization.rst
```diff
@@ -38,6 +34,30 @@ quite sure, consult their documentations. In general, these methods
...
+            loss = super().get_loss(y_pred, y_true, X=X, training=training)
+            loss += self.lambda1 * sum([w.abs().sum() for w in self.module_.parameters()])
+            return loss
+
+.. note:: This example also regularizes the biases, which you typically
+    don't need to do.
+
+It is often a good idea to call ``super`` of the method you override, to make
+sure that everything that needs to happen inside that method does happen. If you
+don't, you should make sure to take care of everything that needs to happen.
```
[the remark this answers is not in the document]
BenjaminBossan: Good point, I added this.

skorch/callbacks/training.py
```diff
@@ -532,7 +532,7 @@ def initialize(self):
         return self
 
     def named_parameters(self, net):
-        return net.module_.named_parameters()
+        return net.get_learnable_params()
```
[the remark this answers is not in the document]
BenjaminBossan: Unfortunately not, the names can be identical. I think this method is really just for the case that you don't need to discern. Otherwise, a user should use `self.module_.named_paramters()`, `self.mymodule_.named_parameters()` etc., as shown in the docs.

skorch/net.py
BenjaminBossan: I agree, this is a worthwhile tradeoff.

BenjaminBossan:
> I would opt for removing the optimizer_name argument. Maybe even rename the function to `get_all_learnable_params` to avoid confusion about its purpose?

GH doesn't let me reply directly, so here is the answer: Yes, I was thinking the same. Changed it accordingly.

> I am a little worried about the statefulness of `__setattr__` / `initialize_{optimizer,module,criterion}` but the risk seems negligible.

Well, these methods have always been stateful ;) but I know what you mean. Yes, it's important that we get the book keeping right, but I believe the current state is unfortunate, since it's so half-baked, and IMO this is a strict improvement (provided no bugs are introduced).

skorch/callbacks/training.py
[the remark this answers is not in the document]
githubnemo: Maybe we need to refactor this callback then, have it implement `named_parameters(self, net, module_name)` and retrieve the module from `net.modules_`. When specifying parameters we could then introduce `custommodule__0__foo=3` and have `custommodule` resolve the module object. But this does not need to happen right now. It is probably an uncommon combination anyway.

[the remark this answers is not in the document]
BenjaminBossan: Yes, good point. Shall I roll back the change here to preserve the current behavior?

skorch/tests/test_net.py
```diff
@@ -2620,33 +2746,597 @@ def initialize_module(self, *args, **kwargs):
...
+            else:
+                check = lambda module: module.training is False
+            assert check(net.module_)
+            assert check(net.mymodule_)
+            assert check(net.criterion_)
+            assert check(net.mycriterion_)
+
+        class CheckTrainingCallback(Callback):
+            def on_batch_end(self, net, batch, training, **kwargs):
+                assert_net_training_mode(net, training=training)
```
thomasjpfan, line 3258: I guess the easiest way to assert that `on_batch_end` is actually being called is to use the `nonlocal` trick from above.

skorch/tests/test_net.py
```diff
@@ -2813,7 +3503,7 @@ def forward(self, x0, x1):
         class MyNet(NeuralNet):
             """Override train_step_single and validation_step"""
             def train_step_single(self, batch, **fit_params):
-                self.module_.train()
+                self._set_training(True)
```
thomasjpfan, line 3506: `self.module_` should still work when this PR is merged?

Given that `_set_training` is private API, I would not expect third-party developers to know about it.

skorch/net.py
```diff
@@ -243,11 +236,33 @@ class NeuralNet:
...
+      this list.
+
     """
-    prefixes_ = ['module', 'iterator_train', 'iterator_valid', 'optimizer',
-                 'criterion', 'callbacks', 'dataset']
+    prefixes_ = ['iterator_train', 'iterator_valid', 'callbacks', 'dataset']
+
+    cuda_dependent_attributes_ = []
 
-    cuda_dependent_attributes_ = ['module_', 'optimizer_', 'criterion_']
```
thomasjpfan, line 250: This is not backward compatible but should be okay.

skorch/net.py
```diff
@@ -243,11 +236,33 @@ class NeuralNet:
...
+    # This attribute keeps track of which initialization method is being used.
+    # It should not be changed manually.
+    init_context_ = None
+
+    modules_ = []
+    criteria_ = []
+    optimizers_ = []
```
thomasjpfan, line 265: Typically, this would not work with grid search, but the way these attributes are updated with `__setattr__` allows it to work.

skorch/net.py
```diff
@@ -567,63 +591,243 @@ def _apply_virtual_params(self, virtual_kwargs):
...
+        # this init context is for consistency and not being used at the moment
+        with self._current_init_context('callbacks'):
+            if self.callbacks == "disable":
+                self.callbacks_ = []
+                return self
+            self.initialize_callbacks()
+            return self
+
+    def _initialize_criterion(self, reason=None):
+        # _initialize_criterion and _initialize_module share the same logic
```
thomasjpfan, line 663: There are ways to try to DRY this code, but I am okay to leave it as is.

skorch/net.py
```diff
@@ -700,6 +904,51 @@ def get_train_step_accumulator(self):
...
+        step_fn : callable or None
+          If None, just call ``optimizer.step()`` without arguments. Else, this
+          will be passed as the training step closure to the optimizer(s). Note
+          that this could lead to the function being called multiple times. If
+          more fine-grained control is desired instead, please override the
+          :meth:`.train_step` method.
+
+        """
+        for name in self.optimizers_:
+            optimizer = getattr(self, name + '_')
```
thomasjpfan, line 946: Using string mangling to get the optimizers or modules is an okay approach.

I am thinking of using `WeakValueDictionary` to hold a weak reference could be better. It could even keep track of itself to avoid the whole `_unregister_attribute:

```python
from weakref import WeakValueDictionary

class A:
    pass

a_obj = A()

my_dict = WeakValueDictionary() 
my_dict["hello"] = a_obj

# object is in dict
assert "hello" in my_dict

# Remove object
del a_obj

# object not in dict
assert "hello" not in my_dict
```

If we use this `WeakValueDictionary`, then `modules_` can be a `@property` returning `self.modules_weak_dict_.keys()`.

I do not have a strong opinion. (`WeakValueDictionary` is not a well know python object)

thomasjpfan (commented):
Using `__setattr__` to dynamically define `modules_` and friends is quite magical, but it works.

skorch/net.py
```diff
@@ -567,63 +591,243 @@ def _apply_virtual_params(self, virtual_kwargs):
...
-            print(msg)
+        # handle deprecated paramter
+        if triggered_directly is not None:
+            warnings.warn(
+                "The 'triggered_directly' argument to 'initialize_optimizer' is "
+                "deprecated, please don't use it anymore.", DeprecationWarning)
 
-        if 'lr' not in kwargs:
-            kwargs['lr'] = self.lr
+        named_parameters = self.get_all_learnable_params()
```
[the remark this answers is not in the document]
BenjaminBossan: Yes, that's the idea. As is, if a user adds `self.module2_`, it's never updated. I have the feeling that this is not expected and it could indeed be very hard to spot. If a user wanted to update `self.module2_`, they would need to override `initialize_module` and add the parameters to the existing optimizer or create a new one.

In contrast, let's assume that the user does _not_ want `self.module2_` to be updated by `self.optimizer_`. In that case, they would most likely have to touch `initialize_module` anyway. If they do, it's trivial to have `self.optimizer_` only update parameters of `self.module_`.

WDYT? Is the proposed behavior unexpected?

skorch/tests/test_net.py
BenjaminBossan: Hmm, we test more than just if the method is called. Therefore, I don't see how `nonlocal` would help to simplify the test here. Maybe it's just too late in the evening for me :)

Anyway, I could simplify the test by removing the callback completely and instead override `on_batch_end` of the net itself.

skorch/tests/test_net.py
BenjaminBossan: Right. So your point is that I should reverse the changes to this test, since it should not depend on using the private API? If so, I agree, I will reverse the changes.

skorch/net.py
BenjaminBossan: I agree it's not the most elegant of solutions, but I couldn't come up with anything better.

BenjaminBossan:
No idea why all the tests are failing/being canceled. The offending tests seem to be:

```
FAILED skorch/tests/callbacks/test_logging.py::TestWandb::test_fit_with_real_experiment
FAILED skorch/tests/callbacks/test_logging.py::TestProgressBar::test_pickle
ERROR skorch/tests/callbacks/test_logging.py::TestTensorBoard::test_writer_closed_automatically
```

Those seem to be unrelated to my latest changes, which were exclusively inside `test_net.py` (and the tests also seem to be unrelated to each other?). I can't reproduce the error locally. Therefore, I will wait for now and run the tests again later, in case this issue fixes itself. Meanwhile, I believe my latest changes are safe to be reviewed.

skorch/net.py
```diff
@@ -243,11 +236,33 @@ class NeuralNet:
...
 
+    modules_ : list of str
+      List of names of all modules that are torch modules. This list is
+      collected dynamically when the net is initialized. Typically, there is no
+      reason for a user to modify this list.
+
+    criteria_ : list of str
+      List of names of all criteria that are torch modules. This list is
+      collected dynamically when the net is initialized. Typically, there is no
+      reason for a user to modify this list.
```
thomasjpfan, line 247: Is there any reason for a developer to directly interact with `criteria_`? Currently, I think its auto populated when one does `self.module2_  = ...`.

thomasjpfan (commented):
There is an argument for making `modules_` private because it is more of an implementation detail for `NeuralNet`. Users and subclass developers do not need to interact with it.

If we keep it public, we need to be super clear that `modules_` is set automatically and should not be touched directly.

BenjaminBossan:
> There is an argument for making `modules_` private because it is more of an implementation detail for `NeuralNet`. Users and subclass developers do not need to interact with it.
> 
> If we keep it public, we need to be super clear that `modules_` is set automatically and should not be touched directly.

You are right, Thomas, there should rarely, if ever, be the necessity for a user to touch these attributes. I renamed `modules_`, `criteria_` and `optimizers_` to `_modules`, `_criteria` and `_optimizers` to mark them as private.

thomasjpfan (approved):
I think this is ready. LGTM

skorch/callbacks/training.py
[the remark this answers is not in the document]
ottonemo: No I think this is fine as is, adding a `FIXME` might be a good idea so that we know that this is an issue and where it comes from.

BenjaminBossan:
After personal communication with @ottonemo he gave his thumbs up, so I went ahead and merged.

PR #754, opened 2021-04-09
Implement loading of checkpoints after training

ottonemo:
In conjunction with early stopping (and particularly grid searches)
this feature can come in handy as you would certainly want to use the
best model state for scoring your model after training without worrying
about loading it from the checkpoint manually.

It would be possible to implement this in other places, such as the
early stopping callback. To me this seems to be a conflict of
separation, which is why I chose this place.


skorch/callbacks/training.py
```diff
@@ -157,6 +169,7 @@ def __init__(
         self.dirname = dirname
         self.event_name = event_name
         self.sink = sink
+        self.load_best = load_best
```
BenjaminBossan, line 172: Using `load_best=True` in conjunction with `monitor=None` wouldn't make sense. Do you think it's worth checking that combination?

skorch/tests/callbacks/test_training.py
```diff
@@ -398,6 +398,48 @@ def initialize_module(self, *args, **kwargs):
             [call(f_mymodule='mymodule.pt')] * len(net.history)
         )
 
+    @pytest.fixture
+    def load_params(self):
+        import torch
+        return torch.load
+
+    @pytest.mark.parametrize('load_best_flag, should_match', [
+        (False, False),
```
BenjaminBossan, line 407: Make those one parameter, since they are always the same?

skorch/tests/callbacks/test_training.py
```diff
@@ -398,6 +398,48 @@ def initialize_module(self, *args, **kwargs):
...
+        # when restoring at the end of training, the parameters
+        # of the net should not differ. if we do not restore
+        # then the parameters must differ.
+        path_cb = tmp_path / 'params_cb.pt'
+        path_net = tmp_path / 'params_net.pt'
+
+        def save_once_monitor(net):
+            return len(net.history) == 1
+
+        net = net_cls(callbacks=[
```
BenjaminBossan, line 424: Maybe reduce number of epochs to 3 or so, to make the test run faster?

skorch/tests/callbacks/test_training.py
```diff
@@ -398,6 +398,48 @@ def initialize_module(self, *args, **kwargs):
...
+        (False, False),
+        (True, True),
+    ])
+    def test_automatically_load_checkpoint(
+            self, net_cls, checkpoint_cls, data, tmp_path,
+            load_params, load_best_flag, should_match,
+    ):
+        # checkpoint once at the beginning of training.
+        # when restoring at the end of training, the parameters
+        # of the net should not differ. if we do not restore
```
BenjaminBossan, line 416:
```suggestion
        # of the net should not differ. If we do not restore
```

BenjaminBossan (approved):
Nice little addition, we should have had this already. I just have a few minor suggestions. And could you please add an entry to `CHANGES`?

skorch/callbacks/training.py
ottonemo: Since it does not impact the user negatively I will just add documentation and prevent the loading from occurring when `self.monitor is None` to save IO/computation.

BenjaminBossan (approved):
Thanks, looks good.

PR #755, opened 2021-04-16
PyTorch 181 and some minor updates

BenjaminBossan:
### Main change

Add PyTorch 1.8.1 to CI, docs. Remove PyTorch 1.4.

### Further changes

- bump to cuda_toolkit 11.1
- add Python 3.9
- remove mention of Python 3.5
- add more common file types to `.gitignore`
- no longer mention nolearn in doc intro (honestly, at this point the reference will be pretty obscure to most people who start with skorch)
- update `environment.yml` (it was very old at this point but I guess rarely used)
- add dirty worktree check to `deploy.sh`
- add `README.rst` rendering check to `deploy.sh` (this way, if the `README.rst` cannot be rendered on PyPI, we get a warning and possibly error message before uploading to PyPI)
- add `flake8` section to setup.cfg (for those wo are using `flake8` (and possibly `flake8-docstrings`))

**Note**: I added `*.ipynb` to the `.gitignore`. This is because I have dozens of untracked notebooks lying around. But if this is not desired (as it requires `git add -f` for notebooks to be tracked), I can reverse that.

**Edit**: I had to remove PyTorch < 1.8 in combination with Python 3.9 from the matrix, there seems to be a segfaulting issue (possibly PyTorch issues 50014, 38228, not linking them to avoid spamming them).


environment.yml
```diff
@@ -2,38 +2,158 @@ name: skorch
...
   - pytorch
-  - conda-forge
 dependencies:
   - _libgcc_mutex=0.1
+  - absl-py=0.12.0
+  - aiohttp=3.7.4
+  - alabaster=0.7.12
+  - astroid=2.5
+  - async-timeout=3.0.1
+  - attrs=20.3.0
```
thomasjpfan, line 12: For `environment.yml`, I have been starting to prefer the following for dev environments:

```bash
conda env export -n skorch --from-history
```

docs/index.rst
```diff
@@ -9,9 +9,7 @@ Introduction
 
 The goal of skorch is to make it possible to use PyTorch_ with
 sklearn_. This is achieved by providing a wrapper around
-PyTorch that has an sklearn interface. In that sense,
-skorch is the spiritual successor to nolearn_, but instead of
-using Lasagne and Theano, it uses PyTorch.
```
thomasjpfan, line 14: `Lasagne` were good times :)

environment.yml
BenjaminBossan: I didn't know about that option, I will try it out and update the environment.yml in case it works, thanks!

docs/index.rst
BenjaminBossan: Yes, Theano and Lasagne were great, but history will only remember keras. Maybe we should sit down one day and write a history for posterity :D

environment.yml
BenjaminBossan: Okay, it worked. It's still quite cluttered but a bit better and I didn't need to clean up as much. What I did need to do: Add non default channels, add the `pip:` section, which is somehow omitted. Good news is that even including the dev deps, only sacred needs to be installed via pip.

thomasjpfan (commented):
I think we need a `python_requires` in `setup.py` to restrict installs of future releases to `>=3.6`.

BenjaminBossan:
Done

setup.py
```diff
@@ -27,7 +30,7 @@
     README = ''
 
 try:
-    CHANGES = open(os.path.join(here, 'CHANGES.txt')).read()
+    CHANGES = open(os.path.join(here, 'CHANGES.md')).read()
```
thomasjpfan, line 33: This can be removed since we do not use `CHANGES` anywhere?

thomasjpfan (approved):
Otherwise LGTM

PR #763, opened 2021-04-28
Docs: Fix class path in predefined_split example

matthiaspfenninger:
It fixes the class path in the example of predefined_split, using `skorch.dataset.Dataset` instead of `skorch.Dataset`

**Reference issues/PRs**
None


BenjaminBossan:
Thanks for the PR, Matthias, good catch.

Issue #769, opened 2021-05-03
Add MLflow builtin callback


cacharle:
MLflow website: <https://mlflow.org/>

I'll be happy to implement it myself if you want.

ottonemo:
Hi! Sounds great :)

There are a few loggers implemented (https://github.com/skorch-dev/skorch/blob/master/skorch/callbacks/logging.py#L733) and tested (https://github.com/skorch-dev/skorch/blob/master/skorch/tests/callbacks/test_logging.py#L19) already, I think you can base your implementation on those if you want.

PR #770, opened 2021-05-05
Added MLflow callback

cacharle:
Here is a small draft for #769, I feel like the different loggers have a lot in common and could be refactored in a common class.


skorch/callbacks/logging.py
```diff
@@ -875,3 +875,70 @@ def on_epoch_end(self, net, **kwargs):
...
+    def __init__(
+        self,
+        log_on_batch_end=False,
+        log_on_epoch_end=True,
+        keys_ignored=[],
+        create_artifact=True,
+        logged_parameters=["lr", "batch_size"],
+    ):
+        import mlflow
+        self.mlflow = mlflow
```
cacharle, line 896: Is this the correct way to get the library without raising an exception if the user doesn't have a use for this callback?

skorch/callbacks/logging.py
```diff
@@ -875,3 +875,70 @@ def on_epoch_end(self, net, **kwargs):
...
+    >>> with mlflow.start_run():
+    ...     net.fit(X, y)
+    """
+    def __init__(
+        self,
+        log_on_batch_end=False,
+        log_on_epoch_end=True,
+        keys_ignored=[],
+        create_artifact=True,
+        logged_parameters=["lr", "batch_size"],
```
cacharle, line 893: The other loggers let the user register the parameters manualy but I thought it could have it's place here

skorch/callbacks/logging.py
```diff
@@ -875,3 +875,70 @@ def on_epoch_end(self, net, **kwargs):
...
+            self.mlflow.log_metric("epoch_" + key, epoch_logs[key])
+
+    def on_train_end(self, net, **kwargs):
+        if not self.create_artifact:
+            return
+        params_filename = "params.pkl"
+        optimizer_filename = "optimizer.pkl"
+        criterion_filename = "criterion.pkl"
+        history_filename = "history.json"
+        net.save_params(
```
cacharle, line 935: Maybe create temporary files to store the model to avoid cluttering the user current directory.

BenjaminBossan:
Thank you for working on this. I have a few change suggestions how to possibly change this. Please tell me what you think.

As you probably saw, this implementation is a bit different than those for tensorboard, neptune, sacred, etc. They all use an object that encompasses something like an "experiment" or "session", which is passed to the logger and is used to run all the methods on. Here, instead we have instead the local import of `mlflow`. If possible, I would like to see the same design pattern used here as for the other loggers.

Without any knowledge of mlflow, I took a dive through the code and it seems like under the hood, `mlflow.log_metric` is indeed just a shortcut to retrieve an `MlflowClient`, on which the `log_metric` method is called. So I wonder if we can make use of that, similar to how it's shown in the mlflow [docs here](https://mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.create_experiment). Since you know mlflow much better, do you believe this is feasible?

And before doing any in depth review and testing out mlflow, could you please rename the callback to `MlflowLogger` or something similar, and also add tests? Again, take inspiration from (and copy) code from the other loggers.

skorch/callbacks/logging.py
BenjaminBossan: See my [comment](https://github.com/skorch-dev/skorch/pull/770#issuecomment-833780968)

skorch/callbacks/logging.py
BenjaminBossan: Could you elaborate?

BenjaminBossan:
@cacharle Any updates, do you need support?

cacharle:
Hi, found a way to use a `Run` and `MlflowClient` instance instead of calling the the basic mlflow API.
I'll try to add the tests by this weekend.

BenjaminBossan:
Please tell me once the PR is ready for review.

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,101 @@ def on_epoch_end(self, net, **kwargs):
...
+        self.epoch_suffix = epoch_suffix
+
+    def initialize(self):
+        if self.run is None:
+            import mlflow
+            self.run = mlflow.active_run()
+        self.run_id = self.run.info.run_id
+        if self.client is None:
+            from mlflow.tracking import MlflowClient
+            self.client = MlflowClient()
```
cacharle, line 920: I added the support for run and client object as you ask
but by default I get the currently active experiment and a default client,
I think it's a good compromise between being customisable and being user friendly.

We can do:

```python
with mlflow.start.run():
    net = NeuralNetClassifier(..., callbacks=[MlflowLogger])
    net.fit(...)
```

instead of:

```python
client = MlflowClient
experiment = client.get_experiment_by_name(name)
run = client.create_run(experiment)
net = NeuralNetClassifier(..., callbacks=[MlflowLogger(run, client)])
net.fit(...)
```

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,103 @@ def on_epoch_end(self, net, **kwargs):
...
+        self,
+        run=None,
+        client=None,
+        create_artifact=True,
+        terminate_after_train=True,
+        log_on_batch_end=False,
+        log_on_epoch_end=True,
+        keys_ignored=None,
+        batch_suffix=None,
+        epoch_suffix=None,
```
BenjaminBossan, line 901: Would it not be easier to set the default to `'_batch'` and `'_epoch'`, respectively? Then you don't need to do the checking against `None` down below.

skorch/callbacks/logging.py
BenjaminBossan: Yes, I guess this is okay, since the user has the choice to create the experiment and run explicitly. Could you do a minor change though and call the variables `client_` and `run_`? This is with keeping the convention to not touch the `__init__` attributes.

And if the user passes a client and run that are not `None`, still do `self.run_ = self.run` and `self.client_ = self.client`.

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,103 @@ def on_epoch_end(self, net, **kwargs):
...
         for key in filter_log_keys(epoch_logs.keys(), self.keys_ignored_):
             self.experiment.log_scalar(key + self.epoch_suffix_, epoch_logs[key], epoch)
+
+
+class MlflowLogger(Callback):
+    """
+    >>> import mlflow
+    >>> net = NeuralNetClassifier(net, callbacks=[MLflowLogger])
+    >>> with mlflow.start_run():
+    ...     net.fit(X, y)
```
BenjaminBossan, line 886: A few things should be addressed in the docstring:

1. add a bit of description for those users who are not familiar with MLFlow.
2. give an install instruction, `$ pip install mlflow`, since it's not part of the skorch dependencies
3. use numpydoc style.
4. document all parameters
5. since there are 2 ways of initializing, it would be nice to show examples of both and explain why you would use one or the other

For most of these points, just look at the other loggers and copy the style used there.

skorch/callbacks/__init__.py
```diff
@@ -37,4 +37,5 @@
     'Unfreezer',
     'WandbLogger',
     'WarmRestartLR',
+    'MlflowLogger',
```
BenjaminBossan, line 40: Could you please put this in alphabetical order?

BenjaminBossan (changes requested):
Thanks a lot for addressing my initial comments. I tested the MLFlow logger out and it worked well. So this is a very welcome addition.

There are a few things that I would like you to address before merging, please take a look at the comments. On top of that, could you please add an entry to the CHANGES.md? Thanks.

skorch/callbacks/logging.py
cacharle: This is to avoid adding a suffix if only one of batch or epoch is logged.

Same as the [SacredLogger](https://github.com/skorch-dev/skorch/blob/master/skorch/callbacks/logging.py#L829).

cacharle:
Should I squash all the commit I made before merging the PR to avoid cluttering the logs?

BenjaminBossan:
> Should I squash all the commit I made before merging the PR to avoid cluttering the logs?

Please don't, it makes reviewing harder. I will squash before merging :)

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,183 @@ def on_epoch_end(self, net, **kwargs):
...
+    ----------
+
+    run : mlflow.entities.Run (default=None)
+      Instantiated ``Run`` class.
+
+    client : mlflow.tracking.MlflowClient (default=None)
+      Instantiated ``MlflowClient`` class.
+
+    create_artifact : bool (default=True)
+      Weather to create artifacts for the network's
```
BenjaminBossan, line 935:
```suggestion
      Whether to create artifacts for the network's
```

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,183 @@ def on_epoch_end(self, net, **kwargs):
...
+        epoch_logs = net.history[-1]
+        self._iteration_log(epoch_logs, self.epoch_suffix_, len(net.history))
+
+    def _iteration_log(self, logs, suffix, step):
+        for key in filter_log_keys(logs.keys(), self.keys_ignored_):
+            self.client_.log_metric(
+                self.run_.info.run_id,
+                key + suffix,
+                logs[key],
+                step=step
```
BenjaminBossan, line 1030:
```suggestion
                step=step,
```

skorch/tests/callbacks/test_logging.py
```diff
@@ -941,3 +942,185 @@ def forward(self, b, e, c, d, a, **kwargs):
...
+        assert logger.batch_suffix_ == batch_suffix
+        assert logger.epoch_suffix_ == epoch_suffix
+
+    @pytest.mark.parametrize('batch_suffix', ['', '_foo'])
+    @pytest.mark.parametrize('epoch_suffix', ['', '_bar'])
+    def test_epoch_batch_custom_suffix(
+            self,
+            logger_mock_cls,
+            batch_suffix,
+            epoch_suffix
```
BenjaminBossan, line 1056:
```suggestion
            epoch_suffix,
```

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,183 @@ def on_epoch_end(self, net, **kwargs):
...
+    >>> experiment = client.get_experiment_by_name('Default')
+    >>> run = client.create_run(experiment.experiment_id)
+    >>> net = NeuralNetClassifier(..., callbacks=[MlflowLogger(run, client)])
+    >>> net.fit(X, y)
+
+    Parameters
+    ----------
+
+    run : mlflow.entities.Run (default=None)
+      Instantiated ``Run`` class.
```
BenjaminBossan, line 929: Could you please add a description about what happens if the value is None?

skorch/callbacks/logging.py
```diff
@@ -875,3 +876,183 @@ def on_epoch_end(self, net, **kwargs):
...
+
+    client : mlflow.tracking.MlflowClient (default=None)
+      Instantiated ``MlflowClient`` class.
```
BenjaminBossan, line 932: Could you please add a description about what happens if the value is None?

BenjaminBossan (changes requested):
Thanks for addressing the comments. The MLFlow logger is almost good to go, there are only a few minor comments left and a minor merge conflict to solve.

skorch/callbacks/logging.py
cacharle: lol

skorch/callbacks/logging.py
cacharle: I haven't look a lot into it but it seems that flake8 extensions exist to enforce that: https://pythonrepo.com/repo/PyCQA-flake8-commas-python-linters-style-checkers

BenjaminBossan: Yes, that would be possible. We currently don't enforce any linting (though we have set up pylint) because it would require quite a lot of clean up. Certainly something to think about in the future (probably replacing pylint with flake8).

BenjaminBossan:
Thanks for implementing these changes and going through the extra effort of linking to the MLFlow docs. Now there is only the merge conflict left to solve :)

BenjaminBossan (approved):
Thanks, great work!

PR #772, opened 2021-05-12
This is just a test to check the CI

BenjaminBossan:
Alright, after a bunch of testing, it turns out that wandb was the culprit. I currently only allow <0.10.29, but hopefully at some time in the future, whatever the issue is will be fixed.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


thomasjpfan (approved):
Is [this](https://github.com/skorch-dev/skorch/runs/2574438128) the failing test?

BenjaminBossan:
> Is [this](https://github.com/skorch-dev/skorch/runs/2574438128) the failing test?

Yes. Interestingly, two seemingly unrelated tests failed (tensorboard + wandb), which initially led me down the wrong path, before narrowing down the error on wandb.

PR #781, opened 2021-06-13
Fix a bug that led to double-registration

BenjaminBossan:
This is a bug introduced by a last minute change in #751 

After `clone`ing a net, `_module`, `_criteria`, and `_optimizers` are already
populated. Then, when loading params, there is yet another registration,
i.e. a double registration. As a consequence, there would be two
`'modules'` etc. This is a fix for that.

The solution is a two-pronged: Don't return `_modules` etc. when cloning,
and make a safety check during registration. Either of those should be
sufficient, but it's probably better to be extra safe.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1792,7 +1792,10 @@ class NeuralNet:
         # special treatment.
         params_cb = self._get_params_callbacks(deep=deep)
         params.update(params_cb)
-        return params
+
+        # don't include the following attributes
+        to_exclude = {'_modules', '_criteria', '_optimizers'}
+        return {key: val for key, val in params.items() if key not in to_exclude}
 
     def _check_kwargs(self, kwargs):
         """Check argument names passed at initialization.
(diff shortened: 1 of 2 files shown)


skorch/net.py
```diff
@@ -2095,11 +2098,13 @@ def _register_attribute(
             self.cuda_dependent_attributes_ = (
                 self.cuda_dependent_attributes_[:] + [name + '_'])
 
-        if self.init_context_ == 'module':
+        # make sure to not double register -- this should never happen, but
+        # still better to check
+        if (self.init_context_ == 'module') and (name not in self._modules):
             self._modules = self._modules[:] + [name]
```
thomasjpfan, line 2104: I think this bound is what fixed the bug. Would a `set` be a better data structure for `_modules`?

skorch/net.py
```diff
@@ -1792,7 +1792,10 @@ def get_params(self, deep=True, **kwargs):
         # special treatment.
         params_cb = self._get_params_callbacks(deep=deep)
         params.update(params_cb)
-        return params
+
+        # don't include the following attributes
+        to_exclude = {'_modules', '_criteria', '_optimizers'}
+        return {key: val for key, val in params.items() if key not in to_exclude}
```
thomasjpfan, line 1798: On `master`, these attributes are not included in `get_params`.

thomasjpfan (commented):
Thank you for the fix!

skorch/net.py
BenjaminBossan: What do you mean? On current master, after merging #751, `get_params` includes those.

thomasjpfan: Ah I was mistaken. (I was on a older version of master locally)

Can we add a test here:

https://github.com/skorch-dev/skorch/blob/812f54d76cf6c7a1aba213dc2290a47607c6f55c/skorch/tests/test_net.py#L1409

Something like:

```python
assert `_modules` not in params
```

BenjaminBossan: Good idea. I added a separate test below because that one was already big enough.

thomasjpfan (approved):

PR #782, opened 2021-06-13
Gaussian processes via gpytorch

BenjaminBossan:
This is a feature to add support for Gaussian Processes (GPs) via integration with [gpytorch](https://github.com/cornellius-gp/gpytorch).

Similar to "vanilla" PyTorch, the idea here is that skorch allows the user to focus on what's important (implementing the mean function and kernel function) and not to bother with stuff like the training loop, callbacks, etc. This is probably best illustrated in the accompanying [notebook](https://github.com/skorch-dev/skorch/blob/feature/gaussian-processes-via-gpytorch/notebooks/Gaussian_Processes.ipynb).

GPs are primarily for regression, hence those are the main focus here. Traditionally, there are "exact" solutions and approximations. skorch will provide an `ExactGPRegressor` and a `GPRegressor` for those two use cases.

On top of that, a `GPBinaryClassifier` is offered, though I suspect it to be rarely used. I couldn't get the `GPClassifier` for multiclass to work, the code is therefore commented out.

The API is mostly the same as for the normal skorch estimators. There are some additions to make working with GPs easier:

- `predict` method takes a `return_std` argument to return the standard deviation as well (as in sklearn's [`GaussianProcessRegressor`](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html#sklearn.gaussian_process.GaussianProcessRegressor); `return_cov` is not supported)
- `sample` method to sample for the model
- `confidence_region` method to get the confidence region

Before merging this, I would like this to be reviewed by some experts on GPs, since I consider myself to know very little about them. Therefore, I will try to find some external help. Especially the following points should be clarified:

- [ ] Do the gyptorch criteria always takes likelihood and module as input arguments?
- [ ] Do gpytorch losses always need to be maximized, i.e. we should minimize the negative objective function?
- [ ] Does sampling work even if we use batching (i.e. for non-exact solutions)?
- [ ] What is the problem with the multiclass classifier? Is there a self-contained example to replicate?

This PR already includes the bugfix in #781


BenjaminBossan:
@ibayer This could be interesting for you. If you have time, maybe you could look at this PR and tell me if there is something terribly wrong :) Probably the best way is to take a look at the provided [notebook](https://github.com/skorch-dev/skorch/blob/feature/gaussian-processes-via-gpytorch/notebooks/Gaussian_Processes.ipynb) to see the integration in action.

ibayer:
@BenjaminBossan Thanks for the ping. I had a quick view ad the nb, I like it. :-) I'll be back with detailed feedback (from a user perspective) as soon as time permits (hopefully sometime next week).

BenjaminBossan:
@ibayer Thanks, looking forward to getting your opinion.

General comment: As to some CI matrix runs failing, this seems to be related to a change in a newer version of GPyTorch. I'm working on it but so far it seems unlikely that it will affect the usage of the feature in any way.

BenjaminBossan:
Note: GPyTorch + PyTorch 1.7.1 don't work together, so I skip GPyTorch tests if this PyTorch version is detected. Officially, PyTorch >= 1.8.1 is supported by GPyTorch, but PyTorch 1.5.1 and 1.6.0 also work, thus I left those tests in.

docs/user/probabilistic.rst
```diff
@@ -0,0 +1,196 @@
...
+            covar_x = self.covar_module(x)
+            return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)
+
+    gpr = ExactGPRegressor(
+        RbfModule,
+        module__X=X,
+        module__y=y,
+    )
+
+    gpr.fit(X, y)
```
thomasjpfan, line 68: Passing `X`, and `y` into `ExactGPRegressor.__init__` and in `fit` is a little strange. Alternatives where we delay initializing the module until `fit` is call feels too hacky.

docs/user/probabilistic.rst
```diff
@@ -0,0 +1,196 @@
...
+
+        def forward(self, x):
+            mean_x = self.mean_module(x)
+            covar_x = self.covar_module(x)
+            return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)
+
+    gpr = ExactGPRegressor(
+        RbfModule,
+        module__X=X,
+        module__y=y,
```
thomasjpfan, line 65: I see that there is a default `likelihood` that I assume routes into `__init__`. I think at this point we can use `train_inputs`, `train_targets` and route into `ExactGPRegressor`. This would be more inline with `ExactGP` itself.

thomasjpfan (commented):
Initial look at the docs/API.

docs/user/probabilistic.rst
BenjaminBossan: Do you mean that `X` and `y` should be renamed to `train_inputs` and `train_targets`?

docs/user/probabilistic.rst
BenjaminBossan: I had started out in a similar direction but it really didn't work well. The current solution is not the most beautiful but it serves its purpose and uses facilities that are already present. But I'm open for suggestions.

skorch/tests/test_probabilistic.py
```diff
@@ -0,0 +1,697 @@
...
+
+        X = data[0]
+        y_proba = gp_fit.predict_proba(X)
+
+        assert isinstance(y_proba, np.ndarray)
+        assert y_proba.shape == (self.n_samples, self.n_targets)
+
+    def test_fit_and_predict_with_cuda(self, gp, data):
+        if not torch.cuda.is_available():
+            pytest.skip()
```
thomasjpfan, line 301:
```suggestion
    @pytest.mark.skipif(not torch.cuda.is_available(), reason="no cuda device")
    def test_fit_and_predict_with_cuda(self, gp, data):
```

skorch/utils.py
```diff
@@ -604,6 +622,15 @@ def _infer_predict_nonlinearity(net):
     if isinstance(criterion, BCEWithLogitsLoss):
         return _sigmoid_then_2d
 
+    likelihood = getattr(net, 'likelihood_', None)
+    if (
+            likelihood
+            and GPYTORCH_INSTALLED
+            and isinstance(likelihood, gpytorch.likelihoods.SoftmaxLikelihood)
```
thomasjpfan, line 629: The code using `SoftmaxLikelihood` is commented out, so this code path is never reached?

skorch/probabilistic.py
```diff
@@ -0,0 +1,895 @@
...
+
+        Returns
+        -------
+        loss : torch Tensor (scalar)
+          The loss to be minimized.
+
+        """
+        loss = super().get_loss(y_pred, y_true, X=X, training=training)
+        if loss.dim() != 0:
+            loss = loss.mean()
```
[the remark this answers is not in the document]
BenjaminBossan: The gpytorch losses don't always return a scalar by default, as e.g. in [this example](https://docs.gpytorch.ai/en/v1.2.1/examples/01_Exact_GPs/Spectral_Delta_GP_Regression.html#Train), so I added this conditional to take care of this. Ideally, it wouldn't be necessary.

docs/user/probabilistic.rst
BenjaminBossan: > As long as we check that `module__X` is `X` in `fit`, I am okay with this workaround.

GPyTorch already checks for this equality (unless debug mode is explicitly switched off). I believe this is enough and skorch doesn't need to do the same work.

> BTW Another hack would be to do something like `nn.LazyLinear`: https://github.com/pytorch/pytorch/blob/e711b5ce6c4982960c293b005da37a6d91253bd7/torch/nn/modules/lazy.py#L52 (I do not recommend it)

I agree that we shouldn't use it (for now), given that it's still experimental. Maybe GPyTorch will adopt it so that we don't have to.

docs/user/probabilistic.rst
BenjaminBossan: :+1:

skorch/tests/test_probabilistic.py
```diff
@@ -0,0 +1,697 @@
...
+            gpytorch.kernels.RBFKernel(eps=eps))
+
+    def forward(self, x):
+        mean_x = self.mean_module(x)
+        covar_x = self.covar_module(x)
+        latent_pred = gpytorch.distributions.MultivariateNormal(mean_x, covar_x)
+        return latent_pred
+
+
+class BaseProbabilisticTests:
```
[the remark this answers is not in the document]
BenjaminBossan: I was doing an experiment here. Do you think this overall improves the code or not? It does reduce code duplication but it may be less obvious what is going on.

skorch/utils.py
BenjaminBossan: Indeed, good catch, I commented this code out (adding a comment), as well as the code for `_transpose`.

thomasjpfan:
Ran through the notebook and left some comments at https://app.reviewnb.com/skorch-dev/skorch/blob/feature/gaussian-processes-via-gpytorch/notebooks/Gaussian_Processes.ipynb/ Let me know if you can see them.

skorch/tests/test_probabilistic.py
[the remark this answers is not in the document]
thomasjpfan: It's mostly okay with me. Looking at this more closely, it all sense and its a reasonable way to "do not repeat yourself". I can see other approaches, but I do not see it as drastic improvement over what we have now.

BenjaminBossan:
@thomasjpfan Thanks for your comments, I could read them. Hopefully, you can read my replies as well. I adjusted the notebook according to your comments/wrote a reply.

skorch/probabilistic.py
```diff
@@ -0,0 +1,941 @@
...
+        except pickle.PicklingError as exc:
+            msg = ("This GPyTorch model cannot be pickled. The reason is probably this:"
+                   " https://github.com/pytorch/pytorch/issues/38137. "
+                   "Try using 'dill' instead of 'pickle'.")
+            raise pickle.PicklingError(msg) from exc
+
+
+class _GPRegressorPredictMixin:
+    """Mixin class that provides a predict method for GP regressors."""
+    def predict(self, X, return_std=False, return_cov=False):
```
jacobrgardner, line 342: When `return_std is False`, consider turning the `gpytorch.settings.skip_posterior_variances` setting on -- this will avoid doing the math for the posterior variances altogether, which will save a great deal of compute.

skorch/probabilistic.py
```diff
@@ -0,0 +1,941 @@
...
+          batches, they must be concatenated.
+
+        Returns
+        -------
+        samples : torch.Tensor
+          Samples from the posterior distribution.
+
+        """
+        self.check_is_fitted()
+        samples = [p.sample(torch.Size([n_samples])) for p in self.forward_iter(X)]
```
jacobrgardner, line 295: Obviously not 100% familiar with how skorch's internals work, but assuming `p` is a `MultivariateNormal`, is there any reason to call `sample` instead of `rsample` here? Outright enforcing that sampling must be done in a `torch.no_grad()` context seems unnecessary at first glance, and a common use case for GPs looks something like: `x -> GP -> p(y|D) -> samples of y -> average some function of y over the samples -> differentiate the average with respect to x`

skorch/probabilistic.py
```diff
@@ -0,0 +1,941 @@
...
+        """
+        if return_cov:
+            msg = ("The 'return_cov' argument is not supported. Please try: "
+                   "'posterior = next(gpr.forward_iter(X)); "
+                   "posterior.covariance_matrix'.")
+            raise NotImplementedError(msg)
+
+        nonlin = self._get_predict_nonlinearity()
+        y_preds, y_stds = [], []
+        for yi in self.forward_iter(X, training=False):
```
jacobrgardner, line 376: One thing that GPyTorch does is we cache a couple things that don't depend on the test points the first time we make predictions so that the next time we make predictions we don't have O(n^3) complexity. These caches get deleted in some cases, like if the model gets put back in `training` mode -- I assume that `forward_iter` wouldn't do any of these things though.

It should be pretty easy to test:
```python
gpr = ...
gpr.fit(...)
with gpytorch.settings.fast_pred_var():
    y_pred, y_std = gpr.predict(test_x, return_std=True)
    # test_x could change here.
    y_pred, y_std = gpr.predict(test_x, return_std=True)   # This call should be significantly faster.

File: docs/user/probabilistic.rst [hunk 10]
```diff
@@ -0,0 +1,193 @@
...
+    num_training_samples = len(X_train)
+
+    gpr = GPRegressor(
+        VariationalModule,
+        module__inducing_points=X_inducing,
+        criterion__num_data=num_training_samples,
+    )
+
+    gpr.fit(X_train, y_train)
+    y_pred = gpr.predict(X_train)
```
jacobrgardner, line 171: The jupyter notebook you sent me currently highlights `gpytorch.settings.fast_pred_samples`. when calling `sample`. That setting won't actually do anything unless you are using KISS-GP. A setting that definitely _will_ make a perf difference is wrapping predict in `gpytorch.settings.fast_pred_var()` though, assuming `gpytorch.settings.skip_posterior_variances()` isn't also on (see my comment about that below).

jacobrgardner (commented):
Mostly looks good to me! Left a few comments, primarily about the prediction and sampling methods.

ibayer:
Since jacobrgardner already review as authority on GPyTorch, I looked at the integration from the perspective of potential user. This is what I came up with.

When I was reading the tutorial I was wondering in which situation I should use skorch's GPyTorch integration
and when not. I got the following impression which (if correct) might be useful for other first time users as well.

Good use cases:

- Scaling GPs with simple standard kernels and data that leads to too slow computation or memory requirements for libraries such as sklearn. 
- Deploying a GP with complex kernel so that contributors not familiar with the nitty gritty details of GPs and GPyTorch features.

Better use GPyTorch directly: 

- Building complex kernel compositions and exploring their data fits in detail.

This would then naturally lead to section titles in the tutorial nb such as "Scaling GP's" and "Deploying GPyTorch Models with Skorch". The material is already there but this would have helped me to jump right into the parts most relevant for me.

Below are a few nitpicking comments on the tutorial text.

> If you need to use batches (say, you don't have enough GPU memory to fit all your data), you have to use variational GPs, as shown later in the notebook.

Sparse GP's can be used to scale to larger data. Not all approaches a based on variational inference.

- you have to -> you can

> even though we have not even trained the model.

This is a bit tricky since GPs are non-parametric methods and "learning" is usually associated with estimating primary parameters and not hyper-parameters as with GPs.

- trained the model -> have not fitted the hyper-parameters

> Since the model learns a real distribution, not just a point prediction as is most often the case for Deep Learning

Again the learning thing ;), I would say the distribution is modeled using kernel engineering and the distribution parameters are fitted.

- Since the model learns a real distribution -> Since the model represents a probability distribution.

That's all, I think the PR is in good shape. Thanks for making GP's easier accessible. :)

skorch/probabilistic.py
BenjaminBossan: Good point, I will add this.

skorch/probabilistic.py
BenjaminBossan: The main reason I used `sample` was because `rsample` was not supported for all distributions, e.g. `Bernoulli`. I didn't think of the possibility of using the sample gradients. I'll make a switch to `rsample` if the used distribution supports it and fall back on `sample` otherwise.

skorch/probabilistic.py
BenjaminBossan: Indeed, when using `forward_iter`, there is no risk of the model being set to `training` again. Therefore, I'll add the `fast_pred_var` context for it. When testing, I didn't find any noticeable speed difference. Perhaps, that is because I'm mostly using toy models on small data in the notebook?

docs/user/probabilistic.rst
BenjaminBossan: I didn't know that, thanks for clarifying. I will remove the usage of `gpytorch.settings.fast_pred_var` in the notebook.

BenjaminBossan:
Thank you @ibayer for your feedback.

> This would then naturally lead to section titles in the tutorial nb such as "Scaling GP's" and "Deploying GPyTorch Models with Skorch". The material is already there but this would have helped me to jump right into the parts most relevant for me.

I couldn't quite manage to rework the sections to fit the titles as you suggested them. Still, as you said, it's important to clarify when and when not to use GPyTorch with skorch. To this end, I added a section at the start of the GP documentation in skorch that discusses this topic. The notebook references the docs too, so that interested readers can look up that information.

> > If you need to use batches (say, you don't have enough GPU memory to fit all your data), you have to use variational GPs, as shown later in the notebook.
> 
> Sparse GP's can be used to scale to larger data. Not all approaches a based on variational inference.

The part you cite talks about using batches and not about scaling per se. For batching, use of sparse GPs would not help and the statement is correct as is ;) But I get what you mean, I'll change the wording to no longer give the impression that variational GPs are the _only_ possibility for scaling.

> This is a bit tricky since GPs are non-parametric methods and "learning" is usually associated with estimating primary parameters and not hyper-parameters as with GPs.

> Again the learning thing ;), I would say the distribution is modeled using kernel engineering and the distribution parameters are fitted.

Thank you, you are right. I'll change the wording to not mislead the reader.

BenjaminBossan:
@thomasjpfan @ottonemo @jacobrgardner @ibayer 

Thank you all for your helpful reviews. Hopefully, I could address all of your concerns.

My proposal is to give one more week for any issues with this PR to be raised. If not, I'll assume it is ready to go now. Then I'd merge and prepare the next skorch release. For that, I would frame this feature as a "beta" with possible breaking changes in the future in case something is just plain wrong. That way, we are not shackled by possible mistakes I made here.

Please let me know if you have any objections.

PR #783, opened 2021-06-20
Fix bug that could lead to duplicate params

BenjaminBossan:
When a net's module references another of that net's modules, the former
will yield the latter's parameters. Therefore, when all parameters are
collected, the latter's parameters appear twice.

This bugfix consists of remembering all yielded parameters in a set and
not yielding those that were already encountered.

This bug was introduced very recently (#751) and should occur very
rarely, since modules don't typically reference each other.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -758,11 +758,24 @@ class NeuralNet:
           pass the named parameters to :meth:`.get_params_for_optimizer`.
 
         """
+        # Note: we have to filter out potential duplicate parameters. This can
+        # happen when a module references another module (e.g. the criterion
+        # references the module), thus yielding that module's parameters again.
+        # The parameter name can be difference, therefore we check only the
+        # identity of the parameter itself.
+        seen = set()
         for name in self._modules + self._criteria:
             module = getattr(self, name + '_')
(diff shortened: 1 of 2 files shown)


skorch/net.py
```diff
@@ -758,11 +758,24 @@ def initialize_optimizer(self, *args, **kwargs):
           pass the named parameters to :meth:`.get_params_for_optimizer`.
 
         """
+        # Note: we have to filter out potential duplicate parameters. This can
+        # happen when a module references another module (e.g. the criterion
+        # references the module), thus yielding that module's parameters again.
```
thomasjpfan, line 763: This could be silly question: What is an example of a criterion with learnable parameters?

BenjaminBossan: Not silly. One use case that has been discussed at some point is for GAN training with the generator being the `module` and the discriminator the `criterion` (though you might need different criteria in this case). Another case (which revealed this bug for me) is when using GPyTorch.
But this issue is not only relevant when the criterion references the module, it could also be one module referencing a second module.

thomasjpfan (approved):

Issue #802, opened 2021-09-26
NeuralNetClassifier cannot fit tensor labels because of NeuralNetClassifier.check_data


TheAutumnOfRice:
A small bug.
Ver. 0.10.0
In classifier.py / Line 107-120: (NeuralNetClassifier.check_data)
```python
    def check_data(self, X, y):
        if (
                (y is None) and
                (not is_dataset(X)) and
                (self.iterator_train is DataLoader)
        ):
            msg = ("No y-values are given (y=None). You must either supply a "
                   "Dataset as X or implement your own DataLoader for "
                   "training (and your validation) and supply it using the "
                   "``iterator_train`` and ``iterator_valid`` parameters "
                   "respectively.")
            raise ValueError(msg)
        if y is not None:
            self.classes_inferred_ = np.unique(y)
```
The line 120 `np.unique(y)` limits that `y` should be numpy array but not torch tensor. As Y cannot be cuda tensors, the numpy array `y`  goes through `to_device` function in every epoch, wasting a lot of time.

I think it might be better if `np.unique` or `torch.unique` can be automatically selected according to the type of y.

BenjaminBossan:
Thank you for reporting.

> I think it might be better if `np.unique` or `torch.unique` can be automatically selected according to the type of y.

That sounds reasonable. Would you like to work on that?

TheAutumnOfRice:
@BenjaminBossan 
Thanks for your replying.
I manage to make a pull request. #803 
Hope it helps!

BenjaminBossan:
Solves via #803

PR #803, opened 2021-09-27
Allow NeuralNetClassifier to fit tensor labels #802 

TheAutumnOfRice:
Issue #802 
The original check_data method can only receive ndarray y-values because of np.unique function. Now the method is modified so that tensor y-values can be processed with torch.unique specifically.

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -13,9 +13,8 @@ from skorch.callbacks import PrintLog
 from skorch.callbacks import EpochScoring
 from skorch.callbacks import PassthroughScoring
 from skorch.dataset import CVSplit
-from skorch.utils import get_dim
+from skorch.utils import get_dim, to_numpy
 from skorch.utils import is_dataset
-
 
 neural_net_clf_doc_start = """NeuralNet for classification tasks
 
@@ -117,14 +116,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
             raise ValueError(msg)
         if y is not None:
             # pylint: disable=attribute-defined-outside-init
-            if isinstance(y,np.ndarray):
-                self.classes_inferred_ = np.unique(y)
-            elif isinstance(y,torch.Tensor):
-                self.classes_inferred_ = torch.unique(y)
-            else:
-                msg = (f"If y-values are given, they must be either numpy ndarray or "
-                       f"torch tensors, but not {type(y)}.")
-                raise ValueError(msg)
+            self.classes_inferred_ = np.unique(to_numpy(y))
 
     # pylint: disable=arguments-differ
     def get_loss(self, y_pred, y_true, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


skorch/classifier.py
```diff
@@ -117,7 +117,14 @@ def check_data(self, X, y):
             raise ValueError(msg)
         if y is not None:
             # pylint: disable=attribute-defined-outside-init
-            self.classes_inferred_ = np.unique(y)
+            if isinstance(y,np.ndarray):
```
BenjaminBossan, line 120:
```suggestion
            if isinstance(y, np.ndarray):
```

skorch/classifier.py
```diff
@@ -117,7 +117,14 @@ def check_data(self, X, y):
...
+                self.classes_inferred_ = np.unique(y)
+            elif isinstance(y,torch.Tensor):
```
BenjaminBossan, line 122:
```suggestion
            elif isinstance(y, torch.Tensor):
```

skorch/classifier.py
```diff
@@ -117,7 +117,14 @@ def check_data(self, X, y):
...
+                self.classes_inferred_ = torch.unique(y)
+            else:
+                msg = (f"If y-values are given, they must be either numpy ndarray or "
+                       f"torch tensors, but not {type(y)}.")
+                raise ValueError(msg)
```
BenjaminBossan, line 127:
```suggestion
                raise TypeError(msg)
```

BenjaminBossan (changes requested):
Thanks a lot for taking this!

There are a few stylistic changes you could make and I believe that `TypeError` is more fitting here.

Another thing: For this kind of change, we should have unit tests. Here, it is a bit tricky because the current code, i.e. using `np.unique` on a torch tensor, seems to work. The test might thus require a bit of patching.

If you need help with implementing the tests, feel free to tell me and I'll help you along the way.

BenjaminBossan:
I see now that the change you made introduces an error elsewhere:

```python
    def test_classes_data_torch_tensor(self, net_cls, module_cls, data):
        X = torch.as_tensor(data[0])
        y = torch.as_tensor(np.arange(len(X)) % 10)
    
        net = net_cls(module_cls, max_epochs=0).fit(X, y)
>       assert (net.classes_ == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]).all()
E       AttributeError: 'bool' object has no attribute 'all'
```

This is because `classes_` used to always be a numpy array but now it can be a torch tensor. Please make sure to cast the torch tensor to a numpy array (using `skorch.utils.to_numpy`).

TheAutumnOfRice:
@BenjaminBossan Thanks for your replying!

Since `skorch.utils.to_numpy` is so powerful, we just need to add `to_numpy` to the original code, and everything is done :p

When it comes to unit tests, It's my first time to be required to use them, and I have no ideas how to deal with them haha. I hope this one-line modification can pass all these tests without my efforts.

BenjaminBossan:
> Since `skorch.utils.to_numpy` is so powerful, we just need to add `to_numpy` to the original code, and everything is done :p

That's great, so this is enough to solve your original issue in #802? If so, yeah I believe we don't need extra tests :)

TheAutumnOfRice:
Yeah, that works pretty well :)

BenjaminBossan:
Great, thanks for your work @TheAutumnOfRice

PR #806, opened 2021-10-01
Add logo variant for dark/light mode

### LINKED ISSUE #800: skorch logo is hard to read with dark theme
When using a dark theme, the black font is hard to read. The logo should be designed such that it's legible using both dark and light themes.

ottonemo:
Should be readable in both cases using a white border.

Resolves #800.


BenjaminBossan:
skorch-logo-border

This is how it looks like for me in dark mode. So this is much better now, good to go for me.

I wondered about the "clover" in the "O": Since the lines are very thin, it looks a bit pixelated on my 1080p screen. Would it be better to just completely fill the "O"? Maybe I'm wrong and it would look bad, then disregard my comment.

Issue #808, opened 2021-10-20
EarlyStopping should have a restore_best_weights arg as in Keras


cedricrommel:
The EarlyStopping callback is very useful to avoid overfitting as it allows to stop training before the validation loss (or any other metric of interest) starts degrading. The problem I see with Skorch's implementation is that such callback only stops the training after the desired patience is reached, keeping the model at its state from the last epoch. It does not allow the user to reload the best model from `patience` epochs before.

This can be problematic for models where the training is noisy (often the case), which need large patience values to be properly regularized. In these cases, the model obtained at the end might have overfitted quite a lot already. This is why other user-friendly deep learning libraries such as Keras, which also implement the EarlyStopping callback, have an argument `restore_best_weights` which allows to restore the best model, from `patience` epochs before earlystopping. I really believe that such an option should be added to the `EarlyStopping` implementation in Skorch, not only because it is useful, but also because it would be coherent with Keras, which is very popular.

BenjaminBossan:
The PR was merged, thus this issue can be closed.

PR #809, opened 2021-10-20
Add `restore_best_weights` argument to EarlyStopping callback

cedricrommel:
This PR is a proposition to solve the issue https://github.com/skorch-dev/skorch/issues/808

It adds a new argument to the EarlyStopping callback allowing to retrieve the best model after stopping the training (from `patience` epochs before the end of the training). This useful for most situations in which EarlyStopping is used and at the same time makes the skorch API more coherent with other user-friendly deeplearning libraries which also implement EarlyStopping callbacks (e.g. Keras).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `load_best` attribute to `EarlyStopping` callback to automatically load module weights of the best result at the end of training
 
 ### Changed
 
@@ -18,7 +19,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `load_best` attribute to `Checkpoint` callback to automatically load state of the best result at the end of training
-- Added `load_best` attribute to `EarlyStopping` callback to automatically load module weights of the best result at the end of training
 - Added a `get_all_learnable_params` method to retrieve the named parameters of all PyTorch modules defined on the net, including of criteria if applicable
 - Added `MlflowLogger` callback for logging to Mlflow (#769)
 - Added `InputShapeSetter` callback for automatically setting the input dimension of the PyTorch module
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `load_best` attribute to `Checkpoint` callback to automatically load state of the best result at the end of training
+- Added `load_best` attribute to `EarlyStopping` callback to automatically load module weights of the best result at the end of training
 - Added a `get_all_learnable_params` method to retrieve the named parameters of all PyTorch modules defined on the net, including of criteria if applicable
 - Added `MlflowLogger` callback for logging to Mlflow (#769)
 - Added `InputShapeSetter` callback for automatically setting the input dimension of the PyTorch module
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -396,6 +396,11 @@ class EarlyStopping(Callback):
         self.sink = sink
         self.load_best = load_best
 
+    def __getstate__(self):
+        # Avoids to save the module_ weights twice when pickling
+        self.best_model_weights_ = None
+        return self.__dict__
+
     # pylint: disable=arguments-differ
     def on_train_begin(self, net, **kwargs):
         if self.threshold_mode not in ['rel', 'abs']:
@@ -424,7 +429,10 @@ class EarlyStopping(Callback):
             raise KeyboardInterrupt
 
     def on_train_end(self, net, **kwargs):
-        if self.load_best and self.best_epoch_ != net.history[-1, "epoch"]:
+        if (
+            self.load_best and self.best_epoch_ != net.history[-1, "epoch"] and
+            self.best_model_weights_ is not None
+        ):
             net.module_.load_state_dict(self.best_model_weights_)
             self._sink("Restoring best model from epoch {}.".format(
                 self.best_epoch_
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -398,7 +398,7 @@ class EarlyStopping(Callback):
 
     def __getstate__(self):
         # Avoids to save the module_ weights twice when pickling
-        state = self.__dict__
+        state = self.__dict__.copy()
         state['best_model_weights_'] = None
         return state
 
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -358,19 +358,21 @@ class EarlyStopping(Callback):
       Ignore score improvements smaller than `threshold`.
 
     threshold_mode : str (default='rel')
-        One of `rel`, `abs`. Decides whether the `threshold` value is
-        interpreted in absolute terms or as a fraction of the best
-        score so far (relative)
+      One of `rel`, `abs`. Decides whether the `threshold` value is
+      interpreted in absolute terms or as a fraction of the best
+      score so far (relative)
 
     sink : callable (default=print)
       The target that the information about early stopping is
       sent to. By default, the output is printed to stdout, but the
       sink could also be a logger or :func:`~skorch.utils.noop`.
 
-    restore_best_weights: bool (default=False)
-        Whether to restore model weights from the epoch with the best value of
-        the monitored quantity. If False, the model weights obtained at the
-        last step of training are used.
+    load_best: bool (default=False)
+      Whether to restore module weights from the epoch with the best value of
+      the monitored quantity. If False, the module weights obtained at the
+      last step of training are used. Note that only the module is restored and
+      that the ``Checkpoint`` callback with the :func:`~Checkpoint.load_best`
+      argument set to ``True``.
 
     """
     def __init__(
@@ -381,7 +383,7 @@ class EarlyStopping(Callback):
             threshold_mode='rel',
             lower_is_better=True,
             sink=print,
-            restore_best_weights=False,
+            load_best=False,
     ):
         self.monitor = monitor
         self.lower_is_better = lower_is_better
@@ -391,7 +393,7 @@ class EarlyStopping(Callback):
         self.misses_ = 0
         self.dynamic_threshold_ = None
         self.sink = sink
-        self.restore_best_weights = restore_best_weights
+        self.load_best = load_best
 
     # pylint: disable=arguments-differ
     def on_train_begin(self, net, **kwargs):
@@ -410,16 +412,16 @@ class EarlyStopping(Callback):
         else:
             self.misses_ = 0
             self.dynamic_threshold_ = self._calc_new_threshold(current_score)
-            if self.restore_best_weights:
-                self.best_model_weights_ = net.module.state_dict()
-                self.best_epoch_ = net.history[-1, "epoch"]
+            self.best_epoch_ = net.history[-1, "epoch"]
+            if self.load_best:
+                self.best_model_weights_ = net.module_.state_dict()
         if self.misses_ == self.patience:
             if net.verbose:
                 self._sink("Stopping since {} has not improved in the last "
                            "{} epochs.".format(self.monitor, self.patience),
                            verbose=net.verbose)
-                if self.restore_best_weights:
-                    net.module.load_state_dict(self.best_model_weights_)
+                if self.load_best:
+                    net.module_.load_state_dict(self.best_model_weights_)
                     self._sink("Restoring best model from epoch {}.".format(
                         self.best_epoch_
                     ), verbose=net.verbose)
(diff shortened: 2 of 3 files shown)


skorch/callbacks/training.py
```diff
@@ -358,15 +358,22 @@ class EarlyStopping(Callback):
       Ignore score improvements smaller than `threshold`.
 
     threshold_mode : str (default='rel')
-        One of `rel`, `abs`. Decides whether the `threshold` value is
-        interpreted in absolute terms or as a fraction of the best
-        score so far (relative)
+      One of `rel`, `abs`. Decides whether the `threshold` value is
+      interpreted in absolute terms or as a fraction of the best
+      score so far (relative)
```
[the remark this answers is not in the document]
cedricrommel: Of course. I just applied some basic boy-scout logic as you said that skorch identation was supposed to be 2 spaces (and it was 4 there).

skorch/callbacks/training.py
```diff
@@ -358,15 +358,22 @@ class EarlyStopping(Callback):
...
       The target that the information about early stopping is
       sent to. By default, the output is printed to stdout, but the
       sink could also be a logger or :func:`~skorch.utils.noop`.
 
+    load_best: bool (default=False)
+      Whether to restore module weights from the epoch with the best value of
+      the monitored quantity. If False, the module weights obtained at the
+      last step of training are used. Note that only the module is restored and
+      that the ``Checkpoint`` callback with the :attr:`~Checkpoint.load_best`
+      argument set to ``True``.
```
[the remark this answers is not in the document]
cedricrommel: Indeed ^^

skorch/callbacks/training.py
```diff
@@ -401,11 +412,19 @@ def on_epoch_end(self, net, **kwargs):
...
             self.dynamic_threshold_ = self._calc_new_threshold(current_score)
+            self.best_epoch_ = net.history[-1, "epoch"]
+            if self.load_best:
+                self.best_model_weights_ = net.module_.state_dict()
         if self.misses_ == self.patience:
             if net.verbose:
                 self._sink("Stopping since {} has not improved in the last "
                            "{} epochs.".format(self.monitor, self.patience),
                            verbose=net.verbose)
+                if self.load_best:
```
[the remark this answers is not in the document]
cedricrommel: Oh, sure ! My bad, I totally missed this point. I agree that `on_train_end` is a better location to implement this, as it comes only with benefits and no costs. No need for discussion on this :)

skorch/tests/callbacks/test_training.py
```diff
@@ -487,6 +497,53 @@ def test_typical_use_case_nonstop(
 
         assert len(net.history) == max_epochs
 
+    def test_weights_restore(
+        self, net_clf_cls, broken_classifier_module, classifier_data,
+            early_stopping_cls):
```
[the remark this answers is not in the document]
cedricrommel: Yes, there was a tab missing, sorry :)

skorch/tests/callbacks/test_training.py
```diff
@@ -497,7 +554,10 @@ def test_typical_use_case_stopping(
         def sink(x):
             side_effect.append(x)
 
-        early_stopping_cb = early_stopping_cls(patience=patience, sink=sink)
+        early_stopping_cb = early_stopping_cls(
+            patience=patience,
+            sink=sink,
+        )
```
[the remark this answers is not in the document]
cedricrommel: Sure!

cedricrommel:
Thanks for the comments! Just pushed a new version addressing your concerns @BenjaminBossan. Concerning point 1, it was not my linter but I indeed changed a few unrelated code lines unconsciously sometimes and on purpose for the docstring as I though I was fixing something. Sorry for that :) Concerning point 2 I totally agree with you and just forgot to add it to my last commits. Hope this time is the good one 🤞

cedricrommel:
Btw, I've used the code in an experiment and I have clues indicating that it is not working as expected 😞 So please don't merge yet even if I addressed your concerns. I might have designed too weak of a test for the feature. I'm checking it tomorrow morning.

cedricrommel:
There it is @BenjaminBossan. There was indeed an error, as I was storing in EarlyStopping the mutable `net.module_.state_dict` object without copying it, which would keep being modified. We were hence reloading the `state_dict` from the last epoch.  And this was not caught by the tests because I was also not making a copy of the initial `state_dict` 😓

So now this is fixed and i'm 100% it works. I've also robustified the tests to make sure it indeed checks what we want.

skorch/tests/callbacks/test_training.py
```diff
@@ -487,6 +493,89 @@ def test_typical_use_case_nonstop(
...
+        # check correct output messages
+        assert len(side_effect) == 2
+
+        msg = side_effect[0]
+        expected_msg = ("Stopping since valid_acc has not improved in "
+                        "the last 3 epochs.")
+        assert msg == expected_msg
+
+        msg = side_effect[1]
+        expected_msg = ("Restoring best model from epoch ")
```
BenjaminBossan, line 546:
```suggestion
        expected_msg = "Restoring best model from epoch "
```

Do we not know the actual epoch, to include it into the message?

skorch/tests/callbacks/test_training.py
```diff
@@ -487,6 +493,89 @@ def test_typical_use_case_nonstop(
...
+        assert expected_msg in msg
+
+        # Extract epoch reloaded by early stopping
+        loaded_epoch = int(msg.split()[-1][:-1])
```
BenjaminBossan, line 550: Can we use `early_stopping_cb.best_epoch_` instead?

CHANGES.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `load_best` attribute to `Checkpoint` callback to automatically load state of the best result at the end of training
+- Added `load_best` attribute to `EarlyStopping` callback to automatically load module weights of the best result at the end of training
```
BenjaminBossan, line 13: We had a skorch release recently, resulting in the addition of a new section in the CHANGES.md. Could you please merge with/rebase against master. Your change will be at the very top of the new release :)

skorch/callbacks/training.py
```diff
@@ -401,13 +413,23 @@ def on_epoch_end(self, net, **kwargs):
...
+                self.best_model_weights_ = deepcopy(net.module_.state_dict())
         if self.misses_ == self.patience:
             if net.verbose:
                 self._sink("Stopping since {} has not improved in the last "
                            "{} epochs.".format(self.monitor, self.patience),
                            verbose=net.verbose)
             raise KeyboardInterrupt
 
+    def on_train_end(self, net, **kwargs):
+        if self.load_best:
```
BenjaminBossan, line 427: Could you please add a check whether the current epoch is not already the best epoch? In that case, restoring would be unnecessary.

BenjaminBossan (changes requested):
Thank you a lot for being so diligent and for being open to suggestions. I have a few minor comments still, please take a look at them.

> I just applied some basic boy-scout logic as you said that skorch identation was supposed to be 2 spaces (and it was 4 there).

This was a good idea. However, we use 4 spaces everywhere, but docstrings use 2 spaces for paragraphs.

> There was indeed an error, as I was storing in EarlyStopping the mutable net.module_.state_dict object without copying it, which would keep being modified.

Ah yes, that's a nasty error, I didn't catch it either.

There is one more thing that comes to my mind, sorry that I didn't think of it earlier: Given this implementation, when someone pickles the net, it would be twice as big because both the weights of the `module_` and the `EarlyStopping.best_model_weights_` would be stored. I wonder if that is a good thing. What do you think about not storing `EarlyStopping.best_model_weights_`?

In that case, one would need to override `__getstate__` from `EarlyStopping` and pop the `best_model_weights_`, probably replacing it with `None`. In `on_train_end`, there would need to be an extra check if `best_model_weights_ is not None` before loading it into the `state_dict`.

I'm trying to imagine if this could cause trouble in practice. Say someone trains a model with early stopping and afterwards pickles the whole model. If later on they decide to load the model and continue training, they would not need the `best_model_weights_` attribute because these weights are already equal to the model's current weights. Is there some other use case where this could matter?

skorch/tests/callbacks/test_training.py
cedricrommel: We do (14) but I preferred not to include it in case something changes in the `conftest.py` or else at some point and this number changes. The person creating the PR will then be confused about the reason this test will not be passing anymore.

skorch/tests/callbacks/test_training.py
cedricrommel: indeed :)

CHANGES.md
cedricrommel: Sure!

skorch/callbacks/training.py
cedricrommel: Good point. I'll add the check that `best_model_weights_` is not `None` as well, as advised above concerning pickling.

cedricrommel:
Indeed, your remark concerning the unnecessary saving of the `best_model_weights_` makes sense and is useful specially when considering large models. I've thought about it and can't find any practical drawback, so I've implemented it as advised to gether with the other remarks :)

skorch/callbacks/training.py
```diff
@@ -385,6 +394,12 @@ def __init__(
         self.misses_ = 0
         self.dynamic_threshold_ = None
         self.sink = sink
+        self.load_best = load_best
+
+    def __getstate__(self):
+        # Avoids to save the module_ weights twice when pickling
+        self.best_model_weights_ = None
+        return self.__dict__
```
BenjaminBossan, line 402:
```suggestion
        state = super().__getstate__()
        state['best_model_weights_'] = None
        return state
```

Just to be safe, let's call `super()` here.

skorch/callbacks/training.py
```diff
@@ -401,13 +418,26 @@ def on_epoch_end(self, net, **kwargs):
...
             if net.verbose:
                 self._sink("Stopping since {} has not improved in the last "
                            "{} epochs.".format(self.monitor, self.patience),
                            verbose=net.verbose)
             raise KeyboardInterrupt
 
+    def on_train_end(self, net, **kwargs):
+        if (
+            self.load_best and self.best_epoch_ != net.history[-1, "epoch"] and
+            self.best_model_weights_ is not None
```
BenjaminBossan, line 434:
```suggestion
            self.load_best and (self.best_epoch_ != net.history[-1, "epoch"])
            and (self.best_model_weights_ is not None)
```

This is probably not necessary but I'm always a bit skittish when it comes to precedence in Python.

BenjaminBossan (commented):
Fantastic, thanks for the hard work. I only have two minor comments left, then we're good to go.

cedricrommel:
Indeed, your way of doing it is safer ^^ Thanks @BenjaminBossan !

cedricrommel:
So the thing is that `super` has not a `__gtestate__` method defined:
```
self = <skorch.callbacks.training.EarlyStopping object at 0x7fbd6801d250>

    def __getstate__(self):
        # Avoids to save the module_ weights twice when pickling
>       state = super().__getstate__()
E       AttributeError: 'super' object has no attribute '__getstate__'

../../miniconda3/envs/braindecode/lib/python3.9/site-packages/skorch/callbacks/training.py:401: AttributeError
```

I'm hence adding the `return __dict__` back if you don't mind.
I've also added a test to check the pickling works as expected.

skorch/callbacks/training.py
```diff
@@ -398,7 +398,7 @@ def __init__(
 
     def __getstate__(self):
         # Avoids to save the module_ weights twice when pickling
-        state = super().__getstate__()
+        state = self.__dict__
```
BenjaminBossan, line 401: Could you please use `self.__dict__.copy()`. I believe otherwise you would delete it from the existing object.

skorch/tests/callbacks/test_training.py
```diff
@@ -547,32 +548,42 @@ def sink(x):
...
             )
         )
 
         # Check validation loss obtained match
-        y_proba_2 = net.predict_proba(val_dataset)
+        y_proba_2 = net2.predict_proba(val_dataset)
         assert es_loss == log_loss(y_val, y_proba_2)
 
+        # Check best_model_weights_ is transformed into None when pickling
+        with tempfile.NamedTemporaryFile() as tmp_file:
```
BenjaminBossan, line 578: You could use `pickle.dumps` and `pickle.loads` so that you don't need to create an actual file.

BenjaminBossan (commented):
My bad regarding the `super().__getstate__`, we don't actually have a custom one on callbacks, sorry for the confusion.

I have (yet again) two small comments, please take a look. The second one for the unit test is just a proposed simplification, you may or may not implement it. After that, however, I'm really certain we're good to go this time :)

skorch/callbacks/training.py
cedricrommel: Yes, good point !

skorch/tests/callbacks/test_training.py
cedricrommel: Oh yes, that's way simpler ! :)

cedricrommel:
Thanks for the comments ! There it is !

BenjaminBossan:
Great work, thanks for your patience.

PR #815, opened 2021-10-31
Docs: pin docutils to fix broken Sphinx build

BenjaminBossan:
Sphinx build is currently broken:

https://readthedocs.org/projects/skorch/builds/15141681/

I tried the fix of pinning docutils<0.18 suggested [here](https://github.yuuza.net/sphinx-doc/sphinx/issues/9727) and it works:

https://readthedocs.org/projects/skorch/builds/15141724/

The better long term solution would be to use a newer version of Sphinx but I haven't checked how to do that yet and I wouldn't be surprised if this surfaces some new problems, hence this short term solution for now.

(Please disregard the other commit in this PR.)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -2,8 +2,8 @@
 set -e
 
 PYTORCH_VERSION=${PYTORCH_VERSION:-""}
-PYTHON_VERSION="3.7"
-TWINE_VERSION="1.12.1"
+PYTHON_VERSION="3.9"
+TWINE_VERSION=">3,<4.0.0dev"
 CONDA_ENV="skorch-deploy"
 CONDA_ENV_YML="environment.yml"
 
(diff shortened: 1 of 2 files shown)


scripts/deploy.sh
```diff
@@ -2,8 +2,8 @@
 set -e
 
 PYTORCH_VERSION=${PYTORCH_VERSION:-""}
-PYTHON_VERSION="3.7"
-TWINE_VERSION="1.12.1"
+PYTHON_VERSION="3.9"
+TWINE_VERSION=">3,<4.0.0dev"
```
thomasjpfan, line 6: This is unrelated to sphinx, but I see no harm in updating twine here.

thomasjpfan (approved):
+1 on the a quick fix since we are still on `Sphinx==1.8.5`.

Issue #760, opened 2021-04-25
Consider integrating Huggingface Accelerate


BenjaminBossan:
[Accelerate](https://github.com/huggingface/accelerate) by Huggingface promises to be "A simple way to train and use PyTorch models with multi-GPU, TPU, mixed-precision".

A possible implementation could be to allow passing an `Accelerator` instance as the `device` argument and to adopt `to_device` in order to deal with it. We may possibly need to add a hook in order to run the line `model, optim, data = accelerator.prepare(model, optim, data)` (where `data` seems to be the data loader).

There are some open questions with this, e.g. if our code/specific callbacks are safe to use with Accelerate. Also, how would this interact with (replace?) our AMP feature (#707)?

BenjaminBossan:
Maybe it's possible to just wrap the optimizer using [AcceleratedOptimizer](https://github.com/huggingface/accelerate/blob/a5ea2a932ccb7c9c880d2151c7dfcb13a12ecc95/src/accelerate/optimizer.py#L39) or we could implement something similar for skorch. Using `autocast` would be left for the user to apply.

BenjaminBossan:
This has been solved via #826

PR #826, opened 2021-12-23
Experimental support for accelerate

BenjaminBossan:
## Description

Proposal to implement [huggingface accelerate](https://github.com/huggingface/accelerate) (#760). If this is adopted, we would probably no longer need implement AMP support, since this is a subset of the offered functionality of accelerate.

For lack of appropriate hardware, I could not test if this would actually work or not.

## Implementation

First idea was to implement this as a callback but that would require to add a lot of new `on_` -methods that would serve only this callback for now.

Therefore, this feature is implemented as a mixin class to serve as proof of concept of how this could look like if implemented in the `NeuralNet` class itself.

I'm not sure if we would need to use [`accelerator.utils.save`](https://github.com/huggingface/accelerate/blob/a5ea2a932ccb7c9c880d2151c7dfcb13a12ecc95/src/accelerate/utils.py#L536).

The failing tests seem to be unrelated to this change.

### DIFF
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -48,6 +48,53 @@ argument ``idx=0``, the default) and one for y (with argument
     gs.fit(X_sl, y_sl)
 
 
+AccelerateMixin
+---------------
+
+This mixin class can be used to add support for huggingface accelerate_ to
+skorch. E.g., this allows you to use mixed precision training (AMP), multi-GPU
+training, or training with a TPU. For the time being, this feature should be
+considered experimental.
+
+To use this feature, create a new subclass of the neural net class you want to
+use and inherit from the mixin class. E.g., if you want to use a
+:class:`.NeuralNet`, it would look like this:
+
+.. code:: python
+
+    from skorch import NeuralNet
+    from skorch.helper import AccelerateMixin
+
+    class AcceleratedNet(AccelerateMixin, NeuralNet):
+        """NeuralNetClassifier with accelerate support"""
+
+The same would work for :class:`.NeuralNetClassifier`,
+:class:`.NeuralNetRegressor`, etc. Then pass an instance of Accelerator_ with
+the desired parameters and you're good to go:
+
+.. code:: python
+
+    from accelerate import Accelerator
+
+    accelerator = Accelerator(...)
+    net = AcceleratedNet(
+        MyModule,
+        accelerator=accelerator,
+        device=None,
+        callbacks__print_log__sink=accelerator.print)
+    net.fit(X, y)
+
+accelerate_ recommends to leave the device handling to the Accelerator_, which
+is why we set ``device=None`` (thus telling skorch not to change the device).
+Furthermore, using ``accelerator.print`` should avoid printing the same output
+multiple times when training concurrently on multiple machines.
+
+To install accelerate_, run the following command inside your Python environment:
+
+    .. code:: bash
+
+      python -m pip install accelerate
+
 Command line interface helpers
 ------------------------------
 
@@ -201,6 +248,8 @@ callbacks through the command line (but you can modify existing ones
 as usual).
 
 
+.. _accelerate: https://github.com/huggingface/accelerate
+.. _Accelerator: https://huggingface.co/docs/accelerate/accelerator.html
 .. _fire: https://github.com/google/python-fire
 .. _numpydoc: https://github.com/numpy/numpydoc
 .. _example: https://github.com/skorch-dev/skorch/tree/master/examples/cli
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -80,12 +80,12 @@ the desired parameters and you're good to go:
     net = AcceleratedNet(
         MyModule,
         accelerator=accelerator,
-        device=None,
     )
     net.fit(X, y)
 
 accelerate_ recommends to leave the device handling to the Accelerator_, which
-is why we set ``device=None`` (thus telling skorch not to change the device).
+is why ``device`` defautls to ``None`` (thus telling skorch not to change the
+device).
 
 To install accelerate_, run the following command inside your Python environment:
 
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -66,7 +66,7 @@ use and inherit from the mixin class. E.g., if you want to use a
     from skorch.helper import AccelerateMixin
 
     class AcceleratedNet(AccelerateMixin, NeuralNet):
-        """NeuralNetClassifier with accelerate support"""
+        """NeuralNet with accelerate support"""
 
 The same would work for :class:`.NeuralNetClassifier`,
 :class:`.NeuralNetRegressor`, etc. Then pass an instance of Accelerator_ with
@@ -89,9 +89,16 @@ device).
 
 To install accelerate_, run the following command inside your Python environment:
 
-    .. code:: bash
+.. code:: bash
 
       python -m pip install accelerate
+
+.. note::
+
+    Under the hood, accelerate uses :class:`~torch.cuda.amp.GradScaler`,
+    which does not support passing the training step as a closure.
+    Therefore, if your optimizer requires that (e.g.
+    :class:`torch.optim.LBFGS`), you cannot use accelerate.
 
 Command line interface helpers
 ------------------------------
--- a/docs/user/helper.rst
+++ b/docs/user/helper.rst
@@ -81,13 +81,11 @@ the desired parameters and you're good to go:
         MyModule,
         accelerator=accelerator,
         device=None,
-        callbacks__print_log__sink=accelerator.print)
+    )
     net.fit(X, y)
 
 accelerate_ recommends to leave the device handling to the Accelerator_, which
 is why we set ``device=None`` (thus telling skorch not to change the device).
-Furthermore, using ``accelerator.print`` should avoid printing the same output
-multiple times when training concurrently on multiple machines.
 
 To install accelerate_, run the following command inside your Python environment:
 
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -12,6 +12,7 @@ from sklearn.base import TransformerMixin
 import torch
 
 from skorch.cli import parse_args  # pylint: disable=unused-import
+from skorch.dataset import unpack_data
 from skorch.utils import _make_split
 from skorch.utils import is_torch_data_type
 from skorch.utils import to_tensor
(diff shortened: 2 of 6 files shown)


skorch/helper.py
```diff
@@ -508,3 +509,83 @@ def describe_signature(self, df):
...
+                    self.accelerator.prepare(criterion)
+
+            return self
+
+    def _initialize_module(self, *args, **kwargs):
+        super()._initialize_module(*args, **kwargs)
+
+        with self._current_init_context('module'):
+            for name in self._modules:
+                module = getattr(self, name + '_')
```
thomasjpfan, line 561: The pattern of `getattr(self, ... + '_')` is used quite a bit in the codebase. Makes me think storing `_` may have been a bit nicer.

(**Not actionable here**)

skorch/helper.py
```diff
@@ -508,3 +509,83 @@ def describe_signature(self, df):
             )
 
         return signature
+
+
+class AccelerateMixin:
+    """Mixin class to add support for huggingface's accelerate
+
+    *Experimental*
```
thomasjpfan, line 517: We can include a link to https://huggingface.co/docs/accelerate/index.html and how to install in the docstring

skorch/helper.py
```diff
@@ -508,3 +509,83 @@ def describe_signature(self, df):
...
+                if isinstance(optimizer, torch.optim.Optimizer):
+                    self.accelerator.prepare(optimizer)
+        return self
+
+    def train_step_single(self, batch, **fit_params):
+        self._set_training(True)
+        Xi, yi = unpack_data(batch)
+        y_pred = self.infer(Xi, **fit_params)
+        loss = self.get_loss(y_pred, yi, X=Xi, training=True)
+        self.accelerator.backward(loss)
```
thomasjpfan, line 582: The [accelerator docs](https://huggingface.co/docs/accelerate/index.html) says it better to let `accelerate` handle device transfers. Is it possible to do the same in `skorch`?

skorch/tests/test_helper.py
```diff
@@ -708,3 +708,47 @@ def test_describe_signature_other_dtypes(self, transformer_cls, df):
...
+            max_epochs=10,
+            lr=0.1,
+            accelerator=accelerator,
+            device=None,
+            callbacks__print_log__sink=accelerator.print,
+        )
+
+    def test_fit_fp16(self, net, data):
+        X, y = data
+        net.fit(X, y)  # does not raise
```
thomasjpfan, line 754: I do not see a public way to check that `accerlate` did something without mocking. All module and optimizer state in `Accelerator` is stored as a private attribute.

skorch/tests/test_helper.py
```diff
@@ -708,3 +708,47 @@ def test_describe_signature_other_dtypes(self, transformer_cls, df):
...
+        pytest.importorskip('accelerate')
+
+        from accelerate import Accelerator
+
+        accelerator = Accelerator(fp16=True)
+        return accelerator
+
+    @pytest.fixture
+    def net(self, net_cls, module_cls, accelerator):
+        return net_cls(
```
thomasjpfan, line 743: As a sanity check, it would be good to check if the `NeuralNet` can be pickled and unpickled with an accelerator.

thomasjpfan (commented):
Thank you for working on this! I tested on hardware and can confirm this works.

Using a mixin adds another design pattern, which can signal a limitation of the callbacks. The most user-friendly way would be to add a kwarg into `NeuraNet`, but that adds more complexity to `NeuraNet`. It's all trade-offs now.

I am happy to try mixins out.

skorch/helper.py
BenjaminBossan: I agree. I was thinking about something like:

```python
        def get_modules(self):
            for name in self._modules:
                module = getattr(self, name + '_')
                if isinstance(module, nn.module):
                    yield module
```

and replacing all those lines like the quoted one with a call to that method. Same for criteria and optimizers. But that would be for another PR.

skorch/helper.py
BenjaminBossan: If we agree to proceed with this design, I would add links and more information.

skorch/helper.py
BenjaminBossan: I thought this is happening when `device_placement=True`, which is the default? That's why I check that attribute during initialization and require `device=None` in that case, i.e. that skorch does not bother with moving to the desired device. Or do you mean something else?

skorch/tests/test_helper.py
BenjaminBossan: I will probably add tests with a fake accelerator that sets an attribute on the object (`module.accelerated = True`) or something like this for testing.

skorch/tests/test_helper.py
BenjaminBossan: Good point.

BenjaminBossan:
> I tested on hardware and can confirm this works.

Nice, thanks for testing this out. I would assume that certain combinations of settings might cause problems, but it will be impossible to check all of them out (especially not with the existing CI). I would also be happy if we could cover the most common use cases, everything else would be almost unfeasible, given the multitude of options available.

> The most user-friendly way would be to add a kwarg into NeuraNet, but that adds more complexity to NeuraNet. It's all trade-offs now.

This is what I meant when I wrote

> this feature is implemented as a mixin class to serve as proof of concept of how this could look like if implemented in the NeuralNet class itself.

but I could have written that more clearly. I agree, the reason why I started like this is just to get something out quickly and collect feedback (especially since it might not have worked at all). If we agree with the design, I could move the functionality to `NeuralNet` itself.

If we decide to go this way, what do you think would be better: Having a couple of `if self.accelerator is not None` checks (verbose but more transparent) or having a fake accelerator object that just performs a no-op when `self.accelerator.prepare` is called (less verbose but less transparent).

skorch/helper.py
thomasjpfan: Ah I see, I missed the `device=None`.

thomasjpfan:
> Having a couple of if self.accelerator is not None checks (verbose but more transparent) or having a fake accelerator object that just performs a no-op when self.accelerator.prepare is called (less verbose but less transparent).

If we place the logic into `NeuralNet`, I would go with the `None` check.

BenjaminBossan:
@thomasjpfan I discussed this with @ottonemo and we decided to leave the feature as a mixin class for now and not to integrate it into `NeuralNet`. The reason is that accelerate is still quite young and its API might change in the future. By not integrating it into `NeuralNet` yet, it becomes clear that this is an experimental feature with possible BC changes coming.

thomasjpfan:
Okay, lets go with the mixin.

BenjaminBossan:
@thomasjpfan I added more documentation of this new feature. This should hopefully help users to discover the feature (and provide feedback if there is any issue).

docs/user/helper.rst
```diff
@@ -48,6 +48,53 @@ argument ``idx=0``, the default) and one for y (with argument
...
+.. code:: python
+
+    from accelerate import Accelerator
+
+    accelerator = Accelerator(...)
+    net = AcceleratedNet(
+        MyModule,
+        accelerator=accelerator,
+        device=None,
+        callbacks__print_log__sink=accelerator.print)
```
thomasjpfan, line 84: Should we do this automatically? This feels like more boilerplate for a user to write.

BenjaminBossan: Yes, true. On the other hand, since the accelerator does not exist when we define `AccelerateMixin`, we cannot set it as default argument. We would need to do something else, like:

```python
def initialize(self):
    super().initialize()
    self.set_params(callbacks__print_log__sink=self.accelerator.print)
    return self
```

Not sure if the user would have any reason _not_ to set this, but if they had, this hard-coding would prove annoying.

thomasjpfan: I think having good defaults is generally better. If a user does not want to set it like this, I'm guessing they are more advanced and would need to to override `initialize` to undo it.

It's a toss up for me.

skorch/tests/test_helper.py
```diff
@@ -708,3 +711,183 @@ def test_describe_signature_other_dtypes(self, transformer_cls, df):
...
+            accelerator = accelerator_cls(mixed_precision=mixed_precision)
+        elif mixed_precision == 'bf16':
+            pytest.skip('bf16 only supported in accelerate version > 0.5.1')
+        else:
+            fp16 = mixed_precision == 'fp16'
+            accelerator = accelerator_cls(fp16=fp16)
+
+        net = net_cls(accelerator=accelerator)
+        X, y = data
+        net.fit(X, y)  # does not raise
```
thomasjpfan, line 766: The following assertion fails here:

```suggestion
        net.fit(X, y)  # does not raise
        assert not np.isnan(net.history[:, "train_loss"]).all()
```

(I have not debugged this yet)

skorch/helper.py
```diff
@@ -508,3 +509,131 @@ def describe_signature(self, df):
...
+    >>> from accelerate import Accelerator
+    >>>
+    >>> class AcceleratedNet(AccelerateMixin, NeuralNetClassifier):
+    ...     '''NeuralNetClassifier with accelerate support'''
+    >>>
+    >>> accelerator = Accelerator(...)
+    >>> net = AcceleratedNet(
+    ...     MyModule,
+    ...     accelerator=accelerator,
+    ...     device=None,
```
thomasjpfan, line 553: We likely can make `device="auto"` too.

BenjaminBossan: I'm not so sure on this one, as it's not as clear cut as the other default. Also, changing device is not as obscure as the print log sink.

thomasjpfan: As with `print`, I can not really think of a case where `device` is not `None` when using `Accelerator`.

BenjaminBossan: Okay, you convinced me, I changed the default to `None`.

skorch/tests/test_helper.py
BenjaminBossan: ping

thomasjpfan: Turning on `fp16` results in weights being nan. I think we can debug this on a CPU:

```python
    def test_force_cpu(self, net_cls, accelerator_cls, data):
        accelerator = accelerator_cls(cpu=True)
        net = net_cls(accelerator=accelerator)
        net.fit(*data)  # does not raise
        
        # this fails
        assert np.isfinite(net.history[:, "train_loss"]).all()
```

thomasjpfan: Looks like everything works now!

thomasjpfan (commented):
I think we can install accelerate on the CI and run the CPU tests.

BenjaminBossan:
@thomasjpfan I did run into one more bug when I got my hands on an NVIDIA T4 GPU. Namely, when using fp16, accelerate uses `GradScaler` under the hood, which [does not support](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler.step) passing the train step as a 'closure'. However, this is what we do by default. Thus I had to sidestep that.

I think in the future, we might consider having the 'closure' pattern be an explicit option that the user can opt into via a parameter, instead of doing it implicitly. Then we can also give a proper error when a user tries to use closure with accelerate.

thomasjpfan (approved):
Thanks for the fix with closures!

I ran the tests locally with an Ampere GPU and everything passes. I also experimented with `notebooks/MNIST.ipynb` to use `Accelerator` and fp16 works as expected.

LGTM

BenjaminBossan:
Thanks @thomasjpfan for doing another round of testing. So now we know it works on Ampere, Turing, and CPU :D That's probably sufficient testing for an experimental feature.

Issue #838, opened 2022-03-04
Isinstance checks for huggingface BatchEncoding


BenjaminBossan:
When using tokenizers from huggingface's `transformers` library, we run into a small annoyance:

```python
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained('bert-base-uncased')
tokens = tokenizer(X)
type(tokens)  # returns transformers.tokenization_utils_base.BatchEncoding,
skorch.dataset.Dataset(tokens)  # fails
```

This is a type that skorch does not understand. The catch is, however, that the type is _almost_ a dictionary. Using:

```python
tokens_dict = {key: val for key, val in tokenizer(X).items}
skorch.dataset.Dataset(tokens_dict)  # works
```
The only problem is that `isinstance(tokens, dict)` fails because it does not inherit from `dict`. Instead, when looking at the `__mro__` of `BatchEncoding`, we get:

```python
(transformers.tokenization_utils_base.BatchEncoding,
 collections.UserDict,
 collections.abc.MutableMapping,
 collections.abc.Mapping,
 collections.abc.Collection,
 collections.abc.Sized,
 collections.abc.Iterable,
 collections.abc.Container,
 object)
```

So one possible solution for us would be to always check for `UserDict` on top of `dict` where we do `isinstance` checks. A quick check turned up these 4 places, but there might be more:

https://github.com/skorch-dev/skorch/blob/6ca5c4e20357963e845e8a0a488b7c7dffdf82f7/skorch/utils.py#L94
https://github.com/skorch-dev/skorch/blob/6ca5c4e20357963e845e8a0a488b7c7dffdf82f7/skorch/utils.py#L126
https://github.com/skorch-dev/skorch/blob/baf058038e299242db3ff6ca22300ba7a46f129e/skorch/dataset.py#L43
https://github.com/skorch-dev/skorch/blob/c0df4d1a11e1e023bc7ce868efaa266d6964f203/skorch/net.py#L1361

This is not the most elegant solution but I can't think of anything better if we want to integrate more smoothly with huggingface (short of asking them to add an option to return plain dicts).

thomasjpfan:
We can check for `Mapping` instead:

```python
from transformers import AutoTokenizer
from collections.abc import Mapping

tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased")
tokens = tokenizer(["This is a sentence"])

print(isinstance(tokens, Mapping))
# True
```

[...]

PR #840, opened 2022-03-06
Facilitate integration with huggingface

BenjaminBossan:
Solves #838

## Description

The Huggingface transformers library (or, more precisely, the tokenizers
library) currently is possibly the most popular library for state of the
art NLP models. Therefore, it would be good if skorch works well with
transformers.

One of the key steps is tokenization. For this, tokenizers implements
their own classes that return a special data type, which is currently
not understood by skorch. This PR makes it so that skorch understands
and thus correctly works with that data type. Without this adjustment,
users would need to write some (light) wrapper code to make skorch work
with these tokenizers.

## Implementation

The problem with this special data type, `BatchEncoding`, is that it's a
`Mapping` but not a `dict`. skorch already deals with `dict`s but since
`Mapping`s are not instances of `dict`s, skorch does not recognize the
`BatchEncoding` correctly. Therefore, some `isinstance` checks had to be
adjusted.

The second difficulty with `BatchEncoding` is that its values are lists
of lists (e.g. lists of lists of ints for the token ids). This clashes
with skorch, which, when encountering a list, believes that the list
contains the actual values. Therefore, when we determine the length of a
`BatchEncoding`, we don't get the length of the outer list, which is the
desired value (the sample size), but we get the length of the inner
list (the sequence length). This in turn leads to an error with our
`Dataset`, which believes that lengths are inconsistent.

To work around this issue, we could change how the `get_len`
implementation deals with lists of values. However, this would introduce
a backwards incompatibility. Thus I took a different, somewhat
inefficient approach: I call the `convert_to_tensors` method on
`BatchEncoding`, which converts the values from lists of lists to (padded)
tensors. For those, `get_len` returns the correct results.

This conversion is obviously inefficient. Hopefully, however, this makes
no big difference overall because the `Dataset` is initialized only once
per fit call.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -70,11 +70,11 @@ def _len(x):
 
 
 def get_len(data):
-    if hasattr(data, 'convert_to_tensors'):
-        # this might be expensive but huggingface BatchEncodings are lists of
-        # lists and thus their length would be determined incorrectly otherwise,
-        # returning the sequence length instead of the number of samples.
-        data = data.convert_to_tensors('pt')
+    if isinstance(data, Mapping) and (data.get('input_ids') is not None):
+        # Special casing Huggingface BatchEncodings because they are lists of
+        # lists and thus their length would be determined incorrectly, returning
+        # the sequence length instead of the number of samples.
+        return len(data['input_ids'])
     lens = [_apply_to_data(data, _len, unpack_dict=True)]
     lens = list(flatten(lens))
     len_set = set(lens)
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -1,5 +1,6 @@
 """Contains custom skorch Dataset and ValidSplit."""
 import warnings
+from collections.abc import Mapping
 from functools import partial
 from numbers import Number
 
@@ -40,7 +41,7 @@ def _apply_to_data(data, func, unpack_dict=False):
     """
     apply_ = partial(_apply_to_data, func=func, unpack_dict=unpack_dict)
 
-    if isinstance(data, dict):
+    if isinstance(data, Mapping):
         if unpack_dict:
             return [apply_(v) for v in data.values()]
         return {k: apply_(v) for k, v in data.items()}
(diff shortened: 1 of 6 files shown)


skorch/dataset.py
```diff
@@ -69,6 +70,11 @@ def _len(x):
 
 
 def get_len(data):
+    if hasattr(data, 'convert_to_tensors'):
+        # this might be expensive but huggingface BatchEncodings are lists of
+        # lists and thus their length would be determined incorrectly otherwise,
+        # returning the sequence length instead of the number of samples.
+        data = data.convert_to_tensors('pt')
```
thomasjpfan, line 77: Since we are special casing for `BatchEncodings` anyways, can we have custom logic get the outer list length just for `BatchEncodings`?

BenjaminBossan: Yes, good idea, I changed the code not to do the conversion but instead just return the length of the outer list. Also, instead of checking if the `convert_to_tensors` method exists, I now check if we deal with a `Mapping` that has the `input_ids` key.

skorch/utils.py
```diff
@@ -172,6 +175,10 @@ def to_device(X, device):
     if isinstance(X, dict):
         return {key: to_device(val, device) for key, val in X.items()}
 
+    if isinstance(X, Mapping):
+        # dict-like but not a dict
+        return type(X)({key: to_device(val, device) for key, val in X.items()})
```
thomasjpfan, line 180: Can this replace the `if isinstance(X, dict)` call above?

thomasjpfan (approved):
Minor comment, otherwise LGTM

skorch/utils.py
BenjaminBossan: I removed it.

BenjaminBossan: Is this now ready to be merged?

skorch/tests/test_dataset.py
```diff
@@ -381,6 +390,78 @@ def test_fit_predict_proba(self, net, data):
...
+        )
+
+    def test_fit_predict_proba(self, net, data):
+        X, y = data
+        net.fit(X, y)
+        y_proba = net.predict_proba(X)
+        assert np.allclose(y_proba.sum(1), 1)
+
+        train_losses = net.history[:, 'train_loss']
+        assert train_losses[0] > 5 * train_losses[-1]
```
ottonemo, line 462:
```suggestion
        # make sure the network trained successfully with an arbitrary wide margin
        assert train_losses[0] > 5 * train_losses[-1]
```

skorch/utils.py
```diff
@@ -90,7 +90,10 @@ def to_tensor(X, device, accept_sparse=False):
 
     if is_torch_data_type(X):
         return to_device(X, device)
-    if isinstance(X, dict):
+    if hasattr(X, 'convert_to_tensors'):
+        # transformers BatchEncoding
```
ottonemo, line 94:
```suggestion
        # huggingface transformers BatchEncoding
```

ottonemo (approved):

PR #842, opened 2022-03-09
Unwrap accelerated model on train end

BenjaminBossan:
It is apparently recommended to "unwrap" the model after training ends:

https://github.com/huggingface/accelerate/blob/6ffab178accebda485295bddf8eb6bf436ff698f/src/accelerate/accelerator.py#L542-L551

https://github.com/huggingface/accelerate/blob/e549cea65ced232131d2d028717b6a0057274add/docs/source/quicktour.rst#savingloading-a-model


thomasjpfan (approved):
LGTM

PR #846, opened 2022-03-20
Update to PyTorch 1.11.0

BenjaminBossan:
- Drop 1.7.1
- Update 1.10.0/1.10.1 to 1.10.2

Moreover, change all pip install instructions from `pip install` to
`python -m pip install`, which is the recommended way.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         python_version: ['3.6', '3.7', '3.8', '3.9']
-        torch_version: ['1.7.1+cpu', '1.8.1+cpu', '1.9.1+cpu', '1.10.1+cpu']
+        torch_version: ['1.8.2+cpu', '1.9.1+cpu', '1.10.2+cpu', '1.11.0+cpu']
         os: [ubuntu-latest]
         exclude:
           - python_version: '3.9'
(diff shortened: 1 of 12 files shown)


BenjaminBossan:
Looking at the [release notes](https://github.com/pytorch/pytorch/releases), I don't see any necessary changes on the skorch side.

We should possibly take a look at the new [TorchData](https://pytorch.org/data/beta/index.html) to see if there is anything required on the skorch side to support it or facilitate its use.

BenjaminBossan:
Hmm, not sure why we get this error:

> ERROR: Could not find a version that satisfies the requirement torch==1.11.0+cpu

1.11.0 for CPU seems to be listed on https://download.pytorch.org/whl/torch_stable.html :/

thomasjpfan:
> 1.11.0 for CPU seems to be listed on https://download.pytorch.org/whl/torch_stable.html :/

Looking through the METADATA of the wheel, `1.11.0` does not support Python 3.6

BenjaminBossan:
> Looking through the METADATA of the wheel, `1.11.0` does not support Python 3.6

Yes, that was it, thanks!

BenjaminBossan:
@thomasjpfan Is this ready to be merged?

thomasjpfan (approved):
LGTM

PR #847, opened 2022-03-24
Trim net for prediction

BenjaminBossan:
## Description

After finishing training, a net has many attributes that are not needed
anymore when only inference is performed. A user might want to get rid
of these attributes to keep the size of the net small. E.g., depending
on the chosen optimizer, it can be quite big.

(In addition, there is the advantage that the less attributes exist, the
more likely a net can be unpickled that used a different version of
skorch, PyTorch, etc.).

This commit adds a convenience function, `trim_for_prediction`, that takes
care of removing all attributes that are no longer required.

## Implementation

In addition to removing unneeded attributes, make sure to clear
callbacks (which, at least at the moment, are used exclusively during
training), the history, and the `train_split`. `training` is set to False.

Also, set a special attribute after trimming. When the net is
initialized/fitted, check if the net is trimmed and raise a useful error
message for the user. This check does not require the attribute to be
set, to prevent possible compatibility issues. For this, a new training
readiness check and exception are introduced.

If someone can think of a better name for `trim_for_prediction`, feel
free to suggest.

An open question: Do we want to provide this functionality as an
argument for `TrainEndCheckpoint`? If so, just as True/False or as a
callable?


BenjaminBossan:
Hmm, for some reason CI is not triggered. The branch also is not listed on github actions. Any ideas?

thomasjpfan (commented):
It would be interesting to take advantage of [inference mode](https://pytorch.org/docs/stable/generated/torch.inference_mode.html), but not a blocker for this PR.

> Do we want to provide this functionality as an
argument for TrainEndCheckpoint

I do not think trimming should be a part of training at all. I prefer the user to call `net.trim_for_predictions()` directly.

skorch/utils.py
```diff
@@ -704,3 +704,73 @@ def _check_f_arguments(caller_name, **kwargs):
...
+      Whether to trim the history. The history contains metrics from training
+      and is thus typically not needed for prediction,
+
+    """
+    # pylint: disable=protected-access
+    if getattr(net, '_trimmed_for_prediction', False):
+        return
+
+    net.check_is_fitted()
+    net._trimmed_for_prediction = True
```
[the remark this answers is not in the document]
BenjaminBossan: Agree, I changed this. Consequently, I also moved the tests from `test_utils.py` to `test_net.py` but I left them inside their own class.

skorch/utils.py
```diff
@@ -704,3 +704,73 @@ def _check_f_arguments(caller_name, **kwargs):
...
+        net_original = deepcopy(net)
+        trim_for_prediction(net)
+        net.predict(X)
+
+    Parameters
+    ----------
+    net : NeuralNetClassifier
+      The net to be trimmed. Trimming is performed inplace.
+
+    trim_criterion : bool (default=True)
```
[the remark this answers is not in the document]
BenjaminBossan: Okay, I removed the options.

skorch/net.py
```diff
@@ -1255,6 +1255,60 @@ def check_is_fitted(self, attributes=None, *args, **kwargs):
...
+        self.check_is_fitted()
+        # pylint: disable=attribute-defined-outside-init
+        self._trimmed_for_prediction = True
+        self._set_training(False)
+
+        if isinstance(self.callbacks, list):
+            self.callbacks.clear()
+        self.callbacks_.clear()
+
+        # could hold a reference to y
```
thomasjpfan, line 1300: Where can `train_split` get a reference to y?

BenjaminBossan: Yes, I was mistaken. There are possible splits that take `y` when `.split` is called but not during `__init__`. Therefore, in theory, they should not hold a reference to `y`.

That said, I think there is no harm in removing `train_split` as well, is there?

skorch/net.py
```diff
@@ -1241,6 +1255,60 @@ def check_is_fitted(self, attributes=None, *args, **kwargs):
...
+        self.train_split = None
+
+        self.history.clear()
+
+        attrs_to_trim = self._optimizers[:] + self._criteria[:]
+
+        for name in attrs_to_trim:
+            setattr(self, name + '_', None)
```
thomasjpfan, line 1308: Should we `del` instead?

```suggestion
            delattr(self, name + '_', None)
```

skorch/net.py
thomasjpfan: I am okay with removing `train_split`. If we went to keep removing "training specific attributes", we can also remove `iterator_train`.

skorch/net.py
```diff
@@ -1241,6 +1255,60 @@ def check_is_fitted(self, attributes=None, *args, **kwargs):
...
+        # pylint: disable=attribute-defined-outside-init
+        self._trimmed_for_prediction = True
+        self._set_training(False)
+
+        if isinstance(self.callbacks, list):
+            self.callbacks.clear()
+        self.callbacks_.clear()
+
+        # could hold a reference to y
+        self.train_split = None
```
thomasjpfan, line 1301: Same here regarding `del`.

thomasjpfan (commented):
In practice, how much memory does this save?  I'm guessing most of the memory is from the history and callbacks. I do not think the optimizers or criterions hold much state (if any at all).

skorch/net.py
BenjaminBossan: That's a good question. I had a weak preference for setting the attributes to `None`, because I thought that this would reduce the probability of some code breaking unnecessarily. Is there a reason to favor deleting?

BenjaminBossan:
@thomasjpfan I made the change of additionally removing the `iterator_train` and I removed the comment about references to `y` in `train_split`. Regarding the question about deleting vs setting to None, see my reply.

> In practice, how much memory does this save? I'm guessing most of the memory is from the history and callbacks. I do not think the optimizers or criterions hold much state (if any at all).

It depends. The optimizer can be quite big. E.g. when using Adam or other adaptive optimizers, since they keep track of the weight updates made so far, they can easily double the size of the net.

Regarding the criterion, it will most frequently hold little to no state. But if we assume someone builds, say, a GAN, and uses the criterion as the discriminator, it could be quite big. (This was also the use case I thought of when I added the option not to trim the criterion.)

skorch/net.py
thomasjpfan: Let's keep it at `None`. I do not have a strong preference either way.

thomasjpfan (approved):
LGTM

PR #850, opened 2022-04-11
Corrected Import for Python >=3.3

terminator-ger:
fixed import from collections


BenjaminBossan:
Good catch, thanks for the fix!

Issue #830, opened 2022-01-16
choice of network output and NeuralNet criterion


FlorinAndrei:
I use PyTorch for classification with one-hot-encoded labels. I'm looking at this page:

https://skorch.readthedocs.io/en/stable/user/neuralnet.html#criterion

It says:

> When you use the NeuralNetClassifier, the criterion is set to PyTorch NLLLoss by default. Furthermore, if you don’t change it loss to another criterion, NeuralNetClassifier assumes that the module returns probabilities and will automatically apply a logarithm on them (which is what NLLLoss expects).

The quickstart example in the docs suggests using F.softmax() on the output of the classifier network, and uses the default value for the criterion in NeuralNetClassifier() - which seems to be NLLLoss().

When training PyTorch with manually-selected hyperparameters, the combinations that are most often used are:

- linear output + nn.CrossEntropyLoss
- log_softmax output + nn.NLLLoss

What if I want to write a model just once, and then pass it without changes to Skorch for hyperparameter tuning, and then again train it on a data set, without any extra transformations of the data, the best option seem to be to just use a linear output with CrossEntropyLoss. If I have softmax on the output, then in regular training I'd have to log the output, then pass it to NLLLoss(), which seems like an unnecessary extra step.

Why not just recommend using a linear output by default, and have the default criterion in Skorch be CrossEntropyLoss?

BenjaminBossan:
> When training PyTorch with manually-selected hyperparameters, the combinations that are most often used are:
> 
>     * linear output + nn.CrossEntropyLoss
> 
>     * log_softmax output + nn.NLLLoss

> Why not just recommend using a linear output by default, and have the default criterion in Skorch be CrossEntropyLoss?

These are some valid points. We have the defaults as they currently are for historical reasons. When skorch was released (July 2017), there were not really any established pytorch standards, but there were standards from other libraries like nolearn and keras. With hindsight, we might have chosen the standard you suggested.

At this point, changing such a fundamental default might not be worth it, it's hard for us to know how much code would be affected. And it's really not hard to make the change you suggested with an extra line of code. That's why we decided to keep it as is.

FlorinAndrei:
I see. In that case, could you add some notes to the Skorch documentation - basically saying "here are the current best practices" etc.

The Quickstart page recommends a softmax() output, but that's not the best choice given the default criterion used by Skorch.

https://skorch.readthedocs.io/en/stable/user/quickstart.html

[...]

PR #852, opened 2022-04-24
Mention that softmax might lead to num instability

BenjaminBossan:
Solves #830

By default, skorch uses softmax + `NLLLoss` for classification. This can
lead to numerical instability issues (when logits have large magnitude).
Therefore, mention in the docs how to prevent that issue.

### DIFF
--- a/docs/user/quickstart.rst
+++ b/docs/user/quickstart.rst
@@ -14,7 +14,6 @@ it on a toy classification dataset using skorch
     import numpy as np
     from sklearn.datasets import make_classification
     from torch import nn
-    import torch.nn.functional as F
 
     from skorch import NeuralNetClassifier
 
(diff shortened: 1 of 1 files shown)


docs/user/quickstart.rst
```diff
@@ -52,6 +52,12 @@ it on a toy classification dataset using skorch
     net.fit(X, y)
     y_proba = net.predict_proba(X)
 
+.. note::
+
+    The use of ``softmax`` might result in numerical instability. If you have
```
thomasjpfan, line 57: In general, I think using `CrossEntropyLoss` is the recommended approach. What do you think about updating the example to use `CrossEntropyLoss` and removing the `softmax`?

BenjaminBossan: As always, you come up with the best suggestions. I changed the example accordingly.

BenjaminBossan:
@thomasjpfan could you please take another look?

thomasjpfan (approved):
LGTM

PR #858, opened 2022-05-21
Bugfix: sklearn 1.1 and SliceDataset

BenjaminBossan:
# Sklearn 1.1 issue

The error message when using `set_params` with an invalid key has been
altered to contain quote marks. Now testing message without or with.

# `SliceDataset` issue

Using `SliceDataset` could result in an error because of this line:

https://github.com/skorch-dev/skorch/blob/090ece088803244659c6eb16876e84bf784967b7/skorch/classifier.py#L119

This would fail when y comes from a `SliceDataset`. However, the error was masked
because `GridSearchCV` swallows errors by default now. Therefore, we now run these
tests with `error='raise'` to surface the bug.

## Implementation

Regarding the bugfix, it is a bit clumsy because we need to make
`to_numpy` work with `SliceDataset`, but we cannot check

`'isinstance(X, SliceDataset)'`

because we don't want to import the helper class. Therefore, we now
check this indirectly by looking at attributes.

Furthermore, `SliceDataset` now also works with torch tensors as X and y,
not only numpy arrays. Tests have been extended to cover this.

### DIFF
--- a/skorch/tests/callbacks/test_all.py
+++ b/skorch/tests/callbacks/test_all.py
@@ -46,12 +46,14 @@ class TestAllCallbacks:
     def test_set_params_with_unknown_key_raises(self, base_cls):
         with pytest.raises(ValueError) as exc:
             base_cls().set_params(foo=123)
-
-        msg_start = (
-            "Invalid parameter foo for estimator <skorch.callbacks.base.Callback")
-        msg_end = (
-            "Check the list of available parameters with "
-            "`estimator.get_params().keys()`.")
         msg = exc.value.args[0]
-        assert msg.startswith(msg_start)
-        assert msg.endswith(msg_end)
+
+        # message contains "'" around variable name starting from sklearn 1.11
+        assert (
+            msg.startswith(
+                "Invalid parameter foo for estimator <skorch.callbacks.base.Callback"
+            )
+            or msg.startswith(
+                "Invalid parameter 'foo' for estimator <skorch.callbacks.base.Callback"
+            )
+        )
--- a/skorch/tests/test_net.py
+++ b/skorch/tests/test_net.py
@@ -2202,9 +2202,12 @@ class TestNeuralNet:
         with pytest.raises(ValueError) as exc:
             net.set_params(foo=123)
 
-        # TODO: check error message more precisely, depending on what
-        # the intended message should be from sklearn side
-        assert exc.value.args[0].startswith('Invalid parameter foo for')
+        msg = exc.value.args[0]
+        # message contains "'" around variable name starting from sklearn 1.11
+        assert (
+            msg.startswith("Invalid parameter foo for")
+            or msg.startswith("Invalid parameter 'foo' for")
+        )
 
     @pytest.fixture()
     def sequence_module_cls(self):
--- a/skorch/utils.py
+++ b/skorch/utils.py
@@ -135,6 +135,13 @@ def to_numpy(X):
     if isinstance(X, (tuple, list)):
         return type(X)(to_numpy(x) for x in X)
 
+    if hasattr(X, 'dataset') and hasattr(X, 'idx') and hasattr(X, 'indices'):
+        # We probably deal with a SliceDataset. Cannot use isinstance because we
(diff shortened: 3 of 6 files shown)


skorch/tests/callbacks/test_all.py
```diff
@@ -46,12 +46,14 @@ def test_on_x_methods_have_kwargs(self, callbacks, on_x_methods):
...
-        msg_start = (
-            "Invalid parameter foo for estimator <skorch.callbacks.base.Callback")
-        msg_end = (
-            "Check the list of available parameters with "
-            "`estimator.get_params().keys()`.")
         msg = exc.value.args[0]
-        assert msg.startswith(msg_start)
-        assert msg.endswith(msg_end)
+
+        # message contains "'" around variable name starting from sklearn 1.11
```
thomasjpfan, line 51:
```suggestion
        # message contains "'" around variable name starting from sklearn 1.1
```

skorch/tests/test_net.py
```diff
@@ -2202,9 +2202,12 @@ def test_set_params_with_unknown_key_raises(self, net):
         with pytest.raises(ValueError) as exc:
             net.set_params(foo=123)
 
-        # TODO: check error message more precisely, depending on what
-        # the intended message should be from sklearn side
-        assert exc.value.args[0].startswith('Invalid parameter foo for')
+        msg = exc.value.args[0]
+        # message contains "'" around variable name starting from sklearn 1.11
```
thomasjpfan, line 2206:
```suggestion
        # message contains "'" around variable name starting from sklearn 1.1
```

skorch/utils.py
```diff
@@ -135,6 +135,13 @@ def to_numpy(X):
     if isinstance(X, (tuple, list)):
         return type(X)(to_numpy(x) for x in X)
 
+    if hasattr(X, 'dataset') and hasattr(X, 'idx') and hasattr(X, 'indices'):
+        # We probably deal with a SliceDataset. Cannot use isinstance because we
+        # don't want to depend on helper.py.
+        if np.isscalar(X[0]):
+            return np.array([X[i] for i in range(len(X))])
+        return np.array([to_numpy(X[i]) for i in range(len(X))])
```
thomasjpfan, line 143: Nit:

```suggestion
        Xs = [X[i] for i in range(len(X))]
        if np.isscalar(out[0]):
            return np.array(Xs)
        return np.array([to_numpy(x) for x in Xs])
```

thomasjpfan (commented):
I'm okay with the workaround with SliceDataset.

skorch/utils.py
BenjaminBossan: Re sklearn version: I was getting ahead of myself a bit :)

Re this suggestion: Could you explain why it is better? (Also, there is no `out` variable, is it referring to `X` or `Xs`?) It seems like with your suggestion, we iterate over `X` a second time if we don't deal with a scalar, is it not potentially more expensive?

Ideally, we could have a version that doesn't need to test the first element at all, but I'm not sure it's achievable.

BenjaminBossan: > I wanted to guard against the second X[0] access, because SliceDataset.transform can do some computation. I have not benchmarked it, but my sense is that the numpy conversation will take up majority of the compute compared to iterating X a second time.

Yeah, it's a difficult call. In the best case, a numpy conversion is very cheap because pytorch and numpy share a memory layout. But in the worst case, as you mentioned, we pay for one additional `dataset.__getitem__` call. I think the more cautious solution is to prevent the worst case, i.e. avoid an additional `__getitem__` but iterate through the list once more.

> We can have to_numpy support scalars and return np.asarray(scalar) if the input is is a scalar?

I think this could be a dangerous solution performance-wise, as we would be checking each item of the list. As is, we only check the first item and then assume that the subsequent items have the same type.

> Side note: A slightly nicer solution would be to define SliceDataset.__array__, so it knows how to turn itself into a ndarray.

I like that, I moved the conversion code into `SliceDataset.__array__`. Of course, performance is the same this way or that, but it looks cleaner.

thomasjpfan (approved):

Issue #859, opened 2022-05-23
Make `NotInitializedError` a subclass of `NotFittedError`


timokau:
Hi,

currently calling `predict` on an uninitialized estimator throws a `NotInitializedError` instead of scikit-learn's `NotFittedError`. This is problematic in cases where we want to be able to handle different kinds of estimators while providing some kind of fallback behavior. Currently we need

```
try:
    prediction = estimator.predict(X)
except (NotFittedError, NotInitializedError):
    # some fallback behavior
```

I think it would be better if `NotInitializedError` was a subclass of `NotFittedError`. In that case we could treat an skorch estimator like a regular scikit-learn estimator and simply catch `NotFittedError`.

Would a PR in that direction be welcome?

timokau:
Great :) See #860.

PR #860, opened 2022-05-24
Make NotInitializedError compatible with sklearn

timokau:
This ensures that a user can catch the sklearn `NotFittedError`, just as
they can with regular sklearn estimators. As discussed in [1].

[1] https://github.com/skorch-dev/skorch/issues/859


timokau:
The test failure seems unrelated as far as I can tell.

BenjaminBossan:
Thanks for the PR @timokau. I'll take a look later.

> The test failure seems unrelated as far as I can tell.

Yes, related to #858

PR #861, opened 2022-06-06
Check Python 3.10 in CI

BenjaminBossan:
No longer check Python 3.6 because it's been unsupported since the end of last year. Instead, add Python 3.10 to the CI. Unfortunately, the only PyTorch version supporting 3.10 right now is 1.11 (issue #66424), so all other versions are not being checked.

While working on this, an issue with wandb + protobuf came up, so I had to pin some versions there.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -10,6 +10,7 @@ numpydoc
 openpyxl
 pandas
 pillow
+protobuf>=3.12.0,<4.0dev
 pylint
 pytest>=3.4
 pytest-cov
@@ -18,4 +19,4 @@ sphinx
 sphinx_rtd_theme
 tensorboard>=1.14.0
 transformers
-wandb<0.10.29
+wandb>=0.12.17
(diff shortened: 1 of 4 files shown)


requirements-dev.txt
```diff
@@ -10,6 +10,7 @@ numpydoc
 openpyxl
 pandas
 pillow
+protobuf>=3.12.0,<4.0dev
```
thomasjpfan, line 13: I've been seeing the protobuf change break a few things now.

thomasjpfan (approved):
LGTM

requirements-dev.txt
BenjaminBossan: It'll probably take some time for libraries to work with version 4. Hopefully, once that has been achieved, we don't forget about this restriction.

BenjaminBossan:
@thomasjpfan I bumped the minimum Python version in the setup.py to 3.7. Technically, we could still support 3.6 but since it's EOL, I would suggest we explicitly don't anymore. WDYT?

Also, we could start using Python 3.7 features if we don't officially support 3.6 anymore.

thomasjpfan:
> I would suggest we explicitly don't anymore. WDYT?

Bumping the minimum version to 3.7 makes sense to me.

PR #867, opened 2022-07-04
Docs: Fix issues callback docs

BenjaminBossan:
I fixed a couple of minor issues I found in the docs

- Added/removed some blank lines for consistency
- Some entries like 'event_' were interpreted as links, so I added code
  markers around them.
- Added usage example to `TensorBoard` docstring
- Fixed a copy paste error in `WandbLogger`
- Some links were not parsed correctly, so I used different syntax
- Fixed issue in code example of `SacredLogger`

Additionally, I removed an outdated comment in `probabilistic.py` docstring.

What I would like to know: Is it possible to have blank lines inside
example code blocks without starting a completely new block? Some
examples are now very cramped and I would like to add some blank lines,
but don't know how.

Also, it looks like in the official docs, the API docs for `probabilistic.py` are empty:

https://skorch.readthedocs.io/en/latest/probabilistic.html

I cannot reproduce this error locally :(

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -756,7 +756,9 @@ class SacredLogger(Callback):
 
     To use this logger, you first have to install Sacred:
 
-    ``$ python -m pip install sacred``
+    .. code-block:: bash
+
+        python -m pip install sacred
 
     You might also install pymongo to use a mongodb backend. See the `upstream
     documentation <https://github.com/IDSIA/sacred#installing>`_ for more
@@ -793,7 +795,10 @@ class SacredLogger(Callback):
     ...     net.fit(X, y)
 
     Then call this from the command line, e.g. like this:
-    ``python sacred-script.py with max_epochs=15``
+
+    .. code-block:: bash
+
+        python sacred-script.py with max_epochs=15
 
     You can also change other options on the command line and optionally
     specify a backend.
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -148,10 +148,9 @@ class NeptuneLogger(Callback):
       manager.
 
     keys_ignored : str or list of str (default=None)
-      Key or list of keys that should not be logged to
-      Neptune. Note that in addition to the keys provided by the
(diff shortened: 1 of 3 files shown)


skorch/callbacks/logging.py
```diff
@@ -747,32 +756,30 @@ class SacredLogger(Callback):
 
     To use this logger, you first have to install Sacred:
 
-    $ python -m pip install sacred
+    ``$ python -m pip install sacred``
```
thomasjpfan, line 759: I think this works for Sphinx?

```suggestion
    .. code-block::
      $ python -m pip install sacred
```

skorch/callbacks/logging.py
```diff
@@ -747,32 +756,30 @@ class SacredLogger(Callback):
...
+    --------
+    >>> # contents of sacred-experiment.py
     >>> import numpy as np
     >>> from sacred import Experiment
     >>> from sklearn.datasets import make_classification
     >>> from skorch.callbacks.logging import SacredLogger
     >>> from skorch.callbacks.scoring import EpochScoring
     >>> from skorch import NeuralNetClassifier
     >>> from skorch.toy import make_classifier
-
```
thomasjpfan, line 765: One way to split this up is to add a comment describing the code.

XREF: https://numpydoc.readthedocs.io/en/latest/format.html#examples

BenjaminBossan: Yes, but I wanted to avoid that. My goal would be to just have a block of code with some blank lines:

```python
from skorch.toy import make_classifier

ex = Experiment()

@ex.config
def my_config():
    max_epochs = 20
    lr = 0.01

X, y = make_classification()
```
 
Adding comments seems unnecessary to me in this case.

skorch/callbacks/logging.py
BenjaminBossan: I used code blocks for the 2 shell commands in this docstring.

skorch/callbacks/logging.py
thomasjpfan: Ah I see. I do not think there is a great way to do this.

PR #868, opened 2022-07-05
Make binary classifier work with BCELoss

BenjaminBossan:
Until now, `NeuralNetBinaryClassifier` assumes that `BCEWithLogitLoss` is used. Now, `BCELoss` works as well.

Note: Unless there is a very good reason not to, it's still recommended to use `NeuralNetClassifier`, even when dealing with a binary classification problem.

### DIFF
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -88,7 +88,7 @@ class TestNeuralNet:
 
     # classifier-specific test
     def test_takes_no_log_without_nllloss(self, net_cls, module_cls, data):
-        net = net_cls(module_cls, criterion=nn.BCELoss, max_epochs=1)
+        net = net_cls(module_cls, criterion=nn.CrossEntropyLoss, max_epochs=1)
         net.initialize()
 
         mock_loss = Mock(side_effect=nn.NLLLoss())
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -255,6 +255,11 @@ class TestNeuralNetBinaryClassifier:
         y_pred_proba = net.predict_proba(X)
         assert y_pred_proba.shape == (X.shape[0], 2)
 
+        # the tests below check that we don't accidentally apply sigmoid twice,
+        # which would result in all probas being withiin [0.2689, 0.7311].
+        assert (y_pred_proba < 0.26).any()
+        assert (y_pred_proba > 0.74).any()
+
         y_pred_exp = (y_pred_proba[:, 1] > threshold).astype('uint8')
 
         y_pred_actual = net.predict(X)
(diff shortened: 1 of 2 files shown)


skorch/utils.py
```diff
@@ -632,7 +633,7 @@ def _infer_predict_nonlinearity(net):
     if isinstance(criterion, CrossEntropyLoss):
         return partial(torch.softmax, dim=-1)
 
-    if isinstance(criterion, BCEWithLogitsLoss):
+    if isinstance(criterion, (BCEWithLogitsLoss, BCELoss)):
```
thomasjpfan, line 636: In the test, `module__output_nonlin=torch.nn.Sigmoid()`, which means the output already has sigmoid applied. I think using this non-linearity would apply another sigmoid.

If we assume that the module already does sigmoid then the module already outputs probabilities. In that case, I think the non-linearity should only be `torch.stack((1 - prob, prob), 1)` part of `_sigmoid_then_2d`.

BenjaminBossan: Very good catch, thanks. I fixed the bug by not applying the sigmoid in the postprocessing step in `utils.py`. We still need it as output nonlinearity because that is what `BCELoss` expects.

To catch this kind of bug in the future, I extended the unit tests. When sigmoid is applied twice, all probabilities would be in the range of [sigmoid(-1), sigmoid(1)]. I now check that at least some returned probabilities are outside that range.

skorch/tests/test_classifier.py
```diff
@@ -255,6 +255,11 @@ def test_predict_predict_proba(self, net, data, threshold):
         y_pred_proba = net.predict_proba(X)
         assert y_pred_proba.shape == (X.shape[0], 2)
 
+        # the tests below check that we don't accidentally apply sigmoid twice,
+        # which would result in all probas being withiin [0.2689, 0.7311].
+        assert (y_pred_proba < 0.26).any()
```
thomasjpfan, line 260: Is this suppose to be `expit(0) = 0.5` on the lower bound? ([scipy.special.expit](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expit.html))

```suggestion
        assert (y_pred_proba < expit(0)).any()
```

BenjaminBossan: Indeed, we are comparing to `expit(1)` and `expit(-1)`. The loewr bound is not `expit(0)` because we create the probabilities as `torch.stack((1 - prob, prob), 1)`, so the lowest value that could be achieved by applying sigmoid twice is `1 - expit(1)`, which is equal to `expit(-1)`.

thomasjpfan: Can you add this information in a code comment?

BenjaminBossan:
Note: I suspect CI is failing because of the new `accelerate` release, I'll investigate.

thomasjpfan (approved):
LGTM

The [failing tests](https://github.com/skorch-dev/skorch/runs/7449764358?check_suite_focus=true) do not look related to this PR, but they look concerning.

PR #871, opened 2022-07-14
Small fixes to SliceDict and SliceDataset

### LINKED ISSUE #816: Slicing SliceDataset removes transform
Hi,
If one slices a `SliceDataset` (as done automatically within scikit-learns `cross_validate`) which has a `transform` , the `transform` is not applied anymore. Below is a minimal working example to demonstrate this:
```
from skorch.helper import SliceDataset

class Dataset():
    def __getitem__(self, i):
        return i, None
    
    def __len__(self):
        return 10

ds = Dataset()
slice_ds = SliceDataset(ds, idx=0)
slice_ds.transform = lambda x: [x]
train_i = [0, 1]
# This is expected to be [0] but is 0
slice_ds[train_i][0]
```

Checking the return type in `__getitem__` and setting the `transform` when it is a `SliceDataset` could be a fix, although I do not know whether it brakes anything.
```
class CustomSliceDataset(SliceDataset):
    def __init__(self, *args, **kwargs):
    
    def __getitem__(self, i):
        return_value = super().__getitem__(i)
        if isinstance(return_value, SliceDataset):
            return_value.transform = self.transform
        return return_value
```

BenjaminBossan:
Fixes #816

- Returning new instances now return the type of self instead of
  hard-coding the class. This is necessary when users subclass these
  classes, as otherwise, the subclass would return instances of the
  parent class.
- Use super() without arguments.

### DIFF
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -54,7 +54,7 @@ class SliceDict(dict):
         else:
             self._len = lengths[0]
 
-        super(SliceDict, self).__init__(**kwargs)
+        super().__init__(**kwargs)
 
     def __len__(self):
         return self._len
@@ -66,8 +66,8 @@ class SliceDict(dict):
             # lengths and shapes.
             raise ValueError("SliceDict cannot be indexed by integers.")
         if isinstance(sl, str):
-            return super(SliceDict, self).__getitem__(sl)
-        return SliceDict(**{k: v[sl] for k, v in self.items()})
+            return super().__getitem__(sl)
+        return type(self)(**{k: v[sl] for k, v in self.items()})
 
     def __setitem__(self, key, value):
         if not isinstance(key, str):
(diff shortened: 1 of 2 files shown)


skorch/helper.py
```diff
@@ -245,7 +246,7 @@ def __getitem__(self, i):
             if i.dtype == np.bool:
                 i = np.flatnonzero(i)
 
-        return SliceDataset(self.dataset, idx=self.idx, indices=self.indices_[i])
+        return cls(self.dataset, idx=self.idx, indices=self.indices_[i])
```
thomasjpfan, line 249: Same here regarding testing (but for `SliceDataset`).

thomasjpfan (commented):
Overall the fix looks good.

skorch/helper.py
BenjaminBossan: Good point, I added the unit tests.

thomasjpfan (approved):
LGTM

PR #873, opened 2022-07-21
Fix broken tests with new accelerate version

BenjaminBossan:
Latest changes in accelerate led to breaking tests. Unfortunately, the
solution is to use a privat method on AcceleratorState but the devs said
that this would be the best solution.

This PR also sets the requirements for accelerate to be v0.11 or higher.
This not only makes our code simpler, but also opens the door to
supporting gradient accumulation with accelerate (which will come in a
separate PR).

### DIFF
--- a/skorch/tests/test_helper.py
+++ b/skorch/tests/test_helper.py
@@ -798,7 +798,13 @@ class TestAccelerate:
         pytest.importorskip('accelerate')
 
         from accelerate import Accelerator
+        from accelerate.state import AcceleratorState
 
+        # We have to use this private method because otherwise, the
+        # AcceleratorState is not completely reset, which results in an error
+        # initializing the Accelerator more than once in the same process.
+        # pylint: disable=protected-access
+        AcceleratorState._reset_state()
         return Accelerator
 
     @pytest.fixture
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
@ottonemo @thomasjpfan CI is currently failing because of the new accelerate version. I hope we can get this PR merged quickly.

skorch/tests/test_helper.py
```diff
@@ -798,7 +798,13 @@ def accelerator_cls(self):
         pytest.importorskip('accelerate')
 
         from accelerate import Accelerator
+        from accelerate.state import AcceleratorState
 
+        # We have to use this private method because otherwise, the
+        # AcceleratorState is not completely reset, which results in an error
+        # initializing the Accelerator more than once in the same process.
+        # pylint: disable=protected-access
+        AcceleratorState._reset_state()
```
thomasjpfan, line 807: Using `AcceleratorState` to influence `Accelerator` is weird from a software design point of view. Is there a public API to reset the `Accelerator`?

Seeing the test changes in https://github.com/huggingface/accelerate/pull/458, I suspect there is no public API.

thomasjpfan (approved):

skorch/tests/test_helper.py
BenjaminBossan: Yes, I agree that it's unfortunate that this has to be done, and the fact that it's necessary in the first place is surprising as a user. The devs told me this is the way to go. Maybe there will be a better solution in the future.

PR #874, opened 2022-07-22
Support for gradient accumulation with accelerate

BenjaminBossan:
The latest accelerate release added gradient accumulation support. This
requires the AccelerateMixin to call the training loop within a context
manager. This is now done. Users can therefore use the gradient
accumulation feature of accelerate.

### DIFF
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -653,6 +653,17 @@ class AccelerateMixin:
 
         return self
 
+    def train_step(self, batch, **fit_params):
+        # Call training step within the accelerator context manager
+        with self.accelerator.accumulate(self.module_):
+            # Why are we passing only module_ here, even though there might be
+            # other modules as well? First of all, there is no possibility to
+            # pass multiple modules. Second, the module_ is only used to
+            # determine if Distributed Data Parallel is being used, not for
+            # anything else. Therefore, passing module_ should be sufficient
+            # most of the time.
+            return super().train_step(batch, **fit_params)
+
     def train_step_single(self, batch, **fit_params):
         self._set_training(True)
         Xi, yi = unpack_data(batch)
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -12,6 +12,7 @@ from sklearn.base import TransformerMixin
 import torch
 
 from skorch.cli import parse_args  # pylint: disable=unused-import
+from skorch.callbacks import LRScheduler
 from skorch.dataset import unpack_data
 from skorch.utils import _make_split
 from skorch.utils import to_numpy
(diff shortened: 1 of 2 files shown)


skorch/helper.py
```diff
@@ -653,6 +653,17 @@ def _initialize_optimizer(self, *args, **kwargs):
...
+    def train_step(self, batch, **fit_params):
+        # Call training step within the accelerator context manager
+        with self.accelerator.accumulate(self.module_):
+            # Why are we passing only module_ here, even though there might be
+            # other modules as well? First of all, there is no possibility to
+            # pass multiple modules. Second, the module_ is only used to
+            # determine if Distributed Data Parallel is being used, not for
+            # anything else. Therefore, passing module_ should be sufficient
+            # most of the time.
+            return super().train_step(batch, **fit_params)
```
thomasjpfan, line 665: From looking at the [accelerate docs on gradient accumlation](https://huggingface.co/docs/accelerate/gradient_accumulation#the-finished-code), the scheduler needs to be prepared by the accelerator as well?

BenjaminBossan: Good catch, thanks for pointing it out. I added code to do that. Unfortunately, I don't think we can identify the usage of schedulers except if users make use of our `LRScheduler` callback. I would assume that they do, but otherwise there may be hidden bugs.

skorch/tests/test_helper.py
```diff
@@ -966,8 +976,69 @@ def train_step_single(self, *args, **kwargs):
             device=None,
             accelerator=accelerator,
             max_epochs=2,
+            callbacks=[('lr_scheduler', LRScheduler)],
```
thomasjpfan, line 979: Most of the usage for lrscheduler is:

```suggestion
            callbacks=[('lr_scheduler',  LRScheduler(ReduceLROnPlateau)],
```

skorch/tests/test_helper.py
```diff
@@ -966,8 +976,69 @@ def train_step_single(self, *args, **kwargs):
...
         # does not raise
         net.fit(X, y)
         net.predict(X)
+
+        # make sure that even after resetting parameters, components are still prepared
+        net.set_params(
+            module__hidden_units=7,
+            lr=0.05,
+            batch_size=33,
+            criterion__reduction='sum',
```
thomasjpfan, line 991: Does the reduction need to be sum here? In practice, I think most use the default 'mean' reduction.

skorch/tests/test_helper.py
BenjaminBossan: We test `ReduceLROnPlateau` below after the `set_params`, so I think we should be good using the default here.

skorch/tests/test_helper.py
BenjaminBossan: I just chose something to change the parameter of each component, in this case the criterion, to a different value, to make sure that even if re-initialized, we're still fine. What exactly that value is doesn't matter.

thomasjpfan (approved):
I ran the new tests on a machine with a Nvidia GPU and everything works.

LGTM

BenjaminBossan:
> I ran the new tests on a machine with a Nvidia GPU and everything works.

Fantastic, thank you @thomasjpfan

PR #875, opened 2022-07-26
Move accelerate and update docs

BenjaminBossan:
After introducing the `skorch/hf.py` module, it made sense to move the
`AccelerateMixin` there from `helper.py`. Of course, this required some code
adjustments, which have been made as well, but there are no functional
changes.

On top of that, I extended some parts of the documentation:

- Created dedicated HF section in docs to put all the HF stuff
- FAQ about gradient accumulation now also points to `AccelerateMixin`
- Mention gradient accumulation in `AccelerateMixin` docstring
- Re-run `Hugging_Face_Finetuning.ipynb` with new accelerate version


thomasjpfan (approved):
This refactoring looks good to me.

PR #876, opened 2022-07-28
Nets now can be loaded even if device=None

BenjaminBossan:
## Description

In #600, we introduced the option to set `device=None`, which means that
skorch should not move any data to any device. However, this setting
introduced a bug when trying to load the net, as that code didn't
explicitly deal with `device=None.` With this PR, the bug is fixed.

## Implementation

If `device=None`, we really have no way of knowing what device to map the
parameters to. The most reasonable thing to do is to use the fallback,
which is 'cpu'.


thomasjpfan (commented):
I am okay with this PR.

Side note: with this fix, models that are pickled with `AccelerateMixin` and trained on GPU, will fall back to CPU when unpickled. I think that is okay behavior, but not obvious.

BenjaminBossan:
> Side note: with this fix, models that are pickled with `AccelerateMixin` and trained on GPU, will fall back to CPU when unpickled. I think that is okay behavior, but not obvious.

@thomasjpfan Yes, absolutely, that's how I discovered the error. I did think about adding a special case for models with an `accelerator` attribute, and I might yet do that. But since the error can occur independently, I thought I'd keep this PR small and work on the other stuff later.

PR #877, opened 2022-07-28
Bugfix/loading set params kwargs without cuda

BenjaminBossan:
This bug occurs when trying to load a net that was trained on CUDA on a
CPU machine. However, it only occurred when there were CUDA-dependent
attribute set via `set_params`. This bug is now fixed.

## Details

The problem started occurring after PR #751, which introduced storing
parameters set via `set_params` in the private attribute `_kwargs`.
Normally, for attributes, we make sure that they can be loaded without
CUDA, but attributes within `_kwargs` were not checked. Thus, loading
those without CUDA failed. Unfortunately, this was not caught by CI
because CI is not CUDA-enabled.

The solution to this problem is that CUDA-dependent attributes are now
removed from `_kwargs` during `__getstate__` and re-added later during
`__setstate__` (a test was introduced for checking that). A little problem
occurring there was that the `cuda_dependent_attributes_` are not part of
`state`. Therefore, I add them to state during `__getstate__` and remove
them later in `__setstate__`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix a bug in `SliceDataset` that prevented it to be used with `to_numpy` (#858)
+- Fix a bug that in some cases could prevent loading a net that was trained with CUDA without CUDA
 
 ## [0.11.0] - 2021-10-11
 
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
After discussion with @ottonemo we agreed on a better solution (making `net._kwargs` a set of key names instead of dict), so this PR needs not be reviewed. I'll still leave it open as a reminder for myself to implement that other solution.

BenjaminBossan:
Superseded by #880

Issue #834, opened 2022-01-30
readthedocs search optimization


ottonemo:
readthedocs analytics says that we have several search results that yield little or no useful results. Let's improvethose:

- gpu (only 2 results): make sure that explanation of `device` parameter mentions `gpu` as well
- gridsearch (0 results): make sure to include the term `gridsearch` in the [meta data](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-metadata) of the `GridSearchCV` page (or similar)
- seed (0 results): think about a FAQ answer
- earlystop (0 results): make sure to include the `earlystop` keyword in the callbacks section

BenjaminBossan:
Should we close this or leave it open as a reminder?

ottonemo:
This was addressed in #879 and #836.

PR #879, opened 2022-08-01
Address #834: Document random seeding

ottonemo:
Since skorch does not manage seeding it might still be helpful to guide users in the right direction and address some of the caveats of model vs. data seeding (i.e. how to fix the validation split or batch shuffling).

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -418,7 +418,7 @@ How can I set the random seed of my model?
 ------------------------------------------
 
 skorch does not provide an unified way for setting the seed of your model.
-You can utilize the `numpy <https://pytorch.org/docs/stable/generated/torch.manual_seed.html>`__
+You can utilize the `numpy <https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html>`__
 and `torch <https://pytorch.org/docs/stable/generated/torch.manual_seed.html>`__
 seeding interfaces to set random seeds before model initialization and will
 get consistent results if you do not employ other libraries that introduce randomness.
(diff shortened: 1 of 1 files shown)


docs/user/FAQ.rst
```diff
@@ -413,6 +413,51 @@ the **greatest** score.
     best_net = grid_searcher.best_estimator_
     print(best_net.score(X, y))
 
+
+How can I set the random seed of my model?
+------------------------------------------
+
+skorch does not provide an unified way for setting the seed of your model.
+You can utilize the `numpy <https://pytorch.org/docs/stable/generated/torch.manual_seed.html>`__
```
BenjaminBossan, line 421: Wrong link for numpy.

docs/user/FAQ.rst
```diff
@@ -413,6 +413,51 @@ the **greatest** score.
...
+and `torch <https://pytorch.org/docs/stable/generated/torch.manual_seed.html>`__
+seeding interfaces to set random seeds before model initialization and will
+get consistent results if you do not employ other libraries that introduce randomness.
```
BenjaminBossan, line 424: Maybe add a word about `torch.cuda.manual_seed`?

BenjaminBossan (changes requested):
Thanks for adding this. Only two minor comments from my side.

docs/user/FAQ.rst
ottonemo: Fixed.

docs/user/FAQ.rst
ottonemo: Yep, since it is called if needed I explained this behavior.

BenjaminBossan (approved):

PR #881, opened 2022-08-02
[WIP] First implementation for HfHubWriter

BenjaminBossan:
This is WIP, more documentation and tests are still missing. This draft should serve to discuss the design.

### DIFF
--- a/skorch/hf.py
+++ b/skorch/hf.py
@@ -1006,3 +1006,115 @@ class AccelerateMixin:
     def on_train_end(self, net, X=None, y=None, **kwargs):
         super().on_train_end(net, X=X, y=y, **kwargs)
         self.module_ = self.accelerator.unwrap_model(self.module_)
+
+
+class HfHubWriter:
+    """Helper class that allows writing data to the Hugging Face Hub.
+
+    Use this, for instance, in combination with checkpoint callbacks such as
+    :class:`skorch.callbacks.training.TrainEndCheckpoint` or
+    :class:`skorch.callbacks.training.Checkpoint` to upload the trained model
+    directly to the Hugging Face Hub instead of storing it locally.
+
+    To use this, it is necessary to install the `Hugging Face Hub library
+    <https://huggingface.co/docs/huggingface_hub/index>`__.
+
+    .. code:: bash
+
+        python -m pip install huggingface_hub
+
+    Parameters
+    ----------
+    hf_api : instance of huggingface_hub.HfApi
+      Pass an instantiated ``huggingface_hub.HfApi`` object here.
+
+    path_in_repo : str
+      The name that the file should have in the repo, e.g. ``my-model.pkl``. If
+      you want each upload to have a different file name, instead of overwriting
+      the file, use a templated name, e.g. ``my-model-{}.pkl``. Then your files
+      will be called ``my-model-1.pkl``, ``my-model-2.pkl``, etc.
+
+    repo_id : str
+      The repository to which the file will be uploaded, for example:
+      ``"username/reponame"``.
+
+    verbose : int (default=0)
+      Control the level of verbosity.
+
+    sink : callable (default=print)
+      The target that the verbose information is sent to. By default, the output
+      is printed to stdout, but the sink could also be a logger or
+      :func:`~skorch.utils.noop`.
+
+    kwargs : dict
+      The remaining arguments are the same as for ``HfApi.upload_file`` (see
+      https://huggingface.co/docs/huggingface_hub/package_reference/hf_api#huggingface_hub.HfApi.upload_file).
+
+    Attributes
+    ----------
+    latest_url_ : str
+      Stores the latest URL that the file has been uploaded to.
+
+    Examples
+    --------
+    >>> from huggingface_hub import create_repo, HfApi
+    >>> model_name = 'my-skorch-model.pkl'
+    >>> repo_name = 'my-user-name/my-repo-name'
+    >>> token = 'my-secret-token'
+    >>> # you can create a new repo like this:
+    >>> create_repo(repo_name, private=True, token=token, exist_ok=True)
+    >>> hf_api = HfApi()
+    >>> hub_writer = HfHubWriter(
+    ...     hf_api,
+    ...     path_in_repo=model_name,
+    ...     repo_id=repo_namke,
+    ...     token=token,
+    ...     verbose=1,
+    >>> )
+    >>> checkpoint = TrainEndCheckpoint(f_pickle=hub_writer)
+    >>> net = NeuralNet(..., checkpoints=[checkpoint])
+    >>> net.fit(X, y)
+    >>> # prints:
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@ottonemo Could you please take a look at the design and let me know if corresponds to what we discussed? If so, I'd continue working on this approach.

BenjaminBossan:
@ottonemo These are the changes following our latest discussion. The main addition is that now it's possible to buffer the parameters on disk instead of having to use in-memory buffers.

Unfortunately, the implementation is not as generic as we had hoped it could be: We cannot simply have a generic buffer argument and pass a `TextIOWrapper` instead of `BytesIO` there. A minor reason is that the buffer has to potentially be re-created multiple times, but that would be solvable using a "factory". A bigger concern is that the file is opened either in write or read mode, but not both. For buffering, we need write, for sending to Hub, we need to read from it. I tried some workarounds but they were all brittle and anyway, they would be specific to files on disk, so it wouldn't be a generic solution as we would have hoped.

In the end, I decided to go with a different approach where we have a `local_storage` argument. It is `None` by default, which means in-memory (using `BytesIO` under the hood). If a string or `Path` is passed, the parameters are buffered to a file on disk at that location. This implementation is more robust than the generic ones I could come up with, and still checks the goal to give an option to avoid in-memory storage. Maybe a better generic implementation is possible, please let me know if you have any ideas.

A note on testing: Since we mock away the `HfApi`, we cannot be sure that these tests wouldn't miss certain types of bugs. Alternatively, we could not use the mock but in this case, we would create real uploads (and need to have a valid token), which we want to avoid. Other than that, we could try to patch more specific functions used by the Hub API, e.g. `requests.post`, but that is more difficult to get right and the success of the patching depends on implementation details of the Hub API. Therefore, the current approach seems to be most reasonable.

skorch/hf.py
```diff
@@ -1004,3 +1006,182 @@ def _step_optimizer(self, step_fn):
...
+
+        python -m pip install huggingface_hub
+
+    Note that writes to the Hub are synchronous. Therefore, if the time it takes
+    to upload the data is long compared to training the model, there can be a
+    signficant slowdown. It is best to use this with
+    :class:`skorch.callbacks.training.TrainEndCheckpoint`, as that checkpoint
+    only uploads the data once, at the end of training. Also, using this writer
+    with :class:`skorch.callbacks.training.LoadInitState` is not supported for
+    now.
```
ottonemo, line 1032: maybe note why so that people are informed and may contribute a solution

skorch/hf.py
```diff
@@ -1004,3 +1006,182 @@ def _step_optimizer(self, step_fn):
...
+    Parameters
+    ----------
+    hf_api : instance of huggingface_hub.HfApi
+      Pass an instantiated ``huggingface_hub.HfApi`` object here.
+
+    path_in_repo : str
+      The name that the file should have in the repo, e.g. ``my-model.pkl``. If
+      you want each upload to have a different file name, instead of overwriting
+      the file, use a templated name, e.g. ``my-model-{}.pkl``. Then your files
+      will be called ``my-model-1.pkl``, ``my-model-2.pkl``, etc.
```
ottonemo, line 1043: This should note that this does not regard the previous contents of the repository. I think this should be made clear to the user

skorch/tests/test_hf.py
```diff
@@ -755,3 +757,376 @@ def train_step(self, *args, **kwargs):
...
+            f_criterion=None,
+            f_history=None,
+        )
+
+        net.set_params(callbacks=[checkpoint], max_epochs=10)
+        net.fit(*data)
+
+        # each time the valid loss improves, there should be a checkpoint
+        num_checkpoints_expected = sum(net.history[:, 'valid_loss_best'])
+        num_checkpoints_actual = len(mock_hf_api.calls)
```
ottonemo, line 947: this test should make sure that the number is > 0. while it is valid that there might not be best losses (currently there should be at least one but that might change), the test would be useless if #best = 0.

skorch/utils.py
```diff
@@ -482,17 +483,38 @@ def noop(*args, **kwargs):
     """
 
 
+class _FileLikeWrapper:
+    """Wrap a buffer or similar to be file like
+
+    This is so as to mirror more closely what PyTorch does under the hood (see
+    https://github.com/pytorch/pytorch/blob/master/torch/serialization.py).
+
+    """
```
ottonemo, line 492: please document the purpose of this class as it is not obvious.

BenjaminBossan:
@ottonemo I addressed your comments, please review again.

thomasjpfan (commented):
API-wise, I am happy with the interaction between `HfHubStorage` and `Checkpoint`.

Do you see a nice way to test this on a live huggingface hub bi-weekly to make sure the code works?

skorch/hf.py
```diff
@@ -1004,3 +1006,186 @@ def _step_optimizer(self, step_fn):
...
+            # two contexts, resulting in __exit__, and thus flush, being called
+            # twice
+            return
+
+        if isinstance(self.local_storage, (str, pathlib.Path)):
+            self._buffer.close()
+            path_or_fileobj = self._buffer.name
+        else:
+            self._buffer.seek(0)
+            path_or_fileobj = self._buffer
```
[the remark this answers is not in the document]
BenjaminBossan: Good point. I'm not absolutely sure if it's required (maybe it's sufficient that the buffer is gc'ed) but I still put it there. Unfortunately, that led to some weird pytest errors (not in the notebook though), I guess somehow related to clean up code for fixtures. So I had to make sure that the buffer is removed from the `state` when the object is pickled (which is probably for the better).

BenjaminBossan (commented):
> Do you see a nice way to test this on a live huggingface hub bi-weekly to make sure the code works?

How about a GH action that triggers once a week or bi-weekly and runs something similar to what the notebook is doing? I.e. basically an integration test. I would add that in another PR though.

skorch/hf.py
```diff
@@ -1004,3 +1006,186 @@ def _step_optimizer(self, step_fn):
...
+    >>> hub_params_writer = HfHubStorage(
+    ...     hf_api,
+    ...     path_in_repo=params_name,
+    ...     repo_id=repo_name,
+    ...     token=token,
+    ...     verbose=1,
+    ... )
+    >>> checkpoints = [
+    ...     TrainEndCheckpoint(f_pickle=hub_pickle_writer),
+    ...     TrainEndCheckpoint(f_params=hub_params_writer),
```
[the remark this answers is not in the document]
thomasjpfan: I can see how it works and makes sense for `skorch`. I'm wondering if Hugging Face's Hub is okay with users uploading just the weights. My guess is that it is okay, but I am not sure.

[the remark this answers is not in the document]
BenjaminBossan: I wouldn't know that there is any issue with that for HF Hub. There are some features that require the whole model to be there, like widgets that allow you to make predictions, but for that we have the pickle option available.

ottonemo (approved):
LGTM

Issue #883, opened 2022-08-04
Error using m1/mps acceleration after training


adelevie:
Not sure if you care to support this, but I tried two different ways of integrating PyTorch's m1/mps acceleration. In both cases, I get this error after training: 

```python
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:07<00:00,  1.17it/s, valid_loss=0.534]---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 pipeline.fit(X_train, y_train)

File ~/.pyenv/versions/project/lib/python3.9/site-packages/sklearn/pipeline.py:382, in Pipeline.fit(self, X, y, **fit_params)
    380     if self._final_estimator != "passthrough":
    381         fit_params_last_step = fit_params_steps[self.steps[-1][0]]
--> 382         self._final_estimator.fit(Xt, y, **fit_params_last_step)
    384 return self

File ~/.pyenv/versions/project/lib/python3.9/site-packages/skorch/classifier.py:141, in NeuralNetClassifier.fit(self, X, y, **fit_params)
    130 """See ``NeuralNet.fit``.
    131 
    132 In contrast to ``NeuralNet.fit``, ``y`` is non-optional to
   (...)
    136 
    137 """
    138 # pylint: disable=useless-super-delegation
    139 # this is actually a pylint bug:
    140 # https://github.com/PyCQA/pylint/issues/1085
--> 141 return super(NeuralNetClassifier, self).fit(X, y, **fit_params)

File ~/.pyenv/versions/project/lib/python3.9/site-packages/skorch/net.py:1235, in NeuralNet.fit(self, X, y, **fit_params)
   1232 if not self.warm_start or not self.initialized_:
   1233     self.initialize()
-> 1235 self.partial_fit(X, y, **fit_params)
   1236 return self

File ~/.pyenv/versions/project/lib/python3.9/site-packages/skorch/net.py:1194, in NeuralNet.partial_fit(self, X, y, classes, **fit_params)
   1192 self.notify('on_train_begin', X=X, y=y)
   1193 try:
-> 1194     self.fit_loop(X, y, **fit_params)
   1195 except KeyboardInterrupt:
   1196     pass

File ~/.pyenv/versions/project/lib/python3.9/site-packages/skorch/net.py:1112, in NeuralNet.fit_loop(self, X, y, epochs, **fit_params)
[...]

PR #884, opened 2022-08-04
Move mps data to CPU in util.to_numpy

adelevie:
Ostensibly solves https://github.com/skorch-dev/skorch/issues/883, though I am not very familiar with this codebase.

### DIFF
--- a/skorch/utils.py
+++ b/skorch/utils.py
@@ -165,6 +165,9 @@ def to_numpy(X):
     if X.is_cuda:
         X = X.cpu()
 
+    if hasattr(X, 'is_mps') and X.is_mps:
+        X = X.cpu()
+
     if X.requires_grad:
         X = X.detach()
 
--- a/skorch/tests/test_utils.py
+++ b/skorch/tests/test_utils.py
@@ -199,6 +199,16 @@ class TestToNumpy:
         expected = "Cannot convert this data type to a numpy array."
         assert e.value.args[0] == expected
 
+    @pytest.mark.skipif(
+        not (hasattr(torch.backends, "mps") and torch.backends.mps.is_available()),
+        reason='Skipped because mps is not available as a torch backend'
+    )
+    def test_mps_support(self, to_numpy, x_tensor):
+        device = torch.device('mps')
+        x_tensor.to(device)
+        x_numpy = to_numpy(x_tensor)
+        self.compare_array_to_tensor(x_numpy, x_tensor)
+
 
 class TestToDevice:
     @pytest.fixture


BenjaminBossan:
Thank you for the bug report and providing a possible solution. This is probably hard to test, especially with the CI we have. Did you test it for your use case and did it solve the issue?

The M1 support from PyTorch still seems to be WIP (see their issue 77764 (not linking there to avoid spam)) but I guess there is no harm in adding support for this even without a proper way of testing, since the change seems innocuous enough.

adelevie:
I can say that the `TypeError: can't convert mps:0 device type tensor to numpy.` error went away and the model did train with no errors.

I figured duck-typing `hasattr(X, 'is_mps')` was the safest way to work around `torch` versions that lack MPS support. Looks like it didn't break anything already in CI.

BenjaminBossan:
> I can say that the `TypeError: can't convert mps:0 device type tensor to numpy.` error went away and the model did train with no errors.

Okay, that's great.

> I figured duck-typing `hasattr(X, 'is_mps')` was the safest way to work around `torch` versions that lack MPS support. Looks like it didn't break anything already in CI.

Yes, I believe that's the best way of doing it (short of PyTorch providing an official, backwards compatbile method to check it).

I don't believe there is much more that can be done here, so I'm willing to merge. I'll just wait a few days to see if @ottonemo or @thomasjpfan have anything to add.

skorch/utils.py
```diff
@@ -165,6 +165,10 @@ def to_numpy(X):
     if X.is_cuda:
         X = X.cpu()
 
+    if hasattr(X, 'is_mps'):
+        if X.is_mps:
+            X = X.cpu()
```
thomasjpfan, line 170: Nit:

```suggestion
    if hasattr(X, 'is_mps') and X.is_mps:
        X = X.cpu()
```

thomasjpfan (commented):
Overall, I am okay with this PR.

I think we can add a quick test for `to_numpy` that only runs when `mps` is available. Here is a snippet to check if `mp` is available:

```python
hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
```

BenjaminBossan:
> I think we can add a quick test for `to_numpy` that only runs when `mps` is available

Yes, why not. It will probably not run for the time being but at least we have something if mps should become available on CI (or one of the skorch users with a MB runs the test locally). @adelevie would you be so kind as to add a test?

BenjaminBossan:
Let's just add a new test function to the existing class. The function checks if mps is available, as suggested by Thomas, and skips if it isn't. Then create an mps tensor, run `to_numpy` on it, and check that it's a numpy array. That should be good enough.

adelevie:
Sounds good, shouldn’t be too much trouble.

adelevie:
Should be good to go. I squashed to single cleaner commit, hope that's ok.

thomasjpfan (approved):
Thank you for the update! I tested this on my m1 mac and the test runs and passes.

BenjaminBossan:
Thanks @adelevie and @thomasjpfan.

> I squashed to single cleaner commit, hope that's ok.

Yeah, that's okay, but we would have squashed anyway, so it wasn't necessary.

Issue #887, opened 2022-08-23
Distutils Version classes (including LooseVersion) are deprecated


timokau:
The `LooseVersion` class (see #579) is now deprecated ([distutils commit](https://github.com/pypa/distutils/commit/92082ee42c), [setuptools changelog](https://setuptools.pypa.io/en/latest/history.html#v59-6-0)), which results in warnings of the form:

```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(sklearn.__version__) >= '0.22.0':
```

We should probably follow the suggestion and move to `packaging.version`.

BenjaminBossan:
I didn't know that, thanks for bringing it up.

Is `packaging` part of the standard Python distribution or an extra dependency for all our supported Python versions? It seems not to exist for < 3.8, could it be? If so, I guess we would need to write the code in a way to still use `LooseVersion` for old Python versions.

timokau:
I'm not sure, but as far as I know it's an extra dependency but it should be available for python >= 3.6 according to [pypi](https://pypi.org/project/packaging/#description).

thomasjpfan:
Using `packaging.version` is known to be slow: https://github.com/pypa/setuptools/issues/926

For Python >= 3.8, the best way to get versions is: [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html). Since we still support 3.7, there is a [shim on pypi](https://pypi.org/project/importlib-metadata/#description).

If we do not want to add another dependency, we can vendor the code for `packaging.version`. For example, this is how SciPy vendors it: https://github.com/scipy/scipy/blob/main/scip

BenjaminBossan:
Why is something that is seemingly simple so complex? :(

> For Python >= 3.8, the best way to get versions is: [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html).

One difference would be that this would compare strings, whereas with `LooseVersion`, we get a bit of a more sophisticated logic. E.g.:

`LooseVersion(sklearn.__version__).version  # => [1, 2, 'dev', 0]`

Maybe it makes no difference for what we use, but I would much prefer a solution that uses the existing logic to compare versions.

> For example, this is how SciPy vendors it: https://github.com/scipy/scipy/blob/main/scip

The link is broken but in general, if we're just dealing with a few lines of code, I'd prefer vendoring to adding a new dependency.

thomasjpfan:
[...]

PR #888, opened 2022-08-26
Remove outdated version checks

BenjaminBossan:
Remove version checks that are no longer required because we don't
support the min versions anymore.

I was working on the proposed solution in #887 when I noticed that none
of the version checks we do is actually required anymore. After removing
them, we thus no longer depend on distutils's `LooseVersion`.

If we ever need to do a version check in the future, I would still go
with the [scipy Version class](https://github.com/scipy/scipy/blob/main/scipy/_lib/_pep440.py).

Also, in the future, we should ensure to add `# TODO` comments to all
version checks so that we can easily find them.

One change I introduced here is to require scikit-learn>=1.0.0. I think
this should be fine because it's been out for some time now and it still
supports Python 3.7.

We could go to 0.24 instead and would still not need a version check,
let me know if you prefer that.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added experimental support for [huggingface accelerate](https://github.com/huggingface/accelerate); use the provided mixin class to add advanced training capabilities provided by the accelerate library to skorch
 
 ### Changed
-- The minimum required scikit-learn version has been bumped to 1.0.0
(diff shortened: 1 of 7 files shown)


thomasjpfan (commented):
> We could go to 0.24 instead and would still not need a version check,
let me know if you prefer that.

I prefer 0.24 because it would be slightly easier to update skorch to the latest version.

Also from looking at the changes, I think we can go down to 0.22?

BenjaminBossan:
Ah yes, you're right, 0.22 it is.

thomasjpfan (approved):
LGTM

PR #889, opened 2022-09-02
Update MNIST.ipynb

chrishwiggins:
threw an error as written; should work w/o the added arguments

error:

```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-19c8d440ea79> in <module>
----> 1 mnist = fetch_openml('mnist_784', as_frame=False, cache=False)

TypeError: fetch_openml() got an unexpected keyword argument 'as_frame'```

### DIFF
--- a/notebooks/MNIST.ipynb
+++ b/notebooks/MNIST.ipynb
@@ -61,7 +61,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "mnist = fetch_openml('mnist_784', as_frame=False, cache=False)"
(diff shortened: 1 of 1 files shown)


ottonemo:
Hey, thanks for the PR :)
What version of sklearn do you use to test this with? As far as I can see version 0.22 has this parameter (and is the minimum version that skorch requires).

BenjaminBossan:
@chrishwiggins I'm closing this for now, given the comment by @ottonemo. If there is something we missed, please open an issue.

PR #890, opened 2022-09-02
Skip GPyTorch tests if torch version too low

BenjaminBossan:
_Just as we thought we could get rid of all version checks..._

GPyTorch requires relatively high versions of PyTorch. Therefore,
certain tests in the CI matrix may fail for certain GPyTorch versions.
We thus skip the GPyTorch tests in case the PyTorch version is not
sufficiently high.

We don't want to force the PyTorch version itself on the user (as long
as it's sufficiently high) because most users probably don't use
GPyTorch.

To check the versions, as discussed in #888, we use the version checking
code copied from scipy.

### DIFF
--- a/skorch/tests/test_probabilistic.py
+++ b/skorch/tests/test_probabilistic.py
@@ -12,11 +12,22 @@ from sklearn.pipeline import Pipeline
 import torch
 from torch.testing import assert_allclose
 
+from skorch._version import Version
 from skorch.utils import is_torch_data_type
 from skorch.utils import to_numpy
 
 
 gpytorch = pytest.importorskip('gpytorch')
+
+# check that torch version is sufficiently high for gpytorch, otherwise skip
+version_gpytorch = Version(gpytorch.__version__)
+version_torch = Version(torch.__version__)
+# TODO: remove if newer GPyTorch versions are released that no longer require
+# the check.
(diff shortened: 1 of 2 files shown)


skorch/tests/test_probabilistic.py
```diff
@@ -12,12 +12,23 @@
...
 
 
 gpytorch = pytest.importorskip('gpytorch')
 
+# check that torch version is sufficiently high for gpytorch, otherwise skip
+version_gpytorch = Version(gpytorch.__version__)
+version_torch = Version(torch.__version__)
+# TODO: remove if newer GPyTorch versions are released that no longer require
+# the check.
+if (version_gpytorch >= Version('1.9')) and (version_torch < Version('1.11')):
```
thomasjpfan, line 27: The alternative way to skip a whole module is setting [pytestmark](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module) to `pytest.mark.skipif`. I have no strong opinion either way.

thomasjpfan (approved):
LGTM

skorch/tests/test_probabilistic.py
BenjaminBossan: I didn't know that, thanks for the pointer. Between the two variants, I think the existing one is a bit more obvious without consulting the pytest docs.

PR #894, opened 2022-09-26
Edited notebook name to fix Colab & GitHub links

### LINKED ISSUE #893: Redirection from notebook example to Open in Colab doesn't work
You need to authorize GitHub from Colab and save a copy in GitHub of a Colab notebook if you want to add a redirection from GitHub to Colab. 
Ekran Resmi 2022-09-26 14 29 01
Ekran Resmi 2022-09-26 14 29 10

merveenoyan:
Should fix #893 turns out the name of the notebook in repository and the links inside notebook don't match.


merveenoyan:
I initially tried looking at https://colab.research.google.com/github/skorch-dev/skorch/blob/master/notebooks/Hugging_Face_Finetuning.ipynb because that was what notebook was called, but it was entered as https://colab.research.google.com/github/skorch-dev/skorch/blob/master/notebooks/Huggingface_Finetuning.ipynb instead which that notebook doesn't exist inside repo so I changed name of the notebook in the repo.

BenjaminBossan:
Ah yes, that error came about because I first started out with `Huggingface` everywhere but then learned that `Hugging Face` is the preferred name.

I think this change (which is in PR) should fix the issue, right?

https://github.com/skorch-dev/skorch/pull/881/files#diff-e4b3e3d658c5d0c4606bd4619921657721c92bc5f27bea3d828fc6ff12c07d7c

If so, we can close this PR.

merveenoyan:
@BenjaminBossan yes it would.

PR #896, opened 2022-09-27
Add tests about saving/loading with accelerate

BenjaminBossan:
When using `AccelerateMixin`, the net cannot be pickled. This is caused by accelerate replacing `forward` methods by wrapped methods:

https://github.com/huggingface/accelerate/blob/6d038e19a168f5a270bb4c535023255a0160fc10/src/accelerate/accelerator.py#L747-L753

Pickle does not know how to restore those methods and thus fails.

In this PR, we're adding unit tests to check this and documentation indicating this missing feature (and pointing users towards using `net.save_params` instead).

It is not impossible to make pickling work. We could save a copy of the original methods before they're replaced (or dig out the original function from `__closure__.cell_contents`) and after training restore the methods.

It's however not that trivial, since preparing happens during initialize, which can be called outside of fit. Therefore, a simple context manager is not enough. Also, we do want to have acceleration during inference, so methods such as predict need to prepare and restore again.

We could also try to mess with `__reduce__` but that seems to be equally brittle. So maybe we just need to live with pickle not working with `accelerate`.


BenjaminBossan:
Regarding the failing tests, see #897

thomasjpfan (approved):
I'm okay with this PR as it documents and test the current behavior.

Is the `nn.Module` itself not picklable after autocasting the forward method? In any case, can we open an issue to track pickling+accelerate?

BenjaminBossan:
> Is the `nn.Module` itself not picklable after autocasting the forward method?

Yes, exactly, since the `forward` method is overwritten, the whole module cannot be loaded. This affects the `module_` and `criterion_`.

> can we open an issue to track pickling+accelerate?

Done: #898

PR #897, opened 2022-09-27
Bugfix tests hf tokenizers

BenjaminBossan:
1. `xfail`ing test doesn't fail anymore (cloning now works)
2. delimiter needs to be `" "` for `tokenizers==v0.13`, other delimiters don't seem to work

### DIFF
--- a/skorch/tests/test_hf.py
+++ b/skorch/tests/test_hf.py
@@ -290,14 +290,14 @@ class TestHuggingfaceTokenizerUninitialized(_HuggingfaceTokenizersBaseTest):
         tokenizer.set_params(
             model__dropout=0.123,
             trainer__vocab_size=123,
-            pre_tokenizer__delimiter='*',
+            pre_tokenizer__delimiter=' ',
             max_length=456,
         )
         tokenizer.fit(data)
 
         assert tokenizer.tokenizer_.model.dropout == pytest.approx(0.123)
         assert len(tokenizer.vocabulary_) == pytest.approx(123, abs=5)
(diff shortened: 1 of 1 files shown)


skorch/tests/test_hf.py
```diff
@@ -290,14 +290,20 @@ def test_set_params(self, data):
...
             trainer__vocab_size=123,
-            pre_tokenizer__delimiter='*',
             max_length=456,
+            # With v0.13 of tokenizers, it seems like delimiter always needs to
+            # be " ", otherwise this error is raised: Error while attempting to
+            # unpickle Tokenizer: data did not match any variant of untagged
+            # enum ModelWrapper at line 1 column 2586. So we cannot change its
+            # value in this test but we should still ensure that set_params
+            # doesn't fail, so we keep it.
+            pre_tokenizer__delimiter=' ',
```
thomasjpfan, line 300: Coming from the [CharDelimiterSplit](https://huggingface.co/docs/tokenizers/v0.13.0/en/api/pre-tokenizers#tokenizers.pre_tokenizers.CharDelimiterSplit) docs, this is the same as `.split(delimiter)`, which I suspect should be `" "` in the first place.

What was the original motivation for `'*'`?

BenjaminBossan: > What was the original motivation for `'*'`?

No reason at all, I just wanted to change it to a different value through `set_params` to ensure that it really works, the actual value didn't matter.

thomasjpfan (approved):

Issue #901, opened 2022-10-06
Integration tests with Hugging Face Hub


BenjaminBossan:
As discussed [here](https://github.com/skorch-dev/skorch/pull/881#pullrequestreview-1126775733), it would be nice to have some kind of integration test with HF Hub, now that we support it directly. However, we shouldn't run it as part of our normal test suite, because:

1. It can be slow and requires i/o to an external service, which can be flaky
2. It requires a HF Hub token. We can add it as a secret on GH but that wouldn't help external users who want to run the tests.

It would probably be sufficient to run these tests once a week or so.

PR #904, opened 2022-10-07
Add Hugging Face integration tests

BenjaminBossan:
Solves #901

This PR adds a github workflow that runs integration tests with Hugging Face. These tests cover:

- using a pretrained tokenizer
- using a pretrained BERT model
- storing model artifacts on the HF model hub

The tests can be triggered manually [here](https://github.com/skorch-dev/skorch/actions) and additionally run once per week on Sundays at 9:00 AM UTC.

I'm not sure if we can really test if the workflow works before merging, any ideas?

### DIFF
--- a/.github/workflows/test-hf-integration.yml
+++ b/.github/workflows/test-hf-integration.yml
@@ -1,0 +1,28 @@
+name: HF integration tests
+
+on:
+  schedule:
+    - cron:  '0 9 * * 0'  # every sunday at 9:00 UTC
+  workflow_dispatch:
+
+jobs:
+  hf-integration-tests:
+
+    runs-on: ubuntu-latest
+    if: "github.repository == 'skorch-dev/skorch'"
+
+    timeout-minutes: 10
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
+    - name: Install Requirements
+      run: |
+        python -m pip install -r requirement.txt
+        python -m pip install transformers tokenizers huggingface_hub torch
+        python -m pip install .
+    - name: run HF integration test
+      env:
+        HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      run: python scripts/hf-integration-tests.py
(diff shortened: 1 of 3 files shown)


thomasjpfan (commented):
> I'm not sure if we can really test if the workflow works before merging, any ideas?

With a secret, I think the only way to check is to run the GitHub Action on a fork and link the results here.

BenjaminBossan:
> With a secret, I think the only way to check is to run the GitHub Action on a fork and link the results here.

That's a good idea, I'll try it out.

BenjaminBossan:
@thomasjpfan So I ran this on my fork of the repo and at least the manuel trigger works:

https://github.com/BenjaminBossan/inferno/actions/workflows/test-hf-integration.yml

As for the cron schedule, it didn't trigger, even though I tried different schedules. There are a few possible reasons for that:

1. GitHub is reportedly not guaranteeing that the scheduled tasks run as indicated, it's more of a perhaps thing.
2. People have reported problems based on the branch but since I merged into master, that should not be the issue.
3. It's caused by the repo being a fork.
4. I made a mistake in the workflow, even though it's pretty simple.

Anyway, I'd be willing to go ahead, worst case the cron schedule doesn't trigger, we'd still have the manual job.

Another question: Do we want to add a badge for this test to the README?

thomasjpfan:
> Another question: Do we want to add a badge for this test to the README?

That is the simplest solution to keep track of the scheduled job.

The alternative solution is to auto-open an issue if the scheduled job fails. An example of a self contained version of this is in [pydata/xarray](https://github.com/pydata/xarray/blob/8eea8bb67bad0b5ac367c082125dd2b2519d4f52/.github/workflows/upstream-dev-ci.yaml#L132-L183).

BenjaminBossan:
I added a badge. This is what it should look like:

https://github.com/BenjaminBossan/inferno/actions/workflows/test-hf-integration.yml/badge.svg

Other than that, I think we can also set the repo to send us notifications when an action fails, so auto-opening an issue for a feature which is not core to skorch seems to be overkill to me.

thomasjpfan (approved):
LGTM

BenjaminBossan:
Just to note, the job successfully run via the scheduler: https://github.com/skorch-dev/skorch/actions/runs/3258933674. However, I couldn't figure out at what time it ran, so I'm not sure if the configured time was indeed used.

PR #905, opened 2022-10-11
The entry for the badge was missing

BenjaminBossan:
A reference to the badge was added in #904 but for it to show, there needs to be an entry using that reference.


thomasjpfan (approved):
Ah I see you use GitHub in Dark mode! 😁

This PR LGTM

PR #906, opened 2022-10-11
Update the Neptune integration

twolodzko:
- Neptune integration was updated to work with a recent version of the Neptune client.
- Additional logging was added: model summary, configuration (learning rate, optimizer, etc), `event_lr`, and changepoints (when available).
- Tests were altered to match the current API of the client library and improved when possible.


skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
...
         keys_ignored = self.keys_ignored
         if isinstance(keys_ignored, str):
             keys_ignored = [keys_ignored]
         self.keys_ignored_ = set(keys_ignored or [])
         self.keys_ignored_.add('batches')
         return self
 
+    def on_train_begin(self, net, X, y, **kwargs):
+
+        self._metric_logger['model/model_type'] = NeptuneLogger._get_obj_name(net.module_)
```
BenjaminBossan, line 202: Can we change all uses of `NeptuneLogger._get_obj_name` to `self._get_obj_name`? Same for `_model_summary_file`. For static methods, it doesn't really matter, but I think that `self` is more intuitive here and if it ever needs to be changed to a bound method, these calls don't need to be edited.

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
...
+        self._metric_logger['model/summary'] = NeptuneLogger._model_summary_file(net.module_)
+
+        self._metric_logger['config/optimizer'] = NeptuneLogger._get_obj_name(net.optimizer_)
+        self._metric_logger['config/criterion'] = NeptuneLogger._get_obj_name(net.criterion_)
```
BenjaminBossan, line 206: Just a minor note: In skorch, nets can have more than one module, criterion, and optimizer. Here is an examplary snippet that shows how to retrieve all of modules and criteria:

https://github.com/skorch-dev/skorch/blob/5ec195c8eb20a85d8e657595593f44248732b5b9/skorch/net.py#L848-L850

The same logic can be applied for optimizers.

It's fine to keep it as is, though, since I assume only few users make use of it and from all skorch classes, only the GPyTorch classes in `probabilistic.py` add custom modules (the likelihood).

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
...
+        except ImportError:
+            # neptune-client>=1.0.0 package structure
+            from neptune.types import File
+
+        return File.from_content(str(model), extension='txt')
+
+    @staticmethod
+    def _get_checkpoint_dirname(net):
+        for callback in net.callbacks:
+            if isinstance(callback, skorch.callbacks.training.Checkpoint):
```
BenjaminBossan, line 273: This is a bit of a crude way of determining the checkpoint file name. E.g., there can be more than one `Checkpoint` or users could use `TrainEndCheckpoint` or some other custom class. I think I would prefer for this to be more explicit, e.g. the user could pass the checkpoint dirname as an argument. WDYT?

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
...
+                return callback.dirname
+
+    def _upload_checkpoint(self, net):
+        checkpoint_dirname = self._get_checkpoint_dirname(net)
+        if checkpoint_dirname is not None:
+            if checkpoint_dirname == '':
+                checkpoint_dirname = '.'
+            self._metric_logger['model/checkpoint'].upload_files([
+                f'{checkpoint_dirname}/*.json',
+                f'{checkpoint_dirname}/*.pt',
```
BenjaminBossan, line 283: In case this code stays: Would it also work on Windows?

BenjaminBossan (changes requested):
Thank you very much for the update. I played around with the new neptune logger and experiment tracking site and was impressed. It's very useful and there are nice additions since last time I checked.

I have a few code comments below, please take a look. On top of those, I have some more general comments:

### Versions

I saw that you have some code that works with `neptune-client>=1.0`, which I didn't find released on PyPI. So I assume it's for forward compatibility for a yet to be released version. Big :+1:.

That said, I ran into some issues with certain Neptune versions (see one of my code comments). So I think that there could still be an issue with supporting all versions `>=0.9.0`. I would thus like to ask you to check again if it works with all intended versions. Maybe some older ones can be dropped? At the end of the day, you know best which versions should still be supported and how.

### Pickling

A minor inconvenience I found is that the `NeptuneLogger` apparently can't be pickled:

> AttributeError: Can't pickle local object `'NeptuneAuthenticator.__init__.<locals>.session_factory'`

Therefore, a user would not be able to save the whole skorch net on Neptune. If there is an easy way to fix this by doing something with `NeptuneLogger.get_state`, that would be nice, but I'm not sure. On the neptune-client side, these lines seem to be the offender:

https://github.com/neptune-ai/neptune-client/blob/81fcf25b204331ae85b67e19afc78be69ddf3f84/src/neptune/oauth.py#L98-L126

Since a reference to a local function is stored, pickle fails. So if you ever want to fix this, you would have to move this definition to the module level.

### Tests

While running the tests, I got this message multiple times:

> Error in atexit._run_exitfuncs:

Not sure what it means and if it can cause any trouble.

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
     ... y_pred = net.predict_proba(X)
```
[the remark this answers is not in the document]
twolodzko: Fixed

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
...
         self.log_on_batch_end = log_on_batch_end
         self.close_after_train = close_after_train
         self.keys_ignored = keys_ignored
 
-    def initialize(self):
-        self.first_batch_ = True
+        if base_namespace.endswith("/"):
+            self._base_namespace = base_namespace[:-1]
+        else:
+            self._base_namespace = base_namespace
```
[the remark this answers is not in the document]
twolodzko: Done

skorch/callbacks/logging.py
twolodzko: Sure, will do.

skorch/callbacks/logging.py
twolodzko: @BenjaminBossan thanks, I wasn't aware of that. I'll remove this functionality from the callback and instead document how tracking the checkpoints can be achieved by just providing the path to the tracked files to `Run` object.

skorch/callbacks/logging.py
```diff
@@ -116,97 +117,171 @@ class NeptuneLogger(Callback):
     ... y_pred = net.predict_proba(X)
     ... auc = roc_auc_score(y, y_pred[:, 1])
     ...
-    ... neptune_logger.experiment.log_metric('roc_auc_score', auc)
+    ... neptune_logger.run["roc_auc_score"].log(auc)
 
-    >>> # log charts like ROC curve
+    >>> # Log charts, such as an ROC curve
```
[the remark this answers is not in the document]
twolodzko: Done.

skorch/callbacks/logging.py
twolodzko: @BenjaminBossan could you point me to a code example using multiple modules? What in such case would be logged by `net.module_`, etc?

skorch/callbacks/logging.py
```diff
@@ -65,18 +66,18 @@ def on_epoch_end(self, net, **kwargs):
 
 
 class NeptuneLogger(Callback):
-    """Logs results from history to Neptune
+    """Logs results from history to Neptune.
```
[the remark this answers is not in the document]
twolodzko: Done

skorch/callbacks/logging.py
BenjaminBossan: As mentioned, this is an edge case, so feel free to ignore it, maybe leave a `# TODO` comment.

In case you want to tackle it, this test probably best demonstrates what it means if a user sets a custom module:

https://github.com/skorch-dev/skorch/blob/5ec195c8eb20a85d8e657595593f44248732b5b9/skorch/tests/test_net.py#L2662-L2671

As you can see, they can give any name like `mymodule_` to the attribute but we can find it inside of `self._modules`. The `module_` is still set but it just means that there is more than one `nn.Module`. There is no telling which one is the "main" module, so it would be best to log all of them, not sure if there is a limit in Neptune.

BenjaminBossan:
@twolodzko Let me know when you are finished and need another review round.

twolodzko:
> ### Pickling
> A minor inconvenience I found is that the `NeptuneLogger` apparently can't be pickled:
> 
> > AttributeError: Can't pickle local object `'NeptuneAuthenticator.__init__.<locals>.session_factory'`
[...]

Would this be blocking? It seems that this is something we should rather fix in the Neptune client library that is unpickable than this logger. Would it be ok if we merged the logger like this, but in the meantime we would figure out on our side how to make the client pickle-friendly?

> ### Tests
> While running the tests, I got this message multiple times:
> 
> > Error in atexit._run_exitfuncs:
> 
> Not sure what it means and if it can cause any trouble.

This is not a problem, they can be ignored. The errors raise because we are using async calls and pytest does not handle closing them well. If I understand correctly, the tests have passed but you just saw this in the logs?

twolodzko:
@BenjaminBossan I think I addressed all of your comments.

skorch/tests/callbacks/test_logging.py
```diff
@@ -16,10 +18,12 @@
 from skorch.tests.conftest import tensorboard_installed
 from skorch.tests.conftest import mlflow_installed
 
-
 @pytest.mark.skipif(
     not neptune_installed, reason='neptune is not installed')
 class TestNeptune:
+    # fields logged by on_train_begin and on_train_end
```
BenjaminBossan, line 24: Regarding the `Error in atexit._run_exitfuncs:` error:

> This is not a problem, they can be ignored. The errors raise because we are using async calls and pytest does not handle closing them well. If I understand correctly, the tests have passed but you just saw this in the logs?

Yes, the tests have passed but the message is confusing. I'm worried that future users (me included) will be confused by this message and think something went wrong. If you know of any way to turn those off, that would be appreciated.

What's strange is that they don't always appear:

```python
pytest -k TestNeptune           # message
pytest -k TestNeptune -s        # no message
pytest -k test_logging          # no message
pytest  # i.e. running all tests, no message

pytest -k "TestTensorboard or TestNeptune"  # message
pytest -k "TestNeptune or TestTensorboard"  # message
pytest -k "TestNeptune or TestSacred"       # message
pytest -k "TestNeptune or TestWandB"        # no message
```

So thankfully it does not appear most of the times, only sometimes if only Neptune or Neptune + a few specific others are tested. This means that a user is most likely to see the message when working with the Neptune tests specifically. If the message cannot be turned off, could you please add a comment here to indicate that the message can be ignored?

BenjaminBossan (changes requested):
Thank you a lot for making the changes. Everything looks very good now. I only have a minor change request.

> Would this be blocking? It seems that this is something we should rather fix in the Neptune client library that is unpickable than this logger. Would it be ok if we merged the logger like this, but in the meantime we would figure out on our side how to make the client pickle-friendly?

No, not an issue. As a skorch user, I can still use the Neptune logger and then purge it from the callbacks to be able to pickle the model. Most often, the logger would not be required after loading anyway, since at that point, most users are only interested in predicting with the model.

If you do make a change in Neptune that ends up enabling pickle, please let us know and we can add a unit test to check this.

skorch/tests/callbacks/test_logging.py
twolodzko: @BenjaminBossan I think I found the cause for those messages to appear in the tests and fixed it. I don't see them when testing it locally, so please comment if you see it on your side.

skorch/callbacks/logging.py
```diff
@@ -65,147 +65,215 @@ def on_epoch_end(self, net, **kwargs):
...
     def on_train_end(self, net, **kwargs):
+        try:
+            self._metric_logger['train/epoch/event_lr'].log(net.history[:, 'event_lr'])
+        except KeyError:
+            pass
         if self.close_after_train:
-            self.experiment.stop()
+            self.run.stop()
+
+    def _log_metric(self, name, logs, batch):
```
shnela, line 240: I'm a little puzzled about this function.
Are those shortcuts like `dur` or `valid` some internal `skorch` strings which we're replacing to increase the readability of our experiment?
Could we add some comment here?

twolodzko: @shnela yes. Skorch logs `dur` dor duration of the epoch and several metrics like `valid_loss` or `train_loss`, where we extend `valid` to `validation` fro readability.

twolodzko:
> I have a suggestion to make regarding the _extract_name_parts function.  [...]

Nice idea. Done.

shnela (approved):

BenjaminBossan (approved):
Thanks a lot @twolodzko @normandy7 @shnela. Very nice additions and well thought out changes.

Herudaio:
@BenjaminBossan is there any rough ETA on when a release including these changes may happen? :)

BenjaminBossan:
No, not really, we typically just release when enough changes have accumulated and the last release is very recent, so it'll probably be a couple of months. If it's important for you to have an earlier release, we could consider a patch-level release.

Herudaio:
Thanks for the information.
> If it's important for you to have an earlier release, we could consider a patch-level release.
@BenjaminBossan that would be awesome, if there's such a possibility! :)

Herudaio:
Thanks for the information.
> If it's important for you to have an earlier release, we could consider a patch-level release.

@BenjaminBossan that would be awesome, if there's such a possibility! :)

BenjaminBossan:
@Herudaio Yes, I think we can do that, hopefully not later than end of next week.

BenjaminBossan:
@Herudaio We just released v0.12.1 of skorch with the updated logger.

PR #908, opened 2022-10-13
Fix false warning if iterator_valid__shuffle=False

### LINKED ISSUE #907: `UserWarning: You set iterator_valid__shuffle=True` after upgrading to 0.12.0, but it is False
After upgrading to 0.12.0, I get the warning before training starts:
```
UserWarning: You set iterator_valid__shuffle=True; this is most likely not what you want because the values returned by predict and predict_proba will be shuffled.
```

However, I definitely create the model with `iterator_valid__shuffle = False`. And, just to confirm, after training I do
```python
print(estimator.get_params()["iterator_valid__shuffle"])  # False
```

I'm hoping the warning itself is wrong, otherwise I don't see how I can fix this within my own code.

BenjaminBossan:
Resolves #907

The regression initially occurred because the we did not check for the value of `iterator_valid__shuffle`, just the presence of the key.

Even though there is a test for this, the test didn't detect the error. This is because during a refactor (#751), the parameter validation was moved to `initialize()` from `__init__()` but the test was not adjusted to take the change into account. For clarity, I moved the test to a separate function (it used to be a subtest of another test).

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1914,11 +1914,12 @@ class NeuralNet:
         # warn about usage of iterator_valid__shuffle=True, since this
         # is almost certainly not what the user wants
         if 'iterator_valid__shuffle' in self._params_to_validate:
-            warnings.warn(
-                "You set iterator_valid__shuffle=True; this is most likely not "
-                "what you want because the values returned by predict and "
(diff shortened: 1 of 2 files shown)


thomasjpfan (approved):
LGTM

PR #909, opened 2022-10-17
DOC Update the documentation in several places

BenjaminBossan:
I went through what I thought would be the most commonly visited parts of our documentation and made improvements when I saw the opportunity.

Mostly, this comes down to wording, adding information about changed/new features, adding links or fixing broken ones, etc., nothing major.

### DIFF
--- a/docs/user/neuralnet.rst
+++ b/docs/user/neuralnet.rst
@@ -382,8 +382,9 @@ the result of
 :func:`~skorch.classifier.NeuralNetClassifier.predict_proba`.
 Obviously, this only makes sense if
 :func:`~skorch.classifier.NeuralNetClassifier.predict_proba` returns
-class probabilities or logits. If this is not true, you should just use
-:func:`~skorch.classifier.NeuralNetClassifier.predict_proba`.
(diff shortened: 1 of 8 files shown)


docs/user/neuralnet.rst
```diff
@@ -374,15 +382,15 @@ the result of
 :func:`~skorch.classifier.NeuralNetClassifier.predict_proba`.
 Obviously, this only makes sense if
 :func:`~skorch.classifier.NeuralNetClassifier.predict_proba` returns
-class probabilities. If this is not true, you should just use
+class probabilities or logits. If this is not true, you should just use
 :func:`~skorch.classifier.NeuralNetClassifier.predict_proba`.
```
ottonemo, line 386:
```suggestion
class probabilities or logits. If this is not true, you should avoid 
:func:`~skorch.classifier.NeuralNetClassifier.predict` and just use
:func:`~skorch.classifier.NeuralNetClassifier.predict_proba` to retrieve predictions.
```

BenjaminBossan: Done

PR #911, opened 2022-11-01
Update CI to use PyTorch 1.13

BenjaminBossan:
Remove 1.9.

Also, removed excluded combination from test matrix to check if they work now.

Locally, PyTorch 1.13 worked, and the change log also doesn't contain anything to suggest otherwise.


BenjaminBossan:
So there is an error when trying to pickle `accelerate`d optimizers with PyTorch 1.13. I'll investigate.

BenjaminBossan:
The bug was quickly fixed [here](https://github.com/huggingface/accelerate/pull/811) but it's not released yet.

PR #913, opened 2022-11-08
Don't fail when gpytorch import fails

BenjaminBossan:
This is mostly relevant for CI, as most users probably don't have gpytorch installed.

The problem is that some gpytorch versions now raise an `AttributeError` with some torch versions. Since we try to import gpytorch in `utils.py` and in `test_probabilistic.py` but only catch `ImportError`, the `AttributeError` will be raised. However, the presence of an incompatible gpytorch version should not preclude the usage of skorch.

In the gpytorch tests themselves, we already check for compatible versions and skip if they are not compatible. However, to check versions, we first need to successfully import gpytorch, which fails.

Here is an example of a job that currently fails because of this:

https://github.com/skorch-dev/skorch/actions/runs/3421070263/jobs/5696711066

The solution now is to catch `AttributeError`s during gpytorch imports. For `utils.py`, we then treat it as if gpytorch were not installed, for the tests we skip. A disadvantage of this indiscriminate skipping is that it could happen that all versions are skipped and we won't notice since CI is green. I don't have a good solution for this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix an error that could occur with specific combinations of gpytorch and PyTorch versions (#913)
+
 ## [0.12.0] - 2022-10-07
 
 ### Added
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
@ottonemo I made the changes as discussed. Please take another look.

PR #918, opened 2022-11-28
Update conda install instructions, remove environment.yml

BenjaminBossan:
Until now, the conda install instructions relied on the environment.yml in this repo. However, this file was rarely if ever updated, so it's actually not a good idea to use it. Furthermore, for the purpose of developing on skorch, there is no good reason why exactly those versions should be used, as they will likely differ from what is run on CI anyway. Therefore, this PR removes the environment.yml and updates the conda install instructions not to rely on it.

One further minor change is not to install cudatoolkit but instead pytorch-cuda, which is what is recommended on pytorch.org.


thomasjpfan (approved):
LGTM. I have always ignored the `environment.yml` for development, because of the strict versions.

PR #921, opened 2022-12-07
Fix FutureWarnings from numpy and torch

BenjaminBossan:
The following changes were made:

1. don't use `torch.testing.assert_allclose`, use `torch.testing.assert_close` instead
2. don't use `np.bool`, use `bool` instead
3. fix test that uses ragged numpy array

Regarding 2, `np.dtype('bool') == bool`, so replacing `np.bool` by `bool` should just work.

Regarding 3, numpy used to allow creation of ragged arrays, e.g. `np.array([[1], [2, 2], [3, 3, 3]])`

This does not work anymore with newer numpy versions (tested on 1.24.0rc2). To make it work, `dtype=object` has to be passed explicitly. This does not affect any skorch code directly, but it was necessary to fix a test.

### DIFF
--- a/skorch/helper.py
+++ b/skorch/helper.py
@@ -242,7 +242,7 @@ class SliceDataset(Sequence):
                 raise IndexError("SliceDataset only supports slicing with 1 "
                                  "dimensional arrays, got {} dimensions instead."
                                  "".format(i.ndim))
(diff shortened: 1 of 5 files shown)


thomasjpfan (approved):
Thank you for the updates. LGTM

PR #922, opened 2022-12-07
Add support for torch.compile

BenjaminBossan:
## Description

PyTorch announced the torch.compile feature for torch 2.0:

https://pytorch.org/get-started/pytorch-2.0

It is necessary that we modify skorch if we want to support the feature, since `torch.compile` needs to be called on the initialized modules and initialization happens inside of skorch.

Now users can create a net like `NeuralNet(..., compile=True)` and all torch modules should be compiled. By default, `compile=False`.

## Status

Torch 2.0 is not officially release yet, but the feature can already be tested.

For this, follow the install instructions here:

https://pytorch.org/get-started/pytorch-2.0/#requirements

(Note that in some places, like PyTorch nightly, the version is called 1.14.)

To actually see speed improvements, a GPU with a new architecture (Volta, Ampere) is required.

## Implementation

I opted for implementing a single new argument, `compile`, which can be set to `True`. In that case, during initialization of the module and criterion, the method `net.torch_compile` will be called on each module separately and the compiled modules will be stored on the net.

So e.g. the `net.module_` will be the compiled module. The original module would be accessible through `net.module_._orig_module`:

https://github.com/pytorch/pytorch/blob/b0bd5c4508a8923685965614c2c74e6a8c82f7ba/torch/_dynamo/eval_frame.py#L71

However, given that it's a private attribute, I assume that the compiled module should just be used, and work, like the original module and that shouldn't cause any trouble.

Furthermore, I wanted users to have the option to easily pass more arguments to `torch.compile`, which, as always, works through the use of dunder names, e.g. `NeuralNet(compile=True, compile__mode=...)`.

If users set `compile=True` with a torch version that doesn't support this, an error is raised.

Furthermore, I decided to add a public `torch_compile` method on `NeuralNet`. This method gets the module and the module name, and returns the compiled module. The idea here is that if users, for instance, want to skip compiling specific modules, they can do so by checking the module name.

## Testing

Since we introduce a new prefix with this PR, testing is a bit more comprehensive than normal. Please take a look at the `TestTorchCompile` test class, which collects all tests related to this new feature.

Since torch 2.0 is not officially released, I did not include it yet in the CI. Instead, I ran tests locally and it works. However, my GPU architecture is too old to see any speed improvements.


skorch/tests/test_net.py
```diff
@@ -3885,3 +3886,137 @@ def test_can_be_cloned(self, net):
...
+        return classifier_module
+
+    @pytest.fixture(scope='module')
+    def net_cls(self):
+        from skorch import NeuralNetClassifier
+        return NeuralNetClassifier
+
+    @pytest.fixture
+    def mock_compile(self, monkeypatch):
+        def fake_compile(module, **kwargs):
```
thomasjpfan, line 3909: Does this need to be mocked if the class is marked with a `skipif(not hasattr(torch, 'compile')`?

Aslo, if we want to have test coverage for `torch<=1.13`, then we can mock if `torch.compile` does not exist. Otherwise for `torch>=2.0`, let `torch.compile` run.

thomasjpfan (commented):
On the API for setting `compile`, I think it makes sense.

In the short term, we could have a nightly test that installs the 2.0 nightly so `torch.compile` gets called. Although, I'm fine with not testing it on CI for now and waiting for 2.0 to come out early next year. (TorchInductor supports CPU, so `compile` will generate code for CPUs as well).

skorch/tests/test_net.py
BenjaminBossan: Good point, I changed the tests to only skip those that don't use the mocked `torch.compile`.

BenjaminBossan:
> In the short term, we could have a nightly test that installs the 2.0 nightly so `torch.compile` gets called. Although, I'm fine with not testing it on CI for now and waiting for 2.0 to come out early next year.

I think there is no hurry, the number of skorch users who already use PyTorch 1.14/2.0 is probably quite small (and they can install this branch if really needed).

skorch/net.py
```diff
@@ -715,10 +729,63 @@ def _initialize_module(self, reason=None):
...
+          The compiled module if ``compile=True``, otherwise the uncompiled module.
+
+        Raises
+        ------
+        ValueError
+          If ``compile=True`` but ``torch.compile`` is not available, raise an
+          error.
+
+        """
+        # TODO: adjust docstring once we no longer PyTorch versions without compile
```
thomasjpfan, line 773:
```suggestion
        # TODO: adjust docstring once we no longer support PyTorch versions without compile
```

thomasjpfan (approved):
I tested this locally with PyTorch 2.0. LGTM

BenjaminBossan:
Thanks @thomasjpfan for testing. I'll leave this PR open until PyTorch 2.0 is officially released, update the CI, run another round of tests and merge if everything passes.

PR #923, opened 2022-12-12
Bugfix: tests that fail in sklearn 1.2

BenjaminBossan:
The reason is that the `normalize` parameter was removed from `LinearRegression`.


thomasjpfan (approved):
LGTM

Issue #925, opened 2022-12-20
Unhelpful error message when setting invalid parameter with `set_params`


githubnemo:
Minimal example:

```python
net.set_params(this_does_not_exist=True)
```

Expected result:
```
Invalid parameter 'this_does_not_exist' for estimator ...
Valid parameters are: ['module', 'criterion', ...]
```

Actual result:
```
Invalid parameter 'this_does_not_exist' for estimator ...
Valid parameters are: <generator object NeuralNet._get_param_names.<locals>.<genexpr> at 0x7f1cf9d1da10>.
```

Exception raising code of sklearn (version 1.2.0) looks like this:
```
    203             if key not in valid_params:
    204                 local_valid_params = self._get_param_names()
--> 205                 raise ValueError(
    206                     f"Invalid parameter {key!r} for estimator {self}. "
    207                     f"Valid parameters are: {local_valid_params!r}."
```

The fix is probably just to convert the generator expression to a list comprehension.

BenjaminBossan:
Closed via #926

PR #926, opened 2022-12-20
Address #925

githubnemo:
Changing the `_get_param_names` method to return a list instead of a generator to fix the exception error message when passing unknown parameters to `set_params`. Before the error message just included the generator `repr`-string as the list of possible parameters. Now the string contains the possible parameter names instead.


BenjaminBossan (changes requested):
Looks good, thanks for the fix. Performance-wise, this should not make a big difference, so using lists is good.

Could you please add tests along the lines of what you showed in the [issue](https://github.com/skorch-dev/skorch/issues/925)?

BenjaminBossan (approved):
Nice, good catch with the sklearn 0.2* check.

PR #927, opened 2022-12-22
Bugfixes: Bugs stemming from the use of non-standard modules and criteria

BenjaminBossan:
These bugfixes were originally provided in #912 but are now moved to their own PR.

1. Inferring the predict nonlinearity made the assumption that a `net.criterion_` exists. However, this might not always be the case. Now, this function works even if the criterion attribute has a different name. Moreover, when more than one criterion is defined, the identity function will be returned.
2. The second bug is that in `check_is_fitted`, we made the check dependent on the `module_` attribute. Again, we should not assume that it always exists, as users may define different names. Now, those custom names will be checked, and only if those don't exist is it assumed that the `module_` attribute should exist.
3. The `helper.py` module now defines an `__all__` attribute. This seems to be necessary for sphinx to build documentation for objects that are imported to, but not defined in, `helper.py`. Specifically, this is the case for the `parse_args` CLI feature.
4. Fixed a small bug in `on_grad_computed`, which has `training=False` as default argument but during training, it should be set to `True`. Thankfully, it had no consequences in our code since it's only used by `GradientNormClipping`, which doesn't check that argument.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixes a couple of bugs related to using non-default modules and criteria (#???)
+
 ## [0.12.1] - 2022-11-18
 
 ### Changed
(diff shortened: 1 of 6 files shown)


thomasjpfan (approved):
LGTM

PR #928, opened 2023-01-04
Fixed imports and Checked Notebooks

sawradip:
All notebooks are checked - working.
Added some helpful comment.


BenjaminBossan (changes requested):
Thanks a lot working on this. I have a few comments, but I'll add them here, instead of on the notebook diff, because the diff view for notebooks is not really ergonomic.

1 If the import fails, you print the line:

> If not already installed, you can install skorch by running 'pip install skorch'

I think it's not necessary, if someone runs the notebook locally, we can assume that skorch is installed. If more dependencies are needed (e.g. gpytorch), we already document that. Therefore, I think you can remove that line and just `pass`.

2 Furthermore, you added a comment `# Installation`. Could you please extend it to be: `# Installation on Google colab`? This might not be clear to some users.

3 For some of the notebooks, I could see that they were not run from top to bottom. E.g. in `Basic_Usage.ipynb`, after cell 12 comes cell 15, then 11. Some cells have not run at all. Before checking in a notebook, please ensure that it runs from top to bottom. This is important for reproducibility.

4 In `Transfer_Learning.ipynb`, we get a bunch of warnings:

> UserWarning: This DataLoader will create 4 worker processes in total. Our suggested max number of worker in current system is 2, which is smaller than what this DataLoader is going to create. Please be aware that excessive worker creation might get DataLoader running slow or even freeze, lower the worker number to avoid potential slowness/freeze if necessary.

They are really distracting, so let's get rid of them. Either suppress the warning, or, better, set the number of workers to 2 in cell 9:

```python
    iterator_train__num_workers=2,
    iterator_valid__num_workers=2,
```

5 You deleted two notebooks, `CORA-geometric.ipynb` and `MNIST-torchvision.ipynb`. Could you please restore them?

6 For some notebooks, the diff view is completely broken: `Gaussian_Processes.ipynb, Hugging_Face_Finetuning.ipynb, Hugging_Face_Model_Checkpoint.ipynb, Transfer_Learning.ipynb`. This makes it difficult for me to check if you changed something besides the install instruction. Could you please say if you did?

7 In `Basic_Usage.ipynb`, you added `import os`in cell 35 but `os` is never used.

8 Finally, you checked in a bunch of files that are not required, in the `datasets` and `runs` folders, could you please remove them?

sawradip:
Thank you for your reviews, and I am fixing those. I thing I have faced and think should be fixed is :

Screenshot (3)

If GPU is connected. This error arises. 

I am seeing, changing this line
``` samples = samples.detach().numpy()   ```
to this line
``` samples = samples.cpu().detach().numpy()   ```
will make it work for both CPU and GPU.

And as you asked if I changed anything in those notebooks, I don't remember changing anything that effects the code. One or two comments might have been added, but as some were already there, it is tough for me also to figure out as `diff` is not a particularly good view for notebook. You can be sure taking a look at the comments.

**Review by BenjaminBossan — CHANGES_REQUESTED** [review]
Thank you for implementing the change so quickly, this looks very good.

I saw that you added some comments, those that I found looked very reasonable (but again, hard to find with the diff). So let's keep it that way.

I only saw some minor issues in `Transfer_Learning.ipynb` being left. In cell 9, could you please remove the line

`iterator_valid__shuffle=True,`

It was already wrong before, so nothing to do with your PR, but it's incorrect, so let's fix it.

In the same notebook, for some reason, there are still warnings about `This DataLoader will create 4 worker processes in total`, even though the number of workers was reduced to 2. When I run the notebook on colab, I don't get any warnings. Could you please try to run again?

Finally, could you please add an entry to `CHANGES.md` about fixing the install command in colab?

**Comment by sawradip** [comment]
- All the notebooks have been updated according to the code review. 
- All notebooks have been run top-to-bottom in Google Colab

As this is my first time updating a PR, can you guide me if I need to do anything now, I have commited all the changes to my corresponding branch.

**Review by BenjaminBossan — APPROVED** [review]
@sawradip There is nothing more you need to do, after pushing to your branch, the PR is updated automatically.

Thanks for implementing the remaining fixes. the `Transfer_Learning.ipynb` notebook looks much better now.

I'm ready to merge, but will leave this PR open for a few days in case @ottonemo or @thomasjpfan want to add anything.

**Comment by BenjaminBossan** [comment]
Since there were no more comments, I assume it means we're good to go :)

**Comment by BenjaminBossan** [comment]
@sawradip 

Just found this suggestion by a google colab PM (ckperry) on reddit:

https://www.reddit.com/r/MachineLearning/comments/114hphp/comment/j8xxnpp/

Conclusion: There is no official way to check if a notebook is running on colab. The suggested snippet is basically doing the same thing as how you changed the notebooks, so I think we're good here.

**Comment by sawradip** [comment]
Awesome! It was a great experience. I will like to contribute more later.

**Comment by sawradip** [comment]
While working on a PR in Openvino, I found another way to verify. This is also not official, but still leaving here, if ever is necessary.

```
from IPython.core.getipython import get_ipython

# Checking if running in Google Colab
IN_COLAB = 'google.colab' in str(get_ipython())
```

BenjaminBossan:
Thanks, good to know. I prefer the current method, but it's good to have a backup.

PR #929, opened 2023-01-17
changing from must to most.

Nobody4591:
Before:
If you would like to use the must recent additions to skorch or help development, you should install skorch from source.

After:
If you would like to use the most recent additions to skorch or help development, you should install skorch from source.

I don't wanna make PR, for just changing 'u' with 'o' but i don't know what else i can do. lol


BenjaminBossan (approved):
It's fine to submit a PR for fixing typos*, thanks :)

*as long as you don't submit 10 PRs with one typo each

PR #934, opened 2023-02-24
update for neptune 1.0

kshitij12345:


skorch/tests/callbacks/test_logging.py
```diff
@@ -43,12 +43,7 @@ def neptune_logger_cls(self):
 
     @pytest.fixture
     def neptune_run_object(self):
-        try:
-            # neptune-client=0.9.0+ package structure
-            import neptune.new as neptune
-        except ImportError:
-            # neptune-client>=1.0.0 package structure
-            import neptune
+        import neptune
```
kshitij12345, line 46: We update `requirements.txt` to install latest `neptune` so no need to guard the imports for older version.

kshitij12345 (commented):
https://github.com/skorch-dev/skorch/blob/909f4d56a6d1a54391cfae208ba9f431b9da798c/skorch/callbacks/logging.py#L260-L265

Update based on the decided approach.

BenjaminBossan:
@kshitij12345 Thanks for the PR. Are you with Neptune?

kshitij12345:
Hi @BenjaminBossan,

Yes. I am with Neptune. Currently the PR is draft and have asked other members to do a preliminary review. 

Once it is ready, I will mark this ready and ping you. Does that work?

Thanks!

BenjaminBossan:
Okay, great, that works. I only asked to ensure that folks at Neptune are okay with dropping support for the old client.

skorch/callbacks/logging.py
```diff
@@ -231,18 +231,27 @@ def on_epoch_end(self, net, **kwargs):
...
-            self._metric_logger['train/epoch/event_lr'].log(net.history[:, 'event_lr'])
+            self._metric_logger['train/epoch/event_lr'].append(net.history[:, 'event_lr'])
         except KeyError:
             pass
         if self.close_after_train:
-            self.run.stop()
+            try:  # >1.0 package structure
+                from neptune.handler import Handler
+            except ImportError:  # <1.0 package structure
+                from neptune.new.handler import Handler
```
twolodzko, line 241: This can be just

```
from neptune.handler import Handler
```

given that the dependency is v >=1.0.0

kshitij12345: This is actual logging code which should be compatible with both pre and post 1.0. 

Other update where we assume 1.0 structure is in test (where we have updated the requirements to install latest version).

twolodzko: But the requirements say `neptune` hence >=1.0.0, so there is no backwards compatibility assumed.

skorch/callbacks/logging.py
```diff
@@ -86,11 +86,11 @@ class NeptuneLogger(Callback):
     Examples
     --------
     $ # Install Neptune
-    $ python -m pip install neptune-client
+    $ python -m pip install neptune
 
     >>> # Create a Neptune run
-    >>> import neptune.new as neptune
-    >>> from neptune.new.types import File
+    >>> import neptune as neptune
```
normandy7, line 92:
```suggestion
    >>> import neptune
```

skorch/callbacks/logging.py
normandy7, line 1: Let's just change the example run on line 84 to: https://app.neptune.ai/o/common/org/skorch-integration/e/SKOR-32/all

normandy7 (approved):

BenjaminBossan:
I got an error in a CI run that seems to be related to Neptune:

```
>       assert isinstance(
            neptune_run_object.get_structure()['training']['model']['checkpoint'],
            FileSet,
        )
E       AssertionError: assert False
E        +  where False = isinstance(<neptune.attributes.file_set.FileSet object at 0x7f942dcdd850>, <class 'neptune.new.attributes.file_set.FileSet'>)
```

https://github.com/skorch-dev/skorch/actions/runs/4436004355/jobs/7783884283#step:6:107

My first guess would be that it's caused by a new neptune version. The installed version is `neptune-client-1.0.2`. It seems this PR might fix the issue, is that right?

kshitij12345:
Yes, this is due to new version and  this PR should fix the issue.

BenjaminBossan:
Great, thanks for letting me know. I don't want to rush this PR, but since CI is currently broken, do you think it will be done soon? Else, I would restrict the neptune version to make CI pass again.

kshitij12345:
I think the PR is ready. @aleksanderWWW for a final review.

Thank you!

AleksanderWWW (approved):
LGTM @kshitij12345 :) The updated imports should resolve the CI issue

kshitij12345:
@BenjaminBossan this should be ready for your review and merging :)

BenjaminBossan (changes requested):
Thanks a lot for the update. Very nice results on Neptune.

Overall, this is almost good to go, just a few comments, please take a look.

BenjaminBossan (approved):
Thanks a lot for keeping the Neptune logger up to date and improving the docs.

PR #936, opened 2023-03-04
Add option to keep the dictionary/SliceDict input packed and skip `_merge_x_and_fit_params` call

josarago:
First things first, thanks for building this awesome open-source library. It's really great.

I'm creating this untested pull request as a suggestion to implement a new feature:

Give the user the option to NOT unpack input variables that are dictionaries. I would actually argue that dictionary should never be unpacked by skorch. If someone uses a dict like structure to store the data, they probably want to keep it like that.

In some situations, it is convenient to be able to pass a dictionary/SliceDict as an argument to the `forward()` call of the pytorch Module. Below I created a minimal example, which shows the issue.
The expected behavior with dict/SliceDict was explained here https://github.com/skorch-dev/skorch/issues/672.

Minimal example:

```
import torch
from torch import nn
from skorch.helper import SliceDict
from skorch import NeuralNetRegressor

vocab_sizes = {
    "user_id": 1_000,
    "user_gender": 7,
    "user_country": 200,
    "item_id": 5_000,
    "item_category": 150,
}


class SomeModule(nn.Module):
    def __init__(self, embedding_dim, vocab_sizes):
        super().__init__()
        self.embedding_dim = embedding_dim
        self.vocab_sizes = vocab_sizes
        self.embeddings = nn.ModuleDict()
        # embedding inputs
        for name, vocab_size in self.vocab_sizes.items():
            self.embeddings[name] = nn.Embedding(
                vocab_size, embedding_dim, sparse=False
            )

    def forward(self, input_dict):
        x_embed_inputs = dict()
        for name in self.vocab_sizes.keys():
            x_embed_inputs[name] = torch.squeeze(
                self.embeddings[name](input_dict[name]), dim=1
            )
        x_embed_user = torch.sum(
            torch.stack(
                [
                    x_embed_inputs[name]
                    for name in self.vocab_sizes.keys()
                    if name.startswith("user_")
                ]
            ),
            dim=0,
        )
        x_embed_item = torch.sum(
            torch.stack(
                [
                    x_embed_inputs[name]
                    for name in self.vocab_sizes.keys()
                    if name.startswith("item_")
                ]
            ),
            dim=0,
        )
        dot = torch.mul(x_embed_user, x_embed_item).sum(dim=1)
        return dot


if __name__ == "__main__":
    embedding_dim = 3
    n_samples = 1_000
    input_dict = SliceDict(
        **{
            name: torch.randint(vocab_size, (n_samples, 1))
            for name, vocab_size in vocab_sizes.items()
        }
    )
    print(input_dict)
    y = torch.rand(1_000, 1)
    module = SomeModule(embedding_dim=embedding_dim, vocab_sizes=vocab_sizes)
    out = module(input_dict)
    print(out.shape)

    net = NeuralNetRegressor(
        module=SomeModule(embedding_dim=embedding_dim, vocab_sizes=vocab_sizes),
        criterion=torch.nn.MSELoss(),
        lr=0.01,
    )
    net.fit(input_dict, y)
```

In this case, I really want to flexibly add and remove some of the inputs from `vocab_sizes`: some columns might be one-hot-encoded instead of passed to an embedding layer etc... (Outside of this toy example, `vocab_sizes` would not be hardcoded but computed from a training set).

The reason why this cannot work with the current implementation is because the dictionary (or any Mapping type input) is always unpacked when passed to the `infer()` method.

A quick way to go around this would be to make the unpacking of the dictionary optional the way I implemented it in this PR. 
I'm not sure that's the best way but happy to try to help if the team thinks this would be a good feature to add.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -283,6 +283,7 @@ class NeuralNet:
             warm_start=False,
             verbose=1,
             device='cpu',
+            unpack_dict_input=True,
             **kwargs
     ):
         self.module = module
@@ -300,6 +301,7 @@ class NeuralNet:
         self.warm_start = warm_start
         self.verbose = verbose
         self.device = device
+        self.unpack_dict_input = unpack_dict_input
 
         self._check_deprecated_params(**kwargs)
         history = kwargs.pop('history', None)
@@ -1426,7 +1428,7 @@ class NeuralNet:
 
         """
         x = to_tensor(x, device=self.device)
-        if isinstance(x, Mapping):
+        if isinstance(x, Mapping) and self.unpack_dict_input:
             x_dict = self._merge_x_and_fit_params(x, fit_params)
             return self.module_(**x_dict)
         return self.module_(x, **fit_params)


BenjaminBossan:
Thanks for the suggestion. I see what you want to achieve on a technical level, but I haven't understood the use case that would make it necessary. For example, I could make your example run with this change:

```python
-    def forward(self, input_dict):
+    def forward(self, **input_dict):
```

I can imagine 2 scenarios where this would not work:

1. Multiple input dicts

If you have multiple input dicts, unpacking them can be tricky. The syntax `some_function(**dict_0, **dict_1)` works in recent Python versions but if keys overlap, it will fail. Still, the PR you suggested would not solve it.

2. Using a pre-existing module that doesn't use `**` in its `forward` method.

In that case, I would recommend wrapping it:

```python
from foo import FooModule

class MyFooModule(FooModule):
    def forward(self, **input_dict):
        return super().forward(input_dict)
```

josarago:
Thanks for your answer!
Maybe I am misunderstanding you but when you say:

> For example, I could make your example run with this change:
> ```
> -    def forward(self, input_dict):
> +    def forward(self, **input_dict):

I tried to run the code with the suggested change. the skorch part of the code does tun but not the direct forward 
call unless I also make the change:
```
-  out = module(input_dict)
+  out = module(**input_dict)
```
Is that correct or did I miss something?

BenjaminBossan:
> Is that correct or did I miss something?

Yes, that's correct, you have to call it either this way or that way.

> Isn't it generally the case?

I wouldn't think so, my personal impression is that it's more common to have the arguments separately, instead of passing a dict and unpacking it manually inside the `forward` call. Just as an example, this is how the `forward` method for many `transformers` models looks like:

https://github.com/huggingface/transformers/blob/7c393181366f868ea4cac7fc9da8af5c5e7972ee/src/transformers/models/bert/modeling_bert.py#L918-L933

If you have an example of a common lib that expects the dict not to be unpacked, please let us know.

In general, I often find that if I want to use a module from another lib, it requires a small amount of tinkering, e.g. because you want to run something on top of its output or only want to return part of the output. Therefore, if I'm already doing that, I need to write a custom `nn.Module` where I may also change the way `forward` is called, if necessary. But I think it's not too much to ask of a user -- after all, subclassing `nn.Module` is the bread and butter of PyTorch. This is why I would tend not to add the option to prevent unpacking, unless there is a common library that requires that.

josarago:
Sorry for the delay. I think the pattern I was describing is fairly common for recommender systems, for instance [here](https://pytorch.org/blog/introducing-torchrec/). I also saw the release of [TensorDict](https://github.com/pytorch-labs/tensordict) which I think goes in a similar direction. Having to name explicitly each input can be inconvenient in these cases. I haven't dived deep enough into the code base to fully understand what would motivate implementing it in a way that requires constraining the inputs like this. Unless I am mistaken, what I am proposing is to offer more flexibility for inputs. I don't see the downsides but I am probably just missing something

BenjaminBossan:
I'm not familiar with torchrec, so my understanding could be wrong. From what it looks like to me, it defines its own special data types like [`JaggedTensor`](https://github.com/pytorch/torchrec/blob/09c3c09f985ddd4f07d5d6e3bc8c52037db57c02/torchrec/sparse/jagged_tensor.py#L168) or [`KeyedJaggedTensor`](https://github.com/pytorch/torchrec/blob/09c3c09f985ddd4f07d5d6e3bc8c52037db57c02/torchrec/sparse/jagged_tensor.py#L682), which are not dicts and would not be unpacked. Most likely, this data type will run into other issues with skorch, so I'm not sure if torchrec and skorch could be used together without a lot of tinkering.

Regarding TensorDict, this seems to be a kinda niche library, not sure if it's needed, given that we have `SliceDict` in skorch.

> Unless I am mistaken, what I am proposing is to offer more flexibility for inputs. I don't see the downsides but I am probably just missing something

As you say, I am reluctant to make that change :) The main reason is that this is a hard to explain parameter on top of all the existing parameters, which could cause confusion, especially since it changes a very technical aspect of skorch. If it were difficult to achieve the outcome otherwise, I could see the upsides exceeding the downsides. However, as I explained earlier, it is quite easy to make adjustment to make this work by wrapping the module. And I would highly suspect most users would discover that way much quicker than the proposed additional argument.

Still it's good to have this kind of discussion. I see why your use case would benefit, so please don't take this as dismissing your concern.

josarago:
> Still it's good to have this kind of discussion. I see why your use case would benefit, so please don't take this as dismissing your concern.

Not at all! I really appreciate your thoughtfulness and I am learning a lot.

About Torchrec I just meant they have a similar need for programmatically creating inputs and it seems natural to use a dict-like data structure for that purpose. 

> And I would highly suspect most users would discover that way much quicker than the proposed additional argument.

To be clear the change I proposed was the smallest change I could think of that seems to not be a breaking one. I agree it is not ideal and I had the impression that the best solution was to not have more constraints for the modules' inputs in skorch that in pytorch. I guess I was hoping that we could remove the mandatory unpacking of `Mapping` objects altogether but I couldn't really assess the scope of that change and clearly it would be a breaking change.  

> Regarding TensorDict, this seems to be a kinda niche library, not sure if it's needed, given that we have SliceDict in skorch

I think TensorDict is just really new, but since it was [developed by folks at Meta](https://pytorch-labs.github.io/tensordict/), I was assuming it would soon become a new standard/official way to deal with this. 

> Most likely, this data type will run into other issues with skorch, so I'm not sure if torchrec and skorch could be used together without a lot of tinkering.

Anyways I'm just sharing an opinion and at the end of the day, you know what's best for skorch. Feel free to close the PR.

BenjaminBossan:
> Anyways I'm just sharing an opinion and at the end of the day, you know what's best for skorch. Feel free to close the PR.

I appreciate your input and took this feedback to create a PR (#941) to document this better, thanks.

> I think TensorDict is just really new, but since it was [developed by folks at Meta](https://pytorch-labs.github.io/tensordict/), I was assuming it would soon become a new standard/official way to deal with this.

Let's keep this in mind, if it gains traction and people want to use it with skorch, we can resurrect your proposal.

PR #947, opened 2023-03-31
Fix accelerate bug, document, add scripts

### LINKED ISSUE #944: AccelerateMixin error with RandomizedSearchCV
Hi, 

Thanks a lot for the great tool! 

I tried the recently added HuggingFace Accelerate integration. I want to perform hyper-parameters optimization using Skorch with Accelerate + ScikitLearn RandomizedSearchCV. 

However, it seems that they do not play nicely at scoring time by the RandomizedSearchCV.

Reproducible example named `skorch_accelerate_issue.py` :

```python
import torch
import numpy as np
from skorch.hf import AccelerateMixin
from accelerate import Accelerator 
from sklearn.model_selection import RandomizedSearchCV
import torch.nn as nn

# FYI: Accelerate also requires the `transformers` packages from HuggingFace

# Generate data
X, y = make_classification(10_000, 100, n_informative=5, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

# PyTorch module
class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

        self.dense0 = nn.Linear(100, 2)
        self.nonlin = nn.ReLU()

    def forward(self, X):
        X = self.dense0(X)
        X = self.nonlin(X)
        return X

# Skorch wrapper
class AcceleratedNeuralNetClassifier(
    AccelerateMixin, 
    NeuralNetClassifier
):
    """NeuralNetClassifier with HuggingFace Accelerate support"""
accelerator = Accelerator()
model = AcceleratedNeuralNetClassifier(
    MyModule,
    accelerator=accelerator,
)

# HPO
rs = RandomizedSearchCV(
    estimator=model,
    param_distributions={
        "lr": [0.0001, 0.001, 0.01, 0.1],
        "batch_size": [10, 20, 30, 40],
    },
    n_iter=10,
    scoring="average_precision",
    n_jobs=1,
    refit=False,
    cv=2,
    verbose=1,
)
rs.fit(X, y)
```

Accelerate config to run this script on 2 GPUs on the same machine: 
```
(/home/razorin/conda_envs/backup) razorin@ML:~/tmp$ accelerate config                                                                         

In which compute environment are you running?
This machine

Which type of machine are you using?
multi-GPU

How many different machines will you use (use more than 1 for multi-node training)? [1]: 
1 

Do you wish to optimize your script with torch dynamo?[yes/NO]:NO
Do you want to use DeepSpeed? [yes/NO]: NO
Do you want to use FullyShardedDataParallel? [yes/NO]: NO
Do you want to use Megatron-LM ? [yes/NO]: NO 

How many GPU(s) should be used for distributed training? [1]:2 

What GPU(s) (by id) should be used for training on this machine as a comma-separated list? [all]:0,5 

Do you wish to use FP16 or BF16 (mixed precision)? no 

accelerate configuration saved at /home/razorin/.cache/huggingface/accelerate/default_config.yaml
```

I ran the code using: `accelerate launch skorch_accelerate_issue.py`

And here is the error:

```
The following values were not passed to `accelerate launch` and had defaults used instead: 
--dynamo_backend was set to a value of no 
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`. 

Fitting 2 folds for each of 10 candidates, totalling 20 fits 
Fitting 2 folds for each of 10 candidates, totalling 20 fits 
epoch    train_loss    valid_acc    valid_loss     dur
-------  ------------  -----------  ------------  ------ 
1        5.4880       0.6060        4.5132  0.3304 
2        4.0846       0.6520        3.3801  0.2789 

/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/model_selection/_validation.py:778: UserWarning: Scoring failed. The score on this train-test partition for these parameters will be set to nan. Details:                                                                                                                           

Traceback (most recent call last): File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/model_selection/_validation.py", line 767, in _score
   scores = scorer(estimator, X_test, y_test) 
File "/home/razorin/conda_envs/backup/lib/python3.9/sitepackages/sklearn/metrics/_scorer.py", line 234, in __call__                                                          
   return self._score( 
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_scorer.py", line 399, in _score                                                            
   return self._sign * self._score_func(y, y_pred, **self._kwargs) 
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_ranking.py", line 234, in average_precision_score 
   return _average_binary_score( 
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_base.py", line 75, in _average_binary_score                                                
   return binary_metric(y_true, y_score, sample_weight=sample_weight)
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_ranking.py", line 207, in _binary_uninterpolated_average_precision
   precision, recall, _ = precision_recall_curve(
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_ranking.py", line 878, in precision_recall_curve                                           
   fps, tps, thresholds = _binary_clf_curve( 
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/metrics/_ranking.py", line 751, in _binary_clf_curve                                                
   check_consistent_length(y_true, y_score, sample_weight) 
File "/home/razorin/conda_envs/backup/lib/python3.9/site-packages/sklearn/utils/validation.py", line 397, in check_consistent_length                                          
  raise ValueError( 

ValueError: Found input variables with inconsistent numbers of samples: [5000, 2500]
```

FYI, when training starts, I can see that the two GPUs are indeed occupied. 
Many thanks in advance for your help.

BenjaminBossan:
Partly resolves #944

There is an issue with using skorch in a multi-GPU setting with accelerate. After some searching, it turns out there were two problems:

1. skorch did not call `accelerator.gather_for_metrics`, which resulted in `y_pred` not having the correct size. For more on this, consult the [accelerate docs](https://huggingface.co/docs/accelerate/quicktour#distributed-evaluation).

2. accelerate has an issue with being deepcopied, which happens for instance when using `GridSearchCV`. The problem is that some references get messed up, resulting in the `GradientState` of the `Accelerator` instance and of the dataloader to diverge. Therefore, the accelerator did not "know" when the last batch was encountered and was thus unable to remove the dummy samples added for multi-GPU inference.

The fix for 1. is provided in this PR. For 2., there is no solution in skorch, but a possible (maybe hacky) fix is suggested in the docs. The fix consists of writing a custom `Accelerator` class that overrides `__deepcopy__` to just return self. I don't know enough about accelerate internals to determine if this is a safe solution or if it can cause more issues down the line, but it resolves the issue.

Since reproducing this bug requires a multi-GPU setup and running the scripts with the accelerate launcher (`accelerate lauch <script.py>`), it cannot be covered by normal unit tests. Instead, this PR adds two scripts to reproduce the issue. With the appropriate hardware, they can be used to check the solution.

### DIFF
--- a/docs/user/huggingface.rst
+++ b/docs/user/huggingface.rst
@@ -80,8 +80,8 @@ functions and classes creating copies are:
   not `Pipeline`).
 
 When using any of those in a multi-GPU setup with :class:`.AccelerateMixin`, you
-may encounter errors. A possible fix is to prevent the :class:`.AccelerateMixin`
-instance from being copied (or, to be precise, deep-copied):
+may encounter errors. A possible fix is to prevent the ``Accelerator`` instance
+from being copied (or, to be precise, deep-copied):
 
 .. code:: python
 
@@ -117,6 +117,11 @@ considerably because inference is called once for each scorer, even if the
 results are always the same. A possible solution to this is to write your own
 scoring callback that records multiple scores to the ``history`` using a single
 inference call.
+
+Moreover, if your training relies on the training history on some capacity, e.g.
+because you want to early stop when the validation loss stops improving, you
+should use :class:`.DistributedHistory` instead of the default history. More
+information on this can be found :ref:`here <dist-history>`.
 
 Tokenizers
 ----------
--- a/docs/user/huggingface.rst
+++ b/docs/user/huggingface.rst
@@ -14,6 +14,9 @@ accelerate_ to skorch. E.g., this allows you to use mixed precision training
 (AMP), multi-GPU training, training with a TPU, or gradient accumulation. For the
 time being, this feature should be considered experimental.
 
+Using accelerate
+^^^^^^^^^^^^^^^^
+
 To use this feature, create a new subclass of the neural net class you want to
 use and inherit from the mixin class. E.g., if you want to use a
 :class:`.NeuralNet`, it would look like this:
@@ -62,6 +65,41 @@ To install accelerate_, run the following command inside your Python environment
     Therefore, if your optimizer requires that (e.g.
     :class:`torch.optim.LBFGS`), you cannot use accelerate.
 
+
+Caution when using a multi-GPU setup
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There is a known issue when using accelerate in a multi-GPU setup *if copies of
+the net are created*. In particular, be aware that sklearn often creates copies
+under the hood, which may not immediately obvious to a user. Examples of
+functions and classes creating copies are:
+
+- `GridSearchCV`, `RandomizedSearchCV` etc.
+- `cross_validate`, `cross_val_score` etc.
+- `VotingClassifier`, `CalibratedClassifierCV` and other meta estimators (but
+  not `Pipeline`).
+
+When using any of those in a multi-GPU setup with :class:`.AccelerateMixin`, you
+may encounter errors. A possible fix is to prevent the :class:`.AccelerateMixin`
+instance from being copied (or, to be precise, deep-copied):
+
+.. code:: python
+
+    class AcceleratedNet(AccelerateMixin, NeuralNet):
+        pass
+
+    class MyAccelerator(Accelerator):
+        def __deepcopy__(self, memo):
+            return self
+
+    accelerator = MyAccelerator()
+    net = AcceleratedNet(..., accelerator=accelerator)
+    # now grid search et al. should work
+    gs = GridSearchCV(net, ...)
+    gs.fit(X, y)
+
+Note that this is a hacky solution, so monitor your results closely to ensure
+nothing strange is going on.
 
 Tokenizers
 ----------
--- a/docs/user/huggingface.rst
+++ b/docs/user/huggingface.rst
@@ -101,6 +101,23 @@ instance from being copied (or, to be precise, deep-copied):
 Note that this is a hacky solution, so monitor your results closely to ensure
 nothing strange is going on.
 
+There is also a problem with caching not working correctly in multi-GPU
+training. Therefore, if using a scoring callback (e.g.
+:class:`skorch.callbacks.EpochScoring`), turn caching off by passing
+``use_caching=False``. Be aware that when using
+:class:`skorch.NeuralNetClassifier`, a scorer for accuracy on the validation set
+is added automatically. Caching can be turned off like this:
+
+.. code:: python
+
+    net = NeuralNetClassifier(..., valid_acc__use_caching=False)
+
+When running a lot of scorers, the lack of caching can slow down training
+considerably because inference is called once for each scorer, even if the
+results are always the same. A possible solution to this is to write your own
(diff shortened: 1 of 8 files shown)


docs/user/huggingface.rst
```diff
@@ -63,6 +66,63 @@ To install accelerate_, run the following command inside your Python environment
...
+from being copied (or, to be precise, deep-copied):
+
+.. code:: python
+
+    class AcceleratedNet(AccelerateMixin, NeuralNet):
+        pass
+
+    class MyAccelerator(Accelerator):
+        def __deepcopy__(self, memo):
+            return self
```
thomasjpfan, line 93: I'm trying to come up with way this can break, but it seems okay. From my understanding, here is how I see the behavior:

- `GridSearchCV.fit` will `clone` the `NeuralNet`, and the PyTorch Module gets deep copied and leave the `Accelerator` in place.
- During `initialize`, the Accelerator will be reconfigured to point to the new Module (If there was a way to "reset the Accelerator" that would be nice too)
- Profit?

Maybe passing an stateful Accelerator object into `AccelerateMixin` was a bad design choice. For `NeutralNet`, I usually prefer passing in the Module class and not an instance. For `AccelerateMixin`, we can allow `accelerator__mixed_precision="fp16"`.

BenjaminBossan: > From my understanding, here is how I see the behavior: [...]

Sorry, I don't quite understand, is this behavior without the `__deepcopy__` patch or with?

> * If there was a way to "reset the Accelerator" that would be nice too

AFAIK, there is no public way to do that (see [here](https://github.com/skorch-dev/skorch/blob/c909e14a1d66de1de23a4619a3c3dc0ef5498a44/skorch/tests/test_hf.py#L534)). The `Accelerator` was always intended to be run in a script that has a single training loop, re-using is normally out of scope. Devs told me there should only be one instance, and it should be created as early as possible in the script.

> Maybe passing an stateful Accelerator object into `AccelerateMixin` was a bad design choice.

Yes, possibly the pattern you suggest should have been the default, although it would not solve the issue with there only ever being a single `Accelerator` instance. In general, though, I believe we could add the suggested pattern without BC breaking, allowing users to pass either instances or classes like with `module`.

thomasjpfan (approved):
This is PR documenting a `deepcopy` workaround and fixing up distrusted evaluation for Accelerate. If the `deepcopy` workaround has no issues, then it can be included directly into `AccelerateMixin` later.

Unfortunately, I am not able to test multi-GPU right now, but the code looks okay.

LGTM

docs/user/huggingface.rst
BenjaminBossan: Got it. I created an issue, #960, for tracking what we discussed about instantiating the `Accelerator`.

Issue #952, opened 2023-04-06
EarlyStopping callback issue with Skorch + HuggingFace Accelerate


Raphaaal:
Hi, 

I am trying to use the `EarlyStopping` callback with `Skorch` + HuggingFace's `accelerate` on two GPUs.

`accelerate` config:

```yaml
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: MULTI_GPU
downcast_bf16: 'no'
dynamo_config: {}
fsdp_config: {}
gpu_ids: 0,1
machine_rank: 0
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'no'
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

IIUC `accelerate`, the model is copied on both GPUs and each GPU will receive a different slice of the batch. Is this assumption correct?

This can lead to inconsistencies where only one of the two model copies has its early stopping callback triggered, leaving the remaining process hanging. In this case, the training stalls, possibly because one process is waiting for the other (that has been interrupted due to early-stopping).

It is a bit tricky to reproduce, but here is my attempt:

```python
import os
os.environ["CUBLAS_WORKSPACE_CONFIG"]=":4096:8"

import torch
import torch.nn as nn
import numpy as np
import random
from skorch import NeuralNetClassifier
from sklearn.datasets import make_classification
from accelerate import Accelerator
from skorch.hf import AccelerateMixin
from skorch.callbacks import EarlyStopping
from skorch.dataset import ValidSplit


# Reproducibility
SEED = 42
def seed_everything(seed=42):
    torch.manual_seed(seed)
    random.seed(seed)
    np.random.seed(seed)
    torch.use_deterministic_algorithms(True)
    torch.backends.cudnn.benchmark = False
    torch.backends.cudnn.deterministic = True
    torch.cuda.manual_seed(seed)
    torch.cuda.manual_seed_all(seed)


class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.dense0 = nn.Linear(100, 50)
        self.dense1 = nn.Linear(50, 25)
        self.dense2 = nn.Linear(25, 2)
        self.nonlin = nn.Softmax(dim=-1)

    def forward(self, X):
        X = self.dense0(X)
        X = self.dense1(X)
        X = self.dense2(X)
        X = self.nonlin(X)
[...]

PR #955, opened 2023-04-14
Add DistributedHistory for multi-gpu training

BenjaminBossan:
## Description

(this is more or less copied from the docs)

When training a net in a distributed setting, e.g. when using `torch.nn.parallel.DistributedDataParallel`, directly or indirectly with the help of `AccelerateMixin`, the default history class should not be used. This is because each process will have its own history instance with no syncing happening between processes. Therefore, the information in the histories can diverge. When steering the training process through the histories, the resulting differences can cause trouble. When using early stopping, for instance, one process could receive the signal to stop but not the other.

`DistributedHistory` will take care of syncing the distributed batch information across processes, which will prevent the issue just described.

This class needs to be initialized with a distributed store provided by PyTorch (https://pytorch.org/docs/stable/distributed.html#distributed-key-value-store). I have only tested `torch.distributed.TCPStore` so far, but `torch.distributed.FileStore` should also work. The `DistributedHistory` also needs to be initialized with its rank and the world size (number of processes) so that it has all the required information to perform the syncing. When using `accelerate`, that information can be retrieved from the `Accelerator` instance.

## Comments

1. Even though the batch information, which is split across processes, is synced, the epoch information, which is not split, is *not* synced. E.g. the recorded duration can be different between processes. It is not quite clear what the "correct" behavior should be here, it would probably depend on what is done based on this information.
2. To make it possible to use the new class, I had to change the net initialization code to not create a new history instance (except when it is `None`) -- instead the history is just cleared. Otherwise, calling fit would always overwrite the `DistributedHistory` with a normal `History` object.
3. Unfortunately, `TCPStore` cannot be pickled. Therefore, I set it to `None` when pickling. This is not tragic as long as users pickle the final model and only load it for inference. If they want to keep on training, they would need to set the `net.history.store` manually.
4. To store values in the kv store as strings, I json-serialize the values, which means that certain types are not supported (say, numpy arrays). I guess the values could be pickled, but that seems like overkill.
5. I had some bugs when testing this with certain Python versions with pytest ("Fatal Python error: Aborted"). This did not happen with other Python versions or when not using pytest. Fingers crossed that this works in CI. I also had to add some `time.sleep` calls in the multiprocessing tests to avoid "broken pipe" etc. **Update**: CI had segfaults for PyTorch 1.11, so I'm skipping tests with `DistributedHistory` for that PyTorch version. Maybe the tests are just flaky, but we don't want that.

### DIFF
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -371,7 +371,7 @@ class DistributedHistory:
     >>> def train(accelerator):
     ...     is_master = accelerator.is_main_process
     ...     world_size = accelerator.num_processes
-    ...     rank = accelerator.process_index
+    ...     rank = accelerator.local_process_index
     ...     store = TCPStore(
     ...         "127.0.0.1", port=1234, world_size=world_size, is_master=is_master)
     ...     dist_history = DistributedHistory(
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -386,10 +386,14 @@ class DistributedHistory:
       :class:`torch.distributed.TCPStore` has been tested to work.
 
     rank : int
-      The rank of this process.
+      The rank of this particular process among all processes. Each process
+      should have a unique rank between 0 and ``world_size`` - 1. If using
+      ``accelerate``, the rank can be determined as
+      ``accelerator.local_process_index``.
 
     world_size : int
-      The number of processes in the training.
+      The number of processes in the training. When using ``accelerate``, the
+      world size can be determined as ``accelerator.num_processes``.
 
     Attributes
     ----------
@@ -520,10 +524,13 @@ class DistributedHistory:
         self._sync_queue.clear()
 
     def __getstate__(self):
-        # unfortunately, TCPStore is not pickleable
         state = self.__dict__.copy()
         try:
+            # Unfortunately, TCPStore and FileStore are not pickleable. We still
+            # try, in case the user provides another type that can be pickled.
             pickle.dumps(state['store'])
-        except TypeError:
+        except (TypeError, pickle.PicklingError):
+            # TCPStore and FileStore raise TypeError, others could be
+            # PicklingError
             state['store'] = None
         return state
--- a/skorch/history.py
+++ b/skorch/history.py
@@ -1,6 +1,7 @@
 """Contains history class and helper functions."""
 
 import json
+import pickle
 
 from skorch.utils import open_file_like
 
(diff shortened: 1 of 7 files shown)


BenjaminBossan:
For the history of this feature, see [this feature](https://github.com/skorch-dev/skorch/issues/952#issuecomment-1505121184). Ping @raphaaal.

skorch/history.py
```diff
@@ -301,3 +302,228 @@ def __getitem__(self, i):
...
+            self._history.record_batch(attr, val)
+
+        self._sync_queue.clear()
+
+    def __getstate__(self):
+        # unfortunately, TCPStore is not pickleable
+        state = self.__dict__.copy()
+        try:
+            pickle.dumps(state['store'])
+        except TypeError:
```
githubnemo, line 527: it would be beneficial to know why this is expected

skorch/history.py
```diff
@@ -301,3 +302,228 @@ def __getitem__(self, i):
...
+    ...     net.fit(X, y)
+
+    Parameters
+    ----------
+    store : torch.distributed.Store
+      The torch distributed ``Store`` instance,
+      :class:`torch.distributed.TCPStore` has been tested to work.
+
+    rank : int
+      The rank of this process.
```
githubnemo, line 389: As a single description of the parameter this is quite tautological. Maybe it is worthwhile to remark the properties and the possible origin of the rank? E.g. "Number differentiating the distributed training processes, e.g. as provided by `accelerate.local_process_index`."?

githubnemo (approved):
Looks clean and well thought out. LGTM

BenjaminBossan:
@githubnemo Very good points, I have extended the docstrings and added more context to the comment in `__getstate__`. I also added a minor change there, namely also catching `PicklingError`. The PyTorch stores will raise `TypeError`, but maybe some users want to provide custom stores, so it's cautious to also catch `PicklingError`.

Oh and btw. coverage claims that [these lines](https://github.com/skorch-dev/skorch/blob/2d78eafa741c8f8787e9bf7f2167bef55175ed6d/skorch/history.py#L517-L519) are not covered, which is not true (adding an error there will raise). I suspect that coverage is not correctly measured because of the use of multiprocessing.

BenjaminBossan:
Failing CI seems to be unrelated.

ottonemo (approved):

PR #956, opened 2023-04-16
Fix failing tests for GPyTorch v1.10

BenjaminBossan:
A couple of changes in GPyTorch v1.10 resulted in our tests failing. This PR fixes those tests. The actual integration has not changed, though, only tests.

Be aware that CI will broken until this PR is merged.

## Note

~The updated tests will now fail for GPyTorch v1.9 and and below. However, I think it is not worth it to write the tests in a backwards compatible manner. First of all, the GPyTorch integration does not appear to be widely used. Second, this only affects the tests, not the actual implementation, so normal users should have no issues. Developers would need to upgrade GPyTorch to avoid the error.~

I had to add code for backwards compatible tests after all because GPyTorch v1.10 is not available for Python 3.7, which we still test.

## Implementation

Initially, we had issues with GPyTorch kernels not being pickleable/copyable. Therefore, the base testing class tested for those errors. Now, 2 of the 3 actual test classes (`ExactGPRegressor` and, since v1.10, `GPRegressor`) work. Therefore, I refactored the base test class to implement the working tests and only the 1 remaining class not working, `GPBinaryClassifier`, to implement the tests that raise.

Content-wise, these failing tests have not been changed. The tests for the working case are now a bit more exact by testing inference results before and after pickling.

Another issue I encountered was that the default likelihood for `GPBinaryClassifier`, `BernoulliLikelihood`, now no longer has any arguments. This is annoying because we want to test a few things with likelihood params, e.g. that params in grid search are passed correctly to the likelihood (since it is an extra module, this could be a real source of errors). Therefore, for testing purposes, I created a likelihood module class that does take an argument (the same as it used to take before v1.10).

I also fixed an incorrect comment from a test, `test_set_params_uninitialized_net_correct_message`. This was probably a copy/paste oversight.

### DIFF
--- a/skorch/tests/test_probabilistic.py
+++ b/skorch/tests/test_probabilistic.py
@@ -124,8 +124,8 @@ class MyBernoulliLikelihood(gpytorch.likelihoods.BernoulliLikelihood):
     a custom class with a (pointless) parameter.
 
     """
-    def __init__(self, *args, max_plate_nesting=1, **kwargs):
-        self.max_plate_nesting = max_plate_nesting
+    def __init__(self, *args, some_parameter=1, **kwargs):
+        self.some_parameter = some_parameter
         super().__init__(*args, **kwargs)
 
 
@@ -340,7 +340,8 @@ class BaseProbabilisticTests:
         params = {
             'lr': [0.01, 0.02],
             'max_epochs': [10, 20],
-            'likelihood__max_plate_nesting': [1, 2],
+            # this parameter does not exist but that's okay
+            'likelihood__some_parameter': [1, 2],
         }
         gp.set_params(verbose=0)
         gs = GridSearchCV(gp, params, refit=True, cv=3, scoring=self.scoring)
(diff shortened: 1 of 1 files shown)


skorch/tests/test_probabilistic.py
```diff
@@ -114,6 +114,21 @@ def forward(self, x):
...
+
+    BernoulliLikelihood used to have parameters before gpytorch v1.10, but now
+    it does not have any parameters anymore. This is not an issue per se, but
+    there are a few things we cannot test anymore, e.g. that parameters are
+    passed to the likelihood correctly when using grid search. Therefore, create
+    a custom class with a (pointless) parameter.
+
+    """
+    def __init__(self, *args, max_plate_nesting=1, **kwargs):
+        self.max_plate_nesting = max_plate_nesting
```
thomasjpfan, line 128: Given that the parameter is not used. Can this name be changed to something that is meaningless in the context of Gaussian Processes?

```suggestion
    def __init__(self, *args, search_parameter=1, **kwargs):
        self.search_parameter = search_parameter
```

thomasjpfan (approved):
Minor comment, otherwise LGTM

skorch/tests/test_probabilistic.py
BenjaminBossan: I renamed the parameter to `some_parameter`. At first, I thought it should not work because the same test is re-used for the 2 other classes of GP that use a different likelihood. But it turns out that passing a non-existing parameter does not cause an error:

https://github.com/cornellius-gp/gpytorch/blob/d23099674f67f04f40467964c37fc02f2e2a8dda/gpytorch/likelihoods/gaussian_likelihood.py#L104-L114

It is a bit strange, thus I put a comment there, but I guess for the purpose of the test we can live with that.

PR #963, opened 2023-05-09
Fix issues with unwrapping modules with accelerate

BenjaminBossan:
There were a few issues at once that are now being fixed:

- When unwrapping, pass `keep_fp32_wrapper=True`, or else the modules are not completely unwrapped and could not be pickled. The test that assumed pickling would fail is now passing.
- Only `net.module_` was unwrapped, but there can be other modules and criteria. Those are now also unwrapped.
- In some circumstances, unwrapping is undesired. For instance, when users want to continue training or benefit from AMP during inference. Therefore, an option was added to prevent automatic unwrapping: set `unwrap_after_train=False`.

Note: This first fix could also work for bf16, but I cannot test it on my machine, so the test still assumes a `PicklingError` to be raised.


thomasjpfan (approved):
This makes sense. LGTM

PR #964, opened 2023-05-10
FIX/DOC: Fix failing readthedocs build

### LINKED ISSUE #962: Docs build failed
The latest doc build failed:

https://readthedocs.org/projects/skorch/builds/20540964/

I haven't investigated too deeply, but this error message:

> Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017...

isn't too ambiguous. 

There is [this issue](https://github.com/readthedocs/readthedocs.org/issues/10290) which seems to discuss the topic, at a glance it's not clear to me if we need to configure something or if they're working on a fix.

BenjaminBossan:
Resolves #962 

Adding this `.readthedocs.yaml` makes the build [pass again](https://readthedocs.org/projects/skorch/builds/20584952/).

I'm not completely happy with this, because some of the configuration now lives in `.readthedocs.yaml`, and some is what we set on the admin panel of the web ui. It is not clear to me what takes precedence when. Ideally, we should migrate to rely solely on `.readthedocs.yaml` but I don't think it's possible. For example, automation rules cannot be specified on the yaml (https://github.com/readthedocs/readthedocs.org/issues/8287). Therefore, if the build error can be fixed without adding a `.readthedocs.yaml`, I'd prefer that.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,6 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.9"
 
 sphinx:
    configuration: docs/conf.py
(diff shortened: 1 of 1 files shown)


thomasjpfan (approved):
LGTM

PR #971, opened 2023-05-26
Add option to globally override callback caching

### LINKED ISSUE #957: Add possibility to disable inference caching globally
To prevent multiple redundant forward calls when calculating metrics with our scoring callbacks, we implemented a [caching mechanism](https://github.com/skorch-dev/skorch/blob/c909e14a1d66de1de23a4619a3c3dc0ef5498a44/skorch/callbacks/scoring.py#L60). Under certain circumstances, e.g. when training in a distributed setting, caching may cause issues and should be disabled. This is already possible by either initializing the scoring callback with `caching=False` or passing that argument as a dunder argument to the net:

```python
# variant 1
net = NeuralNet(..., callbacks=[EpochScoring(..., use_caching=False)])

# variant 2, especially for pre-existing callbacks
net = NeuralNetClassifier(..., valid_acc__use_caching=False)
```

This is okay but can be quite annoying when multiple scoring callbacks are involved. Forgetting to disable caching on a single one of them is enough to potentially cause bugs. Therefore, it would be good to have a 3rd variant to disable caching globally. 

I have not made up my mind on the API for this. It could be an option on the net itself, it could be context manager, or something else completely.

```python
# variant 3a
net = NeuralNet(..., callbacks=[...], override_disable_caching=True)

# variant 3b
net = NeuralNet(...)
with skorch.utils.caching_disabled:
    net.fit(X, y)

# variant 3c
...
```

A minor consideration to make is that if we add this feature, it would be nice if users who have custom callbacks that implement their own caching mechanism can make use of this override option too.

BenjaminBossan:
Resolves #957 (look there for motivation)

I went with the option to add an argument to the `NeuralNet` class itself that can override the caching behavior of the scoring callbacks. This seemed more straightforward than going with a context manager, which is not a pattern we typically require of skorch users. The disadvantage is that we have yet one more parameter on the `NeuralNet`.

By default, the caching behavior of the callbacks is not changed, i.e. this should be fully backwards compatible.

## Naming

I opted to go with the same name of the argument, `use_caching`. I also thought about something like `global_caching_override` or something, but that didn't seem to really fit with the default value of `'auto'`.

## Implementation

Ideally, I would have liked to implement this in a way that any new (or user defined) callback don't have to do anything special to honor the parameter. However, this is not really possible. Although I moved the logic of whether to use inference caching inside of `skorch.callbacks.scoring._cache_net_forward_iter`, so that the global override is automatically taken into account here, there are other places in the callbacks whose behavior changes depending on the use of caching (e.g. target extraction). The check for whether there is a caching override thus has to be performed multiple times.

I think there is no easy way to avoid this.


ottonemo (approved):

BenjaminBossan:
@ottonemo I made to small changes after your review, sorry for that. Could you please check those?

ottonemo (approved):

PR #975, opened 2023-06-01
Add version logging to NeptuneLogger

AleksanderWWW:
- added version logging
- changed import order in one of the methods to avoid unnecessary warnings

### DIFF
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -285,11 +285,11 @@ class NeptuneLogger(Callback):
     @staticmethod
     def _model_summary_file(model):
         try:
-            # neptune-client=0.9.0+ package structure
-            from neptune.new.types import File
-        except ImportError:
             # neptune-client>=1.0.0 package structure
             from neptune.types import File
+        except ImportError:
+            # neptune-client=0.9.0+ package structure
+            from neptune.new.types import File
 
         return File.from_content(str(model), extension='txt')
 
--- a/skorch/callbacks/logging.py
+++ b/skorch/callbacks/logging.py
@@ -176,6 +176,18 @@ class NeptuneLogger(Callback):
         self.keys_ignored = keys_ignored
         self.base_namespace = base_namespace
 
+        self._log_integration_version()
+
+    def _log_integration_version(self) -> None:
+        try:
+            from importlib.metadata import version
+            skorch_version = version("skorch")
+        except ImportError:
+            import pkg_resources
+            skorch_version = pkg_resources.get_distribution('skorch').version
+
+        self.run['source_code/integrations/skorch'] = skorch_version
+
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Hey, thanks for the PR, is this ready for review? Would you mind adding a test for the new change (the version log, not the import order)?

skorch/callbacks/logging.py
```diff
@@ -176,6 +176,18 @@ def __init__(
...
 
+        self._log_integration_version()
+
+    def _log_integration_version(self) -> None:
+        try:
+            from importlib.metadata import version
+            skorch_version = version("skorch")
+        except ImportError:
+            import pkg_resources
+            skorch_version = pkg_resources.get_distribution('skorch').version
```
BenjaminBossan, line 187: Is there any advantage over this method vs just `skorch.__version__`? I know that using these functions is more generic, but since we're only interested in skorch here, and the `__version__` attribute is not going to change, we can safely use it here, right?

BenjaminBossan (commented):
Thx for the update to the neptune logger. I have a comment about a possible simplification, the rest looks good.

skorch/callbacks/logging.py
AleksanderWWW: Got circular imports when trying what you suggest. Look at how this is implemented in `skorch/skorch/__init__.py`. Maybe I missed something

BenjaminBossan: Ah yes, makes sense, `__init__.py` imports `callbacks`, which imports the `NeptuneLogger`. It would, however, work, if the skorch import is done locally inside the method, right?

One more thing I noticed: Could you please move the `_log_integration_version` call from `__init__` to `initialize`?

AleksanderWWW: Done ;)

AleksanderWWW: Shall I also add `__version__` to `__all__` in the `skorch/__init__.py` file to avoid IDE complaints?

BenjaminBossan: > Shall I also add `__version__` to `__all__` in the `skorch/__init__.py` file to avoid IDE complaints?

Hmm, what does the IDE complain about, that `__version__` does not exist? Anyway, I'd rather have it in a separate PR.

AleksanderWWW: `'__version__' is not declared in __all__`

BenjaminBossan: Really not sure about the advantage of adding `__version__` to `__all__`. Maybe if I see a convincing argument :). Anyway, this PR looks good, thx!

BenjaminBossan (approved):

Raalsky:
Post merge but LGTM :wink:

PR #982, opened 2023-06-13
Fix #981: replace deprecated `pkg_resources` calls

### LINKED ISSUE #981: Remove deprecated pkg_resources import
`pkg_resources` is deprecated as an API. I think the alternative is to use `packaging.version`.

The import occurs here: 

https://github.com/skorch-dev/skorch/blob/5b222a5897071c35a3ab0e0663796be366fdc954/skorch/__init__.py#L6-L7

And it makes CIs fail if they raise on deprecation warnings. Eg: https://github.com/fairlearn/fairlearn/actions/runs/5242050040/jobs/9464954323?pr=1241

In `scikit-learn` we vendor `packaging` instead of having it as a dependency, but it's a very lightweight dependency.

ottonemo:
The `pkg_resources` module is deprecated and mostly superseded by `importlib`. To find the torch version we don't actually need to check the metadata, as torch also provides the `__version__` attribute which can be checked directly.

Therefore, `importlib.metadata.version` is only needed to fetch the currently installed skorch version as a drop-in replacement for `pkg_resources.distribution('skorch').version`.


skorch/__init__.py
```diff
@@ -3,15 +3,15 @@
 import sys
 import warnings
 
-import pkg_resources
-from pkg_resources import parse_version
+import importlib.metadata
```
thomasjpfan, line 6: The tests are failing because `importlib.metadata` is only available `Python>=3.8`.

PyTorch has dropped Python 3.7 recently: https://github.com/pytorch/pytorch/commit/5976f0bdfed856ab48824bd41487318f3cf6e615 so I think it's okay for us to drop it as well.

thomasjpfan (commented):
Thank you for the PR!

ottonemo:
@thomasjpfan do you think this is mergeable?

thomasjpfan (approved):

PR #983, opened 2023-06-14
Remove support for python 3.7

ottonemo:
Since it is EOL this month anyway we should drop support and encourage users to upgrade to a more stable python branch.

I removed the TODOs related to python 3.7 to the best of my knowledge.

This is, in part, related to https://github.com/skorch-dev/skorch/pull/982 where `pkg_resources` is replaced with `importlib.metadata` which is only available in python 3.8.


thomasjpfan (approved):
LGTM

PR #987, opened 2023-06-26
This commit fixes the release/deploy script

ottonemo:
Before it was not possible to execute this script without errors as the conda update was broken. I moved the activation of the conda env to `conda run` which is apparently the recommended way.


BenjaminBossan (approved):
This is annoying, not sure why it's necessary. It doesn't affect how the script is run, so we can merge it, but I dislike it when the bash script looks very different from how the commands would be run by hand :-/

PR #990, opened 2023-06-30
MNT: Update CI, add Python 3.11

BenjaminBossan:
- use latest versions of `checkout` and `setup-python` GH actions
- add Python 3.11 to CI

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,8 +16,8 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel all jobs when one fails
       matrix:
-        python_version: ['3.8', '3.9', '3.10']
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Strange, according to [this](https://github.com/pytorch/pytorch/issues/86566#issuecomment-1614227063), PyTorch 2.0.1 should work with Python 3.11 but the CI fails with:

> RuntimeError: Python 3.11+ not yet supported for torch.compile

kit1980:
> Strange, according to [this](https://github.com/pytorch/pytorch/issues/86566#issuecomment-1614227063), PyTorch 2.0.1 should work with Python 3.11 but the CI fails with:

@BenjaminBossan I don't believe the person that made that comment works on PyTorch.

BenjaminBossan:
> I don't believe the person that made that comment works on PyTorch.

Yes, I had still hoped they were telling the truth :D

ottonemo (approved):

PR #1001, opened 2023-08-01
FIX: typo in notebook name

BenjaminBossan:
Was "tranformer", not "transformer".

I noticed because the name in the nbviewer link was correct but led to 404.


thomasjpfan (approved):

BenjaminBossan:
Thx for the quick review :)

PR #1004, opened 2023-08-02
Better error message when NeuralNetClassifier misses the classes_ attribute

BenjaminBossan:
See #1003 

On `NeuralNetClassifier`, when `classes_` is accessed but does not exist, we try to use `classes_inferred_` instead. If that is also not defined, users currently just get an `AttributeError` that `classes_inferred_` does not exist. This can be very confusing.

This PR intercepts the error, provides a (hopefully) helpful error message, then reraises. It is unfortunately not easily possible to deduce what exactly went wrong to get to this point, thus the error message makes an educated guess.

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -98,7 +98,21 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
                 raise AttributeError("{} has no attribute 'classes_'".format(
                     self.__class__.__name__))
             return self.classes
-        return self.classes_inferred_
+
+        try:
+            return self.classes_inferred_
+        except AttributeError as exc:
+            # It's not easily possible to track exactly what circumstances led
+            # to this, so try to make an educated guess and provide a possible
+            # solution.
+            msg = (
+                f"{self.__class__.__name__} could not infer the classes from y; "
+                "this error probably occurred because the net was trained without y "
+                "and some function tried to access the '.classes_' attribute; "
+                "a possible solution is to provide the 'classes' argument when "
+                f"initializing {self.__class__.__name__}"
+            )
+            raise AttributeError(msg) from exc
 
     # pylint: disable=signature-differs
     def check_data(self, X, y):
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Hello @BenjaminBossan, @thomasjpfan, @ottonemo! (cc. @agramfort)

Is this not an API break? This becomes a mandatory self.classes_. In braindecode, a library that used NeuralNetwork as a base, we built an API without considering the n_classes parameter, and everything worked. Now, the whole library is broken if we use the latest version of Skorch.

BenjaminBossan:
@bruAristimunha Thanks for reporting. Could you give more details please? What is the stacktrace? Maybe I'm missing something, but the same code that used to raise an `AttributeError` still raises an `AttributeError` after the change, just with a different error message. I'm not sure how this can break existing code.

BenjaminBossan:
@bruAristimunha Any updates? We plan new skorch release very soon. If there is indeed a breaking change, it would be nice to fix that before the release.

bruAristimunha:
Hi @BenjaminBossan,

So sorry for the delay. I started to make a little code to show the difference, but I realized that there is something more structural that changed in skorch that affected the braindecode. Some update affected everything when we created an environment from scratch. It was a structural thing that even changed learning behaviour. I'm still working on isolating the root; I hope to finish by Friday.

Well, from the beginning, at braindecode, we had a philosophy that we would always use the most up-to-date versions of the libraries we depend on. This was integrated into our CI; we were using skorch directly from Git Hub.

The first error we got was that almost all acceptance tests broke (https://github.com/braindecode/braindecode/pull/461). In acceptance tests, we basically trained on a toy dataset for 10 epochs and compared the results with historical data. I fixed this by updating the tests since it didn't look serious. At first, I thought it was some change related to version 0.14.

Preventively, we limited the skorch on the CI to try to resolve it, and it seemed to have resolved it. However, now, when we install braindecode with sub-dependency skorch, everything breaks. I'll show you in CI here: https://github.com/braindecode/braindecode/pull/471

BenjaminBossan:
Thanks for reporting back. I created a separate issue (#1016), since this seems to be important and can easily missed in the discussion of a closed PR.

PR #1014, opened 2023-08-30
Release 0.15.0

BenjaminBossan:
This is going to be a rather small release. However, since I'll do a presentation about skorch in 2 weeks at PyData Amstersdam, it would be nice to have the bugfixes related to accelerate, as this will be part of the talk.

- Update VERSION
- Remove code for backwards compatibility with v0.11 (see #880): The plan was to remove this 2023-09. We're almost there, I think it's fine to remove it a bit earlier.
- Update CHANGES: Because of the above, I also added an entry on how to transition these pickle files.

---

_Proposed release text:_

This is a smaller release, but it still contains changes which will be interesting to some of you.

We added the possibility to store weights using [safetensors](https://skorch.readthedocs.io/en/latest/user/save_load.html#using-safetensors). This can have several advantages, listed [here](https://github.com/huggingface/safetensors/#yet-another-format-). When calling `net.save_params` and `net.load_params`, just pass `use_safetensors=True` to use `safetensors` instead of pickle.

Moreover, there is a new argument on `NeuralNet`: You can now pass `use_caching=False` or `True` to disable or enable caching for all callbacks at once. This is useful if you have a lot of scoring callbacks and don't want to toggle caching on each individually.

Finally, we fixed a few issues related to using skorch with [accelerate](https://skorch.readthedocs.io/en/stable/user/huggingface.html#accelerate).

Thanks Zach Mueller (@ muellerzr) for his first contribution to skorch.

Find the full list of changes here: https://github.com/skorch-dev/skorch/compare/v0.14.0...v0.15.0


thomasjpfan (approved):
The fixes and release makes sense to me. LGTM

BenjaminBossan:
Thanks Thomas. I updated the date before merging.

PR #1015, opened 2023-08-30
Fix install via test

muellerzr:
Ensures `requirements-dev.txt` is used if installed via `pip install -e .[testing]`


BenjaminBossan:
Great, thanks for fix. Since `requirements-dev.txt` already contains `pytest` and `pytest-cov`, it should be possible to remove their addition completely.

muellerzr:
Nice catch!

BenjaminBossan (approved):
Thanks a lot. Also, just in time for the next skorch release :)

PR #1019, opened 2023-08-31
DOC: Attempt at fixing theme on RTD

### LINKED ISSUE #1018: ReadTheDocs: Wrong theme of docs
On the `latest` version, the skorch docs use the wrong theme: https://skorch.readthedocs.io/en/latest/

image

This not only affects the visual presentation, about also other parts, such as the navigation on the left side, which is is not showing the right level.

The [build logs](https://readthedocs.org/api/v2/build/21770621.txt) don't show anything wrong that could explain this, at least from what I can see. When built locally, the theme is also correct.

BenjaminBossan:
Resolves #1018

Now always enforce the use of `sphinx_rtd_theme`. Previously, this was not done when the docs were built on RTD, instead using the default theme. I'm not really sure why that was so, maybe the default theme already was `sphinx_rtd_theme` but was changed later?

The docs after this change can be found here:

https://skorch.readthedocs.io/en/docs-try-fixing-rtd-theme/

At first I was suspicious if the change in `conf.py` was really the cause of the fix or if this was not instead somehow related to the `latest` version of the docs. Therefore, I created an empty PR #1020 and checked its docs. Those use the false theme, as can be witnessed here:

https://skorch.readthedocs.io/en/docs-experimental-branch-to-check-theme/

Therefore, I think the provided fix should also work for the `latest`, and any other, doc version.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -126,7 +126,7 @@ def setup(app):
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
 
-if not on_rtd:  # only import and set the theme if we're building docs locally
+if True or not on_rtd:  # only import and set the theme if we're building docs locally
     import sphinx_rtd_theme
     html_theme = 'sphinx_rtd_theme'
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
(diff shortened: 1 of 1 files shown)


thomasjpfan (approved):
LGTM

PR #1027, opened 2023-10-24
Add PyTorch v2.1.0 to officially supported version

BenjaminBossan:
At the same time, PyTorch 1.11.0 is no longer officially supported (though probably still works).


thomasjpfan (approved):
LGTM

PR #1034, opened 2023-11-24
Don't fail when pickling models which haven't been trained further

marco-c:
`pbar_` is only set in `on_epoch_begin`, so if `on_epoch_begin` is never called there will be a failure in saving the model.


BenjaminBossan:
Thanks for the bugfix, good catch. Ideally, we would have a test to ensure that there is no regression on this bug. Do you want to give it a shot? Otherwise, we can merge this as is and will add the test later.

marco-c:
@BenjaminBossan I added a test.

BenjaminBossan:
Great, Marco. If you could add a reference to this PR (`#1034`) to the comment of the test, that would be perfect.

BenjaminBossan (approved):
Thanks a lot for fixing the bug and adding the test. LGTM

PR #1045, opened 2024-02-08
Fix issues with sklearn 1.4

BenjaminBossan:
Some skorch tests fail with sklearn v1.4. This commit fixes them:

1. Inheritance structure of scorers seems to have changed.
2. `classes_` attribute should always be numpy array
3. `CalibratedClassifierCV` only works with the output of `predict_proba` being float64

To fix the latter, I'm now casting the output of `predict_proba` to float64. However, I'm not sure if this is a good idea, as it might break existing code. I'm just adding it in for now to check if the tests pass.

**Update**: So the tests pass. Regarding the issue with `CalibratedClassifierCV` and float32, here are some suggestions:

1. Always cast to float64 and :crossed_fingers: that nothing breaks
2. Just keep float32 and skip the tests until the [sklearn fix](https://github.com/scikit-learn/scikit-learn/pull/28247) is released
3. Add an argument to `NeuralNet` to optionally cast `y_proba` to float64.

Personally, I'm in favor of 2.

### DIFF
--- a/skorch/callbacks/scoring.py
+++ b/skorch/callbacks/scoring.py
@@ -100,7 +100,7 @@ def convert_sklearn_metric_function(scoring):
 
         # those are scoring objects returned by make_scorer starting
         # from sklearn 0.22
-        scorer_names = ('_PredictScorer', '_ProbaScorer', '_ThresholdScorer')
+        scorer_names = ('_PredictScorer', '_ProbaScorer', '_ThresholdScorer', '_Scorer')
         if (
                 hasattr(module, 'startswith') and
                 module.startswith('sklearn.metrics.') and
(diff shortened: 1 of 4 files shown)


thomasjpfan (commented):
> Just keep float32 and skip the tests until the sklearn fix is released

I am +1 on waiting for the release.

BenjaminBossan:
@foster999 Good point, thanks. I also just tested scikit-learn 1.5.0 and this time around, there don't seem to be any regressions. Release will come very soon.

foster999:
Excellent, thanks for the quick response! Shall keep my eyes peeled for a release 🚀

BenjaminBossan:
@foster999 Release is out (v1.0.0 no less)

foster999:
Thanks for the heads up @BenjaminBossan ❤️

PR #1053, opened 2024-05-11
Add `CrossEntropyLoss` loss to classifier

amine759:
I had to use `CrossEntropyLoss` for my use case, I thought of creating this PR since `NLLLoss` is the only loss function supported :).

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -14,6 +14,7 @@ from skorch.callbacks import EpochScoring
 from skorch.callbacks import PassthroughScoring
 from skorch.dataset import ValidSplit
 from skorch.utils import data_from_dataset, is_dataset, get_dim, to_numpy
+from typing import Union
 
 neural_net_clf_doc_start = """NeuralNet for classification tasks
 
@@ -58,7 +59,7 @@ class NeuralNetClassifier(NeuralNet, ClassifierMixin):
             self,
             module,
             *args,
-            criterion=torch.nn.NLLLoss,
+            criterion: Union[torch.nn.NLLLoss, torch.nn.CrossEntropyLoss] = torch.nn.NLLLoss,
             train_split=ValidSplit(5, stratified=True),
             classes=None,
             **kwargs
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Thanks for proposing the PR. A few points:

- We don't have type annotations in PEFT. Even if we did, we would support more than those two criteria.
- `CrossEntropyLoss` should already work with `NeuralNetClassifier`.
- The actual addition of your PR seems to be converting one-hot encoded targets to label-encoded targets. I don't think that this is specifically related to `CrossEntropyLoss`, is it?
- Also, AFAIK, sklearn does not generally accept one-hot encoded targets for classification, so I don't think skorch should either.

amine759:
@BenjaminBossan Hi, Thanks for your reply, 
As far as I understood it seems we're restricted to using only the `NLLLoss` loss function with `skorch.NeuralNetClassifier`. When attempting to use any other loss function, like `CrossEntropyLoss`, we encounter an error indicating that the `criterion` parameter only accepts `NLLLoss`.
Whereas the optimizer we can use any other than the default `SGD`. I assumed this is because `get_loss` does not handle the instance of a `torch.nn.CrossEntropyLoss` and since skorch don't use type annotation, I see now it wasn't necessary in the first place. I could have just changed `get_loss` and that's it, right?

BenjaminBossan:
> When attempting to use any other loss function, like `CrossEntropyLoss`, we encounter an error indicating that the `criterion` parameter only accepts `NLLLoss`.

Could you please provide an example to reproduce this error? CE should absolutely work in skorch as is, e.g.:

```python
import numpy as np
from torch import nn

X, y = make_classification(1000, 20, n_informative=10, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

class MyModule(nn.Module):
    def __init__(self, num_units=10, nonlin=nn.ReLU()):
        super().__init__()

        self.dense0 = nn.Linear(20, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(0.5)
        self.dense1 = nn.Linear(num_units, num_units)
        self.output = nn.Linear(num_units, 2)

        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = self.nonlin(self.dense1(X))
        X = self.output(X)
        return X


net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    criterion=nn.CrossEntropyLoss(),
    lr=0.1,
    # Shuffle training data on each epoch
    iterator_train__shuffle=True,
)

net.fit(X, y)
y_proba = net.predict_proba(X)
```

ottonemo:
This got a bit stale :)

@amine759 is this something you're still working on?

PR #1059, opened 2024-06-17
CI: Fix numpy version to below 2.0

BenjaminBossan:
Pin numpy to < v2.0.0, as only PyTorch 2.3 and higher support it.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -34,10 +34,12 @@ jobs:
       with:
         python-version: ${{ matrix.python_version }}
     - name: Install dependencies
+      # TODO remove numpy version constraint once we no longer support PyTorch < 2.3
       run: |
         python -m pip install --upgrade pip
         python -m pip install -r requirements-dev.txt
         python -m pip install -r requirements.txt
+        python -m pip install --force-reinstall -U "numpy<2.0.0"
         python -m pip install pytest-pretty
         python -m pip install torch==${{ matrix.torch_version }} -f https://download.pytorch.org/whl/torch_stable.html
         python -m pip list


BenjaminBossan:
So tests with Python 3.9+ are failing with:

> skorch/tests/conftest.py:163: in <module>
>     import wandb
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/wandb/__init__.py:27: in <module>
>     from wandb import sdk as wandb_sdk
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/wandb/sdk/__init__.py:25: in <module>
>     from .artifacts.artifact import Artifact
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/wandb/sdk/artifacts/artifact.py:45: in <module>
>     from wandb import data_types, env, util
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/wandb/data_types.py:32: in <module>
>     from .sdk.data_types import _dtypes
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/wandb/sdk/data_types/_dtypes.py:393: in <module>
>     NumberType.types.append(np.float_)
> /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/numpy/__init__.py:397: in __getattr__
>     raise AttributeError(
> E   AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.

This comes from wandb, which doesn't work with numpy 2.0: https://github.com/wandb/wandb/issues => 7810. There isn't much we can do from the skorch side, but this should only wandb users, who should pin to numpy v1.26.

BenjaminBossan:
Update: wandb has released a fix. Now encountering a different error on Windows and MacOS:

> RuntimeError: Numpy is not available

BenjaminBossan:
The numpy issues still seem to persist, as can for instance be seen here: https://github.com/skorch-dev/skorch/actions/runs/10682583044/job/29744698675?pr=1064. All runs using either Python > 3.8 and PyTorch < 2.3 fail, as they'll use numpy > 2.0, which is incompatible with the respective PyTorch versions.

The numpy issue was gone intermittently. I think that was because one of the dependencies (possibly accelerate) had pinned it. Now that the pin is gone, the issues resurface. But either way, we should not depend on another dependency to do the pinning for us.

ping @ottonemo @thomasjpfan

PR #1064, opened 2024-08-22
Fix warning from torch.load starting in torch 2.4

BenjaminBossan:
See discussion in #1063

Starting from PyTorch 2.4, there is a warning when `torch.load` is called without setting the weights_only argument. This is because in the future, the default will switch from `False` to `True`, which can result in a lot of errors when trying to load torch files (which are pickle files and thus insecure).

In this PR, we add a possibility for the user to influence the `kwargs` passed to `torch.load` so that they can control that behavior. If not further indicated by the user, we will use the same defaults as the installed torch version. Therefore, users will only encounter this issue via skorch if they would have encountered it via torch anyway.

Since it's not 100% certain if the default will switch in torch 2.6.0, we may have to adjust the version check in the future.

Besides directly testing the `kwargs` being passed on, a test was also added that `net.load_params` does not give any warnings. This is already indirectly tested through some accelerate tests that are currently failing with torch 2.4, but it's better to have an explicit test.

After this is merged, the CI should pass when using torch 2.4.0.

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -46,6 +46,7 @@ from skorch.utils import params_for
 from skorch.utils import to_device
 from skorch.utils import to_numpy
 from skorch.utils import to_tensor
+from skorch.utils import check_torch_weights_only_default_true
 
 
 # pylint: disable=too-many-instance-attributes
@@ -235,6 +236,33 @@ class NeuralNet:
       callbacks.
       Implementation note: It is the job of the callbacks to honor this setting.
 
+    torch_load_kwargs : dict or None (default=None)
+      Additional arguments that will be passed to torch.load when load pickled
+      parameters.
+
+      In particular, this is important to because PyTorch will switch (probably
+      in version 2.6.0) to only allow weights to be loaded for security reasons
+      (i.e weights_only switches from False to True). As a consequence, loading
+      pickled parameters may raise an error after upgrading torch because some
+      types are used that are considered insecure. In skorch, we will also make
+      that switch at the same time. To resolve the error, follow the
+      instructions in the torch error message to designate the offending types
+      as secure. Only do this if you trust the source of the file.
+
+      If you want to keep loading non-weight types the same way as before,
+      please pass:
+
+          torch_load_kwargs={'weights_only': False}
+
+      You should be aware that this is considered insecure and should only be
+      used if you trust the source of the file. However, this does not introduce
+      new insecurities, it rather corresponds to the status quo from before
+      torch made the switch.
+
+      Another way to avoid this issue is to pass use_safetensors=True when
+      calling save_params and load_params. This avoid using pickle in favor of
+      the safetensors format, which is secure by design.
+
     Attributes
     ----------
     prefixes_ : list of str
@@ -311,6 +339,7 @@ class NeuralNet:
             device='cpu',
             compile=False,
             use_caching='auto',
+            torch_load_kwargs=None,
             **kwargs
     ):
         self.module = module
@@ -330,6 +359,7 @@ class NeuralNet:
         self.device = device
         self.compile = compile
         self.use_caching = use_caching
+        self.torch_load_kwargs = torch_load_kwargs
 
         self._check_deprecated_params(**kwargs)
         history = kwargs.pop('history', None)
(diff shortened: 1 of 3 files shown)


skorch/utils.py
```diff
@@ -768,3 +769,16 @@ def _check_f_arguments(caller_name, **kwargs):
             key = 'module_' if key == 'f_params' else key[2:] + '_'
             kwargs_module[key] = val
     return kwargs_module, kwargs_other
+
+
+def check_torch_weights_only_default_true():
```
thomasjpfan, line 774: Given how specific this function is to `torch.load`, can this return `torch_load_kwargs` itself?

BenjaminBossan: Good point, I made the suggested change.

thomasjpfan (approved):

skorch/utils.py
```diff
@@ -768,3 +769,18 @@ def _check_f_arguments(caller_name, **kwargs):
             key = 'module_' if key == 'f_params' else key[2:] + '_'
             kwargs_module[key] = val
     return kwargs_module, kwargs_other
+
+
+def get_torch_load_kwargs():
+    """Returns the kwargs passed to torch.load the correspond to the current
```
ottonemo, line 775:
```suggestion
    """Returns the kwargs passed to torch.load that correspond to the current
```

skorch/utils.py
```diff
@@ -768,3 +769,18 @@ def _check_f_arguments(caller_name, **kwargs):
             key = 'module_' if key == 'f_params' else key[2:] + '_'
             kwargs_module[key] = val
     return kwargs_module, kwargs_other
+
+
+def get_torch_load_kwargs():
```
ottonemo, line 774:
```suggestion
def get_default_torch_load_kwargs():
```

skorch/net.py
```diff
@@ -2620,10 +2650,14 @@ def _get_state_dict(f_name):
 
                 return state_dict
         else:
+            torch_load_kwargs = self.torch_load_kwargs
+            if torch_load_kwargs is None:
+                torch_load_kwargs = get_torch_load_kwargs()
```
ottonemo, line 2655:
```suggestion
                torch_load_kwargs = get_default_torch_load_kwargs()
```

skorch/tests/test_net.py
```diff
@@ -2983,6 +2996,74 @@ def test_save_load_state_dict_custom_module(
         weights_loaded = net_new.custom_.state_dict()['sequential.3.weight']
         assert (weights_before == weights_loaded).all()
 
+    def test_torch_load_kwargs_auto_weights_only_false_when_load_params(
+            self, net_cls, module_cls, monkeypatch, tmp_path
+    ):
+        # Here we assume that the torch version is low enough that weights_only
+        # defaults to False. Check that when no argument is set in skorch, the
+        # right default is used.
+        # See discussion in 1063
```
ottonemo, line 3005: I wonder if we should add a test that tests the default for known old PyTorch versions?
We don't know whether PyTorch 2.6 will have this default but we know that <=2.6.0 should have this behavior.

It is unlikely that we'll add such a test at a later point in time.

BenjaminBossan: Not sure I fully understand the suggestion. Would this amount to monkeypatching `torch.__version__` instead of `get_torch_load_kwargs`? I didn't want to write a test that one day suddenly fails because the torch versions in CI were updated.

ottonemo: No, I meant more like a `skipif` that checks the pytorch version to make sure that functionality is coupled with the pytorch version in the environment.

BenjaminBossan: Okay, I think I got it. The test was added: https://github.com/skorch-dev/skorch/pull/1064/commits/ab9c536516466f2fdb029caf71fd43f2d6e5d375

BenjaminBossan:
CI is failing for unrelated reasons since the latest accelerate release, I opened an issue about it:

https://github.com/huggingface/accelerate/issues/3070

byi8220:
Quick question about the (unrelated) failing CI, are the CI and integration tests run on multigpu environments at all?

BenjaminBossan:
> Quick question about the (unrelated) failing CI, are the CI and integration tests run on multigpu environments at all?

No, we're only using the free runners from GitHub on this repo. Is there anything that we should check specifically on GPU?

byi8220:
> Is there anything that we should check specifically on GPU?

Not sure. I think the only way GPU training would affect pickling is on distributed setups. I'm actually not sure how reliable pickling a running distributed accelerator is (e.g. there are a LOT of stackoverflow or forum posts about running into issues with pickling generators or in a multiprocessing context)

BenjaminBossan:
If such a setting causes trouble, it's probably not just because of `accelerator`, so I think we can disregard that for now.

BenjaminBossan:
@ottonemo have your points been addressed?

PR #1065, opened 2024-09-05
training: fixed EarlyStopping._calc_new_threshold to account for negative scores

MaxBalmus:
This fix is meant to account for training models with potentially negative scores as we see in the case of cost functions based on the log likelihood. The current fix ensures that `abs_threshold_change` is always positive, guaranteeing the intended monotonicity of the threshold.

### DIFF
--- a/skorch/callbacks/training.py
+++ b/skorch/callbacks/training.py
@@ -460,7 +460,7 @@ class EarlyStopping(Callback):
     def _calc_new_threshold(self, score):
         """Determine threshold based on score."""
         if self.threshold_mode == 'rel':
-            abs_threshold_change = self.threshold * score
+            abs_threshold_change = self.threshold * np.abs(score)
         else:
             abs_threshold_change = self.threshold
 


BenjaminBossan:
Thanks for creating this PR. Just to ensure that I got things right: Let's assume we have a negative score, say, -10, and a threshold of 0.1. Then, with the current code, we would have:

`abs_threshold_change = self.threshold * score = -1`

Then, for `lower_is_better = True`, we would have `new_threshold = score - abs_threshold_change = score - -1 = score + 1`. However, we want `score - 1` (vice versa for`lower_is_better = False`) . Is my understanding correct?

Ideally, we could have a unit test for this. I would need to be a bit tricky. Maybe we can create a custom scoring function that just returns from a list of predefined negative scores, and set the `monitor` value to that scoring function. WDYT?

MaxBalmus:
Yes, you described the issue exactly as I see it. 

Regarding the test: I am not as familiar with code, but I think I follow what you mean and I think it should work.

BenjaminBossan:
> Yes, you described the issue exactly as I see it.

Great, thanks for confirming.

> Regarding the test: I am not as familiar with code, but I think I follow what you mean and I think it should work.

Would you be interested in giving it a try? The tests reside here: https://github.com/skorch-dev/skorch/blob/9252477fd0de02d80e08b54203ca19fbe318c94e/skorch/tests/callbacks/test_training.py#L505. If not, it's fine too, just let me know.

BenjaminBossan:
@MaxBalmus gentle ping :)

Mukller (commented):
I traced through the logic and the fix is correct — nice catch. Walkthrough for reviewers, since this is a one-liner with non-obvious consequences:

**`lower_is_better=True` (e.g., negative log-likelihood), `score = -0.8`, `threshold = 0.1`, rel mode:**

- master: `abs_threshold_change = 0.1 * -0.8 = -0.08` → `new_threshold = -0.8 - (-0.08) = -0.72`
  The new threshold is **above** the current score, so a plateau at `-0.8` keeps satisfying `_is_score_improved()` forever → **early stopping never triggers**.
- PR: `abs_threshold_change = 0.1 * |-0.8| = +0.08` → `new_threshold = -0.88 < -0.8` ✓ requires genuine further improvement.

The same failure mode exists in the `lower_is_better=False` branch (negative monitored score → threshold ends up below score → plateau counts as improvement). `np.abs()` fixes both symmetrically.

Two things that would make this ready to merge:

1. **Tests** — the PR currently has none. I'd suggest one parametrized test covering both `lower_is_better` values with negative scores, asserting `dynamic_threshold_` lies on the "requires improvement" side of the current score.
2. **Docstring** — `_calc_new_threshold` could briefly document that in `'rel'` mode the change is proportional to `|score|`.

Behavior is unchanged for positive scores (`abs(x) == x` there), so no backward-compatibility concerns. The log-likelihood monitor motivation in the description makes sense — with tests added I think this is good to merge.

PR #1069, opened 2024-10-18
Add PyTorch 2.5.1 to CI

BenjaminBossan:
* Drop 2.1.2, update 2.4.0 to 2.4.1
* Drop Python 3.8, add Python 3.12
* Filter new torch warning about `weights_only` as that trips up some warning tests
* Unpin protobuf from dev requirements
* Filter warnings caused by usage of `utcnow` by protobuf
* Skip `torch.compile` tests when torch < 2.4 and Python 3.12 are used, as this is not supported by torch

Note that I had to upgrade torch and Python at the same time because torch 2.5 requires Python > 3.8 and because Python 3.12 requires torch > 2.1 (which is dropped in favor of 2.5).

### DIFF
--- a/skorch/tests/test_regressor.py
+++ b/skorch/tests/test_regressor.py
@@ -134,7 +134,14 @@ class TestNeuralNetRegressor:
         X, y = X[:100], y[:100].flatten()  # make y 1d
         net.fit(X, y)
 
-        w0, w1 = recwarn.list  # one warning for train, one for valid 
+        # protobuf uses deprecated utcnow
+        # TODO: remove once protobuf fixed this
+        w_list = [
+            warning for warning in recwarn.list
+            if "protobuf" not in warning.message.args[0]
+            and "utcnow" not in warning.message.args[0]
+        ]
+        w0, w1 = w_list  # one warning for train, one for valid
         # The warning comes from PyTorch, so checking the exact wording is prone to
         # error in future PyTorch versions. We thus check a substring of the
         # whole message and cross our fingers that it's not changed.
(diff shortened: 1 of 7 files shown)


skorch/tests/test_regressor.py
```diff
@@ -134,7 +134,14 @@ def test_dimension_mismatch_warning(self, net_cls, module_cls, data, recwarn):
...
 
-        w0, w1 = recwarn.list  # one warning for train, one for valid 
+        # protobuf uses deprecated utcnow
+        # TODO: remove once protobuf fixed this
+        w_list = [
+            warning for warning in recwarn.list
+            if "protobuf" not in warning.message.args[0]
+            and "utcnow" not in warning.message.args[0]
+        ]
+        w0, w1 = w_list  # one warning for train, one for valid
```
thomasjpfan, line 144: Assuming two warnings by coming out of this filter can be a bit flaky. Can we directly pull out the warnings from training and validation warnings?

BenjaminBossan: Yeah, I thought so too. Do you have an elegant solution for that? I was wondering if there was an API for setting filters to `recwarn` but I didn't see anything.

At least, with this implementation, we catch unexpected warnings ;-)

BenjaminBossan: After spending too much time fiddling with `pytest.mark.filterwarnings` without success, I rewrote the filters in a similar way as you suggested to make them more robust.

thomasjpfan (approved):

PR #1070, opened 2024-10-18
Drop Python 3.8, add Python 3.12

BenjaminBossan:
Python 3.8 is going to be end of life by the end of this month (Oct 2024). PyTorch 2.5 requires Python 3.9 or higher.


BenjaminBossan:
Okay, can't upgrade torch to 2.5 (and drop 2.1) because it requires Python 3.9+ and can't upgrade Python to 3.9+ because torch 2.1 does not support it :) Moving everything into a single PR (#1069) and closing this one.

PR #1075, opened 2024-12-11
ENH: ReduceLROnPlateau records the learning rate and works on batches

BenjaminBossan:
Previously, when using `ReduceLROnPlateau`, we would not record the learning rates in history. The code comment says that's because this class does not expose the `get_last_lr` method. I checked it again and it's now present, so let's use it.

Furthermore, I made a change to enable `ReduceLROnPlateau` to step on each batch (so far, only each epoch was supported). This is consistent with other learning rate schedulers.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 ### Changed
+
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
The CI is failing because of a change in sklearn 1.6 and requires PR #1076.

githubnemo (approved):

PR #1077, opened 2024-12-15
Add step_args parameter to LRScheduler.simulate

githubnemo:
The purpose of this change is to allow for simulation of scheduling policies such as `ReduceLROnPlateau` to be simulated properly.

If `step_args` is an indexable object, it is indexed using the current simulated epoch to get closer to real-life behavior, e.g. simulating a real loss curve and the corresponding behavior of the LR scheduler policy.

### DIFF
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -87,10 +87,12 @@ class LRScheduler(Callback):
         initial_lr: float
           Initial learning rate
 
-        step_args: Any
-          Argument to the `.step()` function of the policy. If it is an
+        step_args: None or float or List[float] (default=None)
+          Argument to the ``.step()`` function of the policy. If it is an
           indexable object the simulation will try to associate every step of
-          the simulation with an entry in ``step_args``.
+          the simulation with an entry in ``step_args``. Scalar values are
+          passed at every step, unchanged. In the default setting (``None``)
+          no additional arguments are passed to ``.step()``.
 
         Returns
         -------
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -75,7 +75,7 @@ class LRScheduler(Callback):
         self.step_every = step_every
         vars(self).update(kwargs)
 
-    def simulate(self, steps, initial_lr):
+    def simulate(self, steps, initial_lr, step_args=None):
         """
         Simulates the learning rate scheduler.
 
@@ -86,6 +86,11 @@ class LRScheduler(Callback):
 
         initial_lr: float
           Initial learning rate
+
+        step_args: Any
+          Argument to the `.step()` function of the policy. If it is an
+          indexable object the simulation will try to associate every step of
+          the simulation with an entry in ``step_args``.
 
         Returns
         -------
@@ -99,10 +104,15 @@ class LRScheduler(Callback):
         sch = policy_cls(opt, **self.kwargs)
 
         lrs = []
-        for _ in range(steps):
+        for step_idx in range(steps):
             opt.step()  # suppress warning about .step call order
             lrs.append(opt.param_groups[0]['lr'])
-            sch.step()
+            if step_args is None:
+                sch.step()
+            elif hasattr(step_args, '__getitem__'):
+                sch.step(step_args[step_idx])
+            else:
+                sch.step(step_args)
 
         return np.array(lrs)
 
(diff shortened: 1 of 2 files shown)


skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -35,6 +35,41 @@ def test_simulate_lrs_batch_step(self, policy):
...
+    def test_simulate_lrs_reduced_lr_on_plateau_scalar(self):
+        # Feed a constant, scalar "loss" to the scheduler.
+        lr_sch = LRScheduler(
+            ReduceLROnPlateau, factor=0.1, patience=1,
+        )
+        lrs = lr_sch.simulate(
+            steps=5, initial_lr=1, step_args=0.5
+        )
+        # O = OK epoch
+        # I = intertolerable epoch
```
BenjaminBossan, line 47: What is "intertolerable"?

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -35,6 +35,41 @@ def test_simulate_lrs_batch_step(self, policy):
...
+        #
+        # O I I I I   epoch classification
+        # 0 1 2 3 4   number of bad epochs
+        #     * * *   epochs with LR reduction
```
BenjaminBossan, line 51: Should it be:
```suggestion
        # O I I I I   epoch classification
        # 0 1 2 3 4   number of bad epochs
        #     *       epochs with LR reduction
```

skorch/tests/callbacks/test_lr_scheduler.py
```diff
@@ -35,6 +35,41 @@ def test_simulate_lrs_batch_step(self, policy):
...
+        #
+        # note that simulate returns the LRs before the step, not after,
+        # so we're seeing only 4 updated values.
+        assert all(lrs == [1] + [1, 1, 0.1, 0.1])
```
BenjaminBossan, line 55: Do we see "4 updated values"? Why use `[1] + ...` here but not in the next test?

skorch/callbacks/lr_scheduler.py
```diff
@@ -87,6 +87,11 @@ def simulate(self, steps, initial_lr):
         initial_lr: float
           Initial learning rate
 
+        step_args: Any
```
BenjaminBossan, line 90: Would it be more helpful to change the type to `None or float or list of float (default=None)`?

Yes, technically it doesn't have to be a list but I don't think it still makes it easier for users to know what they need to pass.

skorch/callbacks/lr_scheduler.py
```diff
@@ -87,6 +87,11 @@ def simulate(self, steps, initial_lr):
...
+          Argument to the `.step()` function of the policy. If it is an
+          indexable object the simulation will try to associate every step of
+          the simulation with an entry in ``step_args``.
```
BenjaminBossan, line 93: Also describe what happens when `None` or `float` are passed?

BenjaminBossan (changes requested):
Thanks for the PR to enhance the capabilities of `simulate`. My comments are mostly about clarifications.

Failing tests require #1078 to be merged.

skorch/tests/callbacks/test_lr_scheduler.py
githubnemo: This is terminology taken from the [docs](https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html#torch.optim.lr_scheduler.ReduceLROnPlateau)

> In the second epoch, if the performance is worse than the baseline, we have what is considered an intolerable epoch.

Drastic terminology but I wanted to adhere to already defined terms :)

BenjaminBossan: Drastic indeed, but it's in the Shakespeare corpus, so it must be a common English word.

skorch/tests/callbacks/test_lr_scheduler.py
githubnemo: No, it's actually

```python
        # O I I I I   epoch classification
        # 0 1 2 1 2   number of bad epochs
        #     *   *   epochs with LR reduction
```

I've updated the comment but the gist is that the bad epoch counter is reset after an LR reduction.

skorch/tests/callbacks/test_lr_scheduler.py
githubnemo: I've removed the `[1]`, it did not help visualize what's going on. I've also changed it to "4 simulated values" since the first value to return is always the init. LR.

skorch/callbacks/lr_scheduler.py
githubnemo: Agreed. Changed!

BenjaminBossan (approved):
Thanks for the changes, it's now much easier to understand what happens.

PR #1078, opened 2024-12-16
Make skorch work with sklearn 1.6.0, attempt 2

BenjaminBossan:
Alternative to #1076

As described in that PR, skorch is currently not compatible with sklearn 1.6.0 or above. As per suggestion, instead of implementing `__sklearn_tags__`, this PR solves the issue by inheriting from `BaseEstimator`.

Related changes:

- It is important to set the correct order when inheriting from `BaseEstimator` and, say, `ClassifierMixin` (`BaseEstimator` should come last).
- As explained in #1076, using `GridSearchCV` with `y` being a torch tensor fails and two tests had to be adjusted.

Unrelated changes

- Removed unnecessary imports from `callbacks/base.py`.

### DIFF
--- a/skorch/tests/test_helper.py
+++ b/skorch/tests/test_helper.py
@@ -451,6 +451,9 @@ class TestSliceDataset:
         gs = GridSearchCV(
             net, params, refit=False, cv=3, scoring='accuracy', error_score='raise'
         )
+        # TODO: after sklearn > 1.6 is released, the to_numpy call should no longer be
+        # required and be removed, see:
+        # https://github.com/skorch-dev/skorch/pull/1078#discussion_r1887197261
         gs.fit(slds, to_numpy(y))  # does not raise
 
     def test_grid_search_with_slds_and_internal_split_works(
@@ -467,6 +470,9 @@ class TestSliceDataset:
         gs = GridSearchCV(
             net, params, refit=True, cv=3, scoring='accuracy', error_score='raise'
         )
+        # TODO: after sklearn > 1.6 is released, the to_numpy call should no longer be
+        # required and be removed, see:
+        # https://github.com/skorch-dev/skorch/pull/1078#discussion_r1887197261
         gs.fit(slds, to_numpy(y))  # does not raise
 
     def test_grid_search_with_slds_X_and_slds_y(
--- a/skorch/tests/test_helper.py
+++ b/skorch/tests/test_helper.py
@@ -437,6 +437,7 @@ class TestSliceDataset:
             self, slds, y, classifier_module):
         from sklearn.model_selection import GridSearchCV
         from skorch import NeuralNetClassifier
+        from skorch.utils import to_numpy
 
         net = NeuralNetClassifier(
             classifier_module,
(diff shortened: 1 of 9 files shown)


BenjaminBossan:
@adrinjalali As per your suggestion, I started inheriting from `BaseEstimator` and the switch was totally painless (inheritance order being the only small trap). This change should probably also play nicer with metadata routing. WDYT?

BenjaminBossan:
> This is nice!

Okay, I'll close the other PR in favor of this one then.

> * scikit-learn now supports array API in many places, which means tensors remain tensors in cross validation etc. It's still experimental and in progress, but you might want to experiment with the config flag.

I'm not sure if the error is directly related to the array API, I think it's more that sklearn is inconsistent in checking the input arrays? Here is an example that shows that mixing numpy arrays and torch tensors works with `LogisticRegression` but not with `GridSearchCV`:

```python
import torch
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import GridSearchCV

X, y = make_classification(1000, 20, n_informative=10, random_state=0)
yt = torch.tensor(y)

lr = LogisticRegression()
lr.fit(X, yt)  # works

gs = GridSearchCV(lr, param_grid={"C": [1, 10, 100]})
gs.fit(X, yt)  # fails
```

The error message is a not very helpful:

> ValueError: Input arrays use different devices: cpu, cpu

(both devices's reprs are "cpu", but one is `"cpu"` and the other  `torch.device("cpu")`)

The reason is the check from lines like this:

https://github.com/scikit-learn/scikit-learn/blob/1922303a79aa776768e2ee89bbda5b6eb4dd5d8b/sklearn/metrics/_classification.py#L224

which causes an error in

https://github.com/scikit-learn/scikit-learn/blob/1922303a79aa776768e2ee89bbda5b6eb4dd5d8b/sklearn/utils/_array_api.py#L141

When removing this check, the score can be calculated correctly, but of course there is no guarantee for that.

> * You might want to have a `set_fit_request` like in this PR to properly enable metadata routing:

Probably, right now it's not officially supported (aka there are no tests for routing). But let's leave that for another PR.

skorch/tests/test_helper.py
```diff
@@ -450,12 +451,13 @@ def test_grid_search_with_slds_works(
         gs = GridSearchCV(
             net, params, refit=False, cv=3, scoring='accuracy', error_score='raise'
         )
-        gs.fit(slds, y)  # does not raise
+        gs.fit(slds, to_numpy(y))  # does not raise
```
thomasjpfan, line 454: Not supporting a mixture of NumPy arrays and CPU pytorch arrays, feels like a regression from scikit-learn's side.

thomasjpfan (approved):
LGTM

skorch/tests/test_helper.py
BenjaminBossan: For me, the question is: is it an intended change or was it introduced accidentally?

If it's unintended, one way to resolve this would be here:

https://github.com/scikit-learn/scikit-learn/blob/1922303a79aa776768e2ee89bbda5b6eb4dd5d8b/sklearn/utils/_array_api.py#L177

There would need to be another check

1. if one of the devices is an instance of `torch.device` 
2. and for this device, `device.type == "cpu"` 
3. and the other device is `"cpu"`

it should be allowed.

thomasjpfan: Looks like there is already a bug report and a PR to fix it in scikit-learn:

https://github.com/scikit-learn/scikit-learn/issues/29107
https://github.com/scikit-learn/scikit-learn/pull/30454

githubnemo (approved):

Issue #1080, opened 2024-12-19
AttributeError: 'NoneType' object has no attribute 'span'


raphaelrubrice:
Hello, 

I am a new user of skorch. When trying to import NeuralNetRegressor I got the error : AttributeError: 'NoneType' object has no attribute 'span'.

This error stems from the documentation retrieval. Either a change in documentation format or a change in the documentation retrieving fonction is causing this error ? Here is the error I got :
```
Cell In[37], line 63
     61 import torch
     62 import torch.nn as nn
---> 63 from skorch import NeuralNetRegressor
     64 from torch.optim import Adam
     66 class MAPNet(nn.Module):

File ~/miniconda3/envs/myenv/lib/python3.13/site-packages/skorch/__init__.py:10
      8 from .history import History
      9 from .net import NeuralNet
---> 10 from .classifier import NeuralNetClassifier
     11 from .classifier import NeuralNetBinaryClassifier
     12 from .regressor import NeuralNetRegressor

File ~/miniconda3/envs/myenv/lib/python3.13/site-packages/skorch/classifier.py:54
     50     return doc
     53 # pylint: disable=missing-docstring
---> 54 class NeuralNetClassifier(NeuralNet, ClassifierMixin):
     55     __doc__ = get_neural_net_clf_doc(NeuralNet.__doc__)
     57     def __init__(
     58             self,
     59             module,
   (...)
     64             **kwargs
     65     ):

File ~/miniconda3/envs/myenv/lib/python3.13/site-packages/skorch/classifier.py:55, in NeuralNetClassifier()
     54 class NeuralNetClassifier(NeuralNet, ClassifierMixin):
---> 55     __doc__ = get_neural_net_clf_doc(NeuralNet.__doc__)
     57     def __init__(
     58             self,
     59             module,
   (...)
     64             **kwargs
     65     ):
     66         super(NeuralNetClassifier, self).__init__(
     67             module,
     68             *args,
   (...)
     71             **kwargs
     72         )

File ~/miniconda3/envs/myenv/lib/python3.13/site-packages/skorch/classifier.py:47, in get_neural_net_clf_doc(doc)
     45 doc = neural_net_clf_doc_start + " " + doc.split("\n ", 4)[-1]
     46 pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+){1,99}')
---> 47 start, end = pattern.search(doc).span()
     48 doc = doc[:start] + neural_net_clf_additional_text + doc[end:]
[...]

PR #1082, opened 2024-12-27
Issue #1080 : Fixing Parsing code for documentation retrieval in Python 3.13.

raphaelrubrice:
Fixes parsing code so that documentation from NeuralNet is properly processed when adapting the documentation for NeuralNetClassifier, NeuralNetBinaryClassifier and NeuralNetRegressor classes in Python 3.13 and below.
See [Issue #1080 ](https://github.com/skorch-dev/skorch/issues/1080) for more details.

### DIFF
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -42,7 +42,7 @@ neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
 
 
 def get_neural_net_clf_doc(doc):
-    doc = neural_net_clf_doc_start + " " + doc.split("\n ", 4)[-1]
+    doc = neural_net_clf_doc_start + " " + doc.split("\n ", 3)[-1]
     pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_clf_additional_text + doc[end:]
@@ -251,7 +251,7 @@ neural_net_binary_clf_criterion_text = """
 
 
 def get_neural_net_binary_clf_doc(doc):
-    doc = neural_net_binary_clf_doc_start + " " + doc.split("\n ", 4)[-1]
+    doc = neural_net_binary_clf_doc_start + " " + doc.split("\n ", 3)[-1]
     pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_binary_clf_criterion_text + doc[end:]
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -43,6 +43,9 @@ neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
 
 def get_neural_net_clf_doc(doc):
     indentation = "    "
+    # dedent/indent roundtrip required for consistent indention in both
+    # Python <3.13 and Python >=3.13
+    # Because <3.13 => not automatic dedent, but it is the case in >=3.13
     doc = neural_net_clf_doc_start + " " + textwrap.indent(textwrap.dedent(doc.split("\n", 5)[-1]), indentation)
     pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
@@ -251,6 +254,9 @@ neural_net_binary_clf_criterion_text = """
 
 def get_neural_net_binary_clf_doc(doc):
     indentation = "    "
+    # dedent/indent roundtrip required for consistent indention in both
+    # Python <3.13 and Python >=3.13
+    # Because <3.13 => not automatic dedent, but it is the case in >=3.13
     doc = neural_net_binary_clf_doc_start + " " + textwrap.indent(textwrap.dedent(doc.split("\n", 5)[-1]), indentation)
     pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -43,7 +43,7 @@ neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
 
 def get_neural_net_clf_doc(doc):
     doc = neural_net_clf_doc_start + " " + doc.split("\n ", 4)[-1]
-    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+){1,99}')
+    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_clf_additional_text + doc[end:]
     doc = doc + neural_net_clf_additional_attribute
@@ -252,7 +252,7 @@ neural_net_binary_clf_criterion_text = """
 
 def get_neural_net_binary_clf_doc(doc):
     doc = neural_net_binary_clf_doc_start + " " + doc.split("\n ", 4)[-1]
-    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+){1,99}')
+    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_binary_clf_criterion_text + doc[end:]
     return doc
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -36,7 +36,7 @@ neural_net_clf_additional_text = """
       skorch behavior should be restored, i.e. raising an
       ``AttributeError``, pass an empty list."""
 
-neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
+neural_net_clf_additional_attribute = """    classes_ : array, shape (n_classes, )
       A list of class labels known to the classifier.
 
 """
@@ -50,7 +50,7 @@ def get_neural_net_clf_doc(doc):
     pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_clf_additional_text + doc[end:]
-    doc = doc + textwrap.indent(neural_net_clf_additional_attribute, indentation)
+    doc = doc + neural_net_clf_additional_attribute
     return doc
 
 # pylint: disable=missing-docstring
--- a/skorch/classifier.py
+++ b/skorch/classifier.py
@@ -43,7 +43,7 @@ neural_net_clf_additional_attribute = """classes_ : array, shape (n_classes, )
 
 def get_neural_net_clf_doc(doc):
     indentation = "    "
-    doc = neural_net_clf_doc_start + " " + textwrap.indent(doc.split("\n", 5)[-1], indentation)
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Thanks for the PR. I checked the docstring based on your branch with Python 3.13 and found that some paragraphs are missing. This is what I get:

```
>>> print(NeuralNetClassifier.__doc__[:1000])
NeuralNet for classification tasks

    Use this specifically if you have a standard classification task,
    with input data X and target y.

  modules will also work.

    criterion : torch criterion (class, default=torch.nn.NLLLoss)
      Negative log likelihood loss. Note that the module should return
      probabilities, the log is applied during ``get_loss``.

    classes : None or list (default=None)
      If None, the ``classes_`` attribute will be inferred from the
      ``y`` data passed to ``fit``. If a non-empty list is passed,
      that list will be returned as ``classes_``. If the initial
      skorch behavior should be restored, i.e. raising an
      ``AttributeError``, pass an empty list.

optimizer : torch optim (class, default=torch.optim.SGD)
  The uninitialized optimizer (update rule) used to optimize the
  module

lr : float (default=0.01)
  Learning rate passed to the optimizer. You may use ``lr`` instead
  of using ``optimizer__lr``, which would result in the sa
```

Is it the same for you?

What I would expect to see is:

```
>>> print(NeuralNetClassifier.__doc__[:2000])
NeuralNet for classification tasks

    Use this specifically if you have a standard classification task,
    with input data X and target y.

    In addition to the parameters listed below, there are parameters
    with specific prefixes that are handled separately. To illustrate
    this, here is an example:

    >>> net = NeuralNet(
    ...    ...,
    ...    optimizer=torch.optimizer.SGD,
    ...    optimizer__momentum=0.95,
    ...)

    This way, when ``optimizer`` is initialized, :class:`.NeuralNet`
    will take care of setting the ``momentum`` parameter to 0.95.

    (Note that the double underscore notation in
    ``optimizer__momentum`` means that the parameter ``momentum``
    should be set on the object ``optimizer``. This is the same
    semantic as used by sklearn.)

    Furthermore, this allows to change those parameters later:

    ``net.set_params(optimizer__momentum=0.99)``

    This can be useful when you want to change certain parameters
    using a callback, when using the net in an sklearn grid search,
    etc.

    By default an :class:`.EpochTimer`, :class:`.BatchScoring` (for
    both training and validation datasets), and :class:`.PrintLog`
    callbacks are added for convenience.

    Parameters
    ----------
    module : torch module (class or instance)
      A PyTorch :class:`~torch.nn.Module`. In general, the
      uninstantiated class should be passed, although instantiated
      modules will also work.

    criterion : torch criterion (class, default=torch.nn.NLLLoss)
      Negative log likelihood loss. Note that the module should return
      probabilities, the log is applied during ``get_loss``.

    classes : None or list (default=None)
      If None, the ``classes_`` attribute will be inferred from the
      ``y`` data passed to ``fit``. If a non-empty list is passed,
      that list will be returned as ``classes_``. If the initial
      skorch behavior should be restored, i.e. raising an
      ``AttributeError``, pass an empty 
```

Do you see the same results?

Apart from that, we could try to fix the inconsistent indentation for Python 3.13 using `textwrap.dedent`, but let's focus on the missing paragraphs first.

raphaelrubrice:
Thanks for catching that, this was stemming from the fact that in Python 3.13 there is one "\n " which becomes a "\n" with no trailing whitespace which lead to one return statement being skipped and leading to missing criterion block as discussed in Issue #1080  . 

Fixing the argument back to "\n"  solved that. However, there was the ":class:`.NeuralNetRegressor`." that was remaining at the end of NeuralNet documentation after splitting with `.split("\n", 4)`. This was fixed by changing it to `.split("\n", 5)`. 

I then proceeded to use `textwrap` as you suggested to fix inconsistent indentation.

Once again I tested imports from both Python 3.13 and Python 3.12 and both works with correct documentation retrieval.

raphaelrubrice:
Okay so I corrected functions body. 

About the roundtrip, I orginally forgot why I had done that after reading your comment, went on testing in Pyhthon 3.13 found that it was not necessary in the version and committed. That was a mistake thus the followig sections explains why we actually need this roundtrip and thus why I have done 2 commits (the last one reverted the removal of `textwrap.dedent`. This is why : 

When using only a simple `doc.split` without using textwrap for indentation or dedentation, documentation is not indented as you would expect. However when using `textwrap.indent(doc.split`, in Python 3.13 documentation is correctly indented but no longer the case in Python 3.12 : 

```
NeuralNet for classification tasks

    Use this specifically if you have a standard classification task,
    with input data X and target y.


        In addition to the parameters listed below, there are parameters
        with specific prefixes that are handled separately. To illustrate
        this, here is an example:

        >>> net = NeuralNet(
        ...    ...,
        ...    optimizer=torch.optimizer.SGD,
        ...    optimizer__momentum=0.95,
        ...)

        This way, when ``optimizer`` is initialized, :class:`.NeuralNet`
        will take care of setting the ``momentum`` parameter to 0.95.

        (Note that the double underscore notation in
        ``optimizer__momentum`` means that the parameter ``momentum``
        should be set on the object ``optimizer``. This is the same
        semantic as used by sklearn.)

        Furthermore, this allows to change those parameters later:

        ``net.set_params(optimizer__momentum=0.99)``

        This can be useful when you want to change certain parameters
        using a callback, when using the net in an sklearn grid search,
        etc.

        By default an :class:`.EpochTimer`, :class:`.BatchScoring` (for
        both training and validation datasets), and :class:`.PrintLog`
        callbacks are added for convenience.

        Parameters
        ----------
        module : torch module (class or instance)
          A PyTorch :class:`~torch.nn.Module`. In general, the
          uninstantiated class should be passed, although instantiated
          modules will also work.

    criterion : torch criterion (class, default=torch.nn.NLLLoss)
      Negative log likelihood loss. Note that the module should return
      probabilities, the log is applied during ``get_loss``.

    classes : None or list (default=None)
      If None, the ``classes_`` attribute will be inferred from the
      ``y`` data passed to ``fit``. If a non-empty list is passed,
      that list will be returned as ``classes_``. If the initial
      skorch behavior should be restored, i.e. raising an
      ``AttributeError``, pass an empty list.

        optimizer : torch optim (class, default=torch.optim.SGD)
          The uninitialized optimizer (update rule) used to optimize the
          module

        lr : float (default=0.01)
          Learning rate passed to the optimizer. You may use ``lr`` instead
          of using ``optimizer__lr``, which would result in the same outcome.
```

So **to ensure correct indentation in both versions**, performing a dedentation operation before applying indentation ensures proper indentation : 
When using the roundtrip : 
```
NeuralNet for classification tasks

    Use this specifically if you have a standard classification task,
    with input data X and target y.


    In addition to the parameters listed below, there are parameters
    with specific prefixes that are handled separately. To illustrate
    this, here is an example:

    >>> net = NeuralNet(
    ...    ...,
    ...    optimizer=torch.optimizer.SGD,
    ...    optimizer__momentum=0.95,
    ...)

    This way, when ``optimizer`` is initialized, :class:`.NeuralNet`
    will take care of setting the ``momentum`` parameter to 0.95.

    (Note that the double underscore notation in
    ``optimizer__momentum`` means that the parameter ``momentum``
    should be set on the object ``optimizer``. This is the same
    semantic as used by sklearn.)

    Furthermore, this allows to change those parameters later:

    ``net.set_params(optimizer__momentum=0.99)``

    This can be useful when you want to change certain parameters
    using a callback, when using the net in an sklearn grid search,
    etc.

    By default an :class:`.EpochTimer`, :class:`.BatchScoring` (for
    both training and validation datasets), and :class:`.PrintLog`
    callbacks are added for convenience.

    Parameters
    ----------
    module : torch module (class or instance)
      A PyTorch :class:`~torch.nn.Module`. In general, the
      uninstantiated class should be passed, although instantiated
      modules will also work.

    criterion : torch criterion (class, default=torch.nn.NLLLoss)
      Negative log likelihood loss. Note that the module should return
      probabilities, the log is applied during ``get_loss``.

    classes : None or list (default=None)
      If None, the ``classes_`` attribute will be inferred from the
      ``y`` data passed to ``fit``. If a non-empty list is passed,
      that list will be returned as ``classes_``. If the initial
      skorch behavior should be restored, i.e. raising an
      ``AttributeError``, pass an empty list.

    optimizer : torch optim (class, default=torch.optim.SGD)
      The uninitialized optimizer (update rule) used to optimize the
      module

    lr : float (default=0.01)
      Learning rate passed to the optimizer. You may use ``lr`` instead
      of using ``optimizer__lr``, which would result in the same outcome.
```

This is the rationale behind the use of the roundtrip.

skorch/classifier.py
```diff
@@ -40,16 +41,15 @@
...
 def get_neural_net_clf_doc(doc):
-    doc = neural_net_clf_doc_start + " " + doc.split("\n ", 4)[-1]
-    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+){1,99}')
+    indentation = "    "
+    doc = neural_net_clf_doc_start + " " + textwrap.indent(textwrap.dedent(doc.split("\n", 5)[-1]), indentation)
+    pattern = re.compile(r'(\n\s+)(criterion .*\n)(\s.+|.){1,99}')
     start, end = pattern.search(doc).span()
     doc = doc[:start] + neural_net_clf_additional_text + doc[end:]
-    doc = doc + neural_net_clf_additional_attribute
+    doc = doc + textwrap.indent(neural_net_clf_additional_attribute, indentation)
```
BenjaminBossan, line 50: Here the doc is indented again but for the other classes, this step is missing. I think it can be removed here, right?

BenjaminBossan (changes requested):
Thanks for the further explanation. To summarize: Since the change in Python 3.13 is basically that the docstring is automatically dedented, it makes sense to do dedent + indent to normalize the indentation for <3.13 and >= 3.13. Let's add this as a comment for future readers.

skorch/classifier.py
raphaelrubrice: Actually it cannot, here's why. 

As it is currently defined, `neural_net_clf_additional_attribute` is no indented whereas for other additional texts for NeuralNetRegressor and NeuralNetBinaryClassifier objects, the indentation is done when defining the text to add. We can however define  `neural_net_clf_additional_attribute` properly to prevent the use of `textwrap.indent`, if so, tell me and I'll modify it. Otherwise it should remain as is.

Note that this is the case no matter the version because here Python 3.13's automatic dedentation will not make any difference with Python 3.12 as there is nothing to dedent at the beginning of the text. Here are some outputs : 

3.13 with indentation of  `neural_net_clf_additional_attribute` : 
```
    _optimizers : list of str
      List of names of all optimizers. This list is collected dynamically when
      the net is initialized. Typically, there is no reason for a user to modify
      this list.

    classes_ : array, shape (n_classes, )
          A list of class labels known to the classifier.
```

3.13 without indentation of  `neural_net_clf_additional_attribute` : 
```
    _optimizers : list of str
      List of names of all optimizers. This list is collected dynamically when
      the net is initialized. Typically, there is no reason for a user to modify
      this list.

classes_ : array, shape (n_classes, )
      A list of class labels known to the classifier.
```

Same for Python 3.12.

BenjaminBossan: I see, thanks for explaining. The real issue is than how we define the `neural_net_clf_additional_attribute` constant:

https://github.com/skorch-dev/skorch/blob/547d3ffaa3c402276fa8f0ccd4766ffb8d6869b7/skorch/classifier.py#L38-L41

In contrast to the other doc snippets, this one lacks the new lines and indentation at the beginning. So how about we add those and then remove the `textwrap.indent` call here? That way, all three classes handle the `__doc__` in a consistent manner.

raphaelrubrice: Yes indeed, it has now been corrected.

BenjaminBossan (approved):
Thanks a lot for fixing the docstring issue with Python 3.13 and ensuring consistency between the versions. I tested it locally and the docstrings are identical.

Issue #1081, opened 2024-12-20
PyPI Readme Rendering


Ball-Man:
Hello,

I wanted to notify that the [PyPI page of skorch](https://pypi.org/project/skorch/) is not rendering correctly. In fact, it looks like it the long description is being treated as plain text, while it is actually an rst document.

## The problem
I did investigate this on a private fork, and by running a twine check this is what I get:
```bash
twine check dist/*

Checking dist/skorch-1.0.1.dev0-py3-none-any.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.
         line 10: Warning: Cannot scale image!
           Could not get size from "https://github.com/skorch-dev/skorch/workflows/tests/badge.svg":
           Requires Python Imaging Library.
           Reading external files disabled.
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
Checking dist/skorch-1.0.1.dev0.tar.gz: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.
         line 10: Warning: Cannot scale image!
           Could not get size from "https://github.com/skorch-dev/skorch/workflows/tests/badge.svg":
           Requires Python Imaging Library.
           Reading external files disabled.
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.
```

I did notice myself that `long_description_content_type` is missing. It may or may not be a good idea to make it explicit, even though it is inferred automatically by the renderer.

The real problem seems to be related to image scaling, in particular of the badges (coverage, etc.). It looks like this is a common issue, notified earlier this year by other devs (https://github.com/pypa/twine/issues/1102). The issue has reached the actual readme_rendered devs (https://github.com/pypa/readme_renderer/issues/304) but it looks like it still remains unsolved.

## A possible solution
What some of the devs in the cited threads suggest, is to remove scaling from the badges. I tried this solution on my fork, and it does solve the rendering issue. I still get some artifacts, but it may be related to the fact that I was testing on TestPyPI.

A practical example:
```rst
.. |coverage| image:: https://github.com/skorch-dev/skorch/blob/master/assets/coverage.svg
    :alt: Test Coverage
.. to be removed
    :scale: 100%
```

My fork on GitHub shows that there is no difference in the visualization of the badges without the explicit scaling factor (at least, not on my desktop): https://github.com/Ball-Man/skorch/tree/no_scaling_readme.

[...]

PR #1086, opened 2025-01-08
Issue #1081: Fix PyPI Readme Rendering

Ball-Man:
The readme renderer refuses to render scaled images (in this particular case, badges), as a result the [skorch PyPi page](https://pypi.org/project/skorch/0.2.0/) is displayed as plain text. This is easily fixed by removing scaling from badges in the README. On top of that, Explicitly specifying RST content type for the long description solves a warning during rendering.
See #1081 for more info.


githubnemo:
Great! Thanks for the changes :)

Issue #1091, opened 2025-01-27
Why the torch_load_kwargs parameter doesn't work?


githubnemo:
### Discussed in https://github.com/skorch-dev/skorch/discussions/1090

<div type='discussions-op-text'>

<sup>Originally posted by **lnzpgo** January 26, 2025</sup>
I'm getting these annoying warning messages from torch.load:

> /home/user/Code/ML-PROJ/.venv/lib/python3.11/site-packages/skorch/net.py:2261: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.

Despite I'm explicitely passing weights_only=True to skorch...

```
# Wrap the PyTorch model in a scikit-learn compatible interface using skorch
net = NeuralNetRegressor(
    module=SimpleNN,
    criterion=nn.MSELoss,
    optimizer=optim.Adam,
    max_epochs=20,
    batch_size=32,
    train_split=None,  # Disable internal validation split (we use cross-validation)
    verbose=0,
    # Set weights_only=True for loading model weights
    torch_load_kwargs={'weights_only': True},
)
# Define hyperparameter grids for Grid Search and Random Search
param_grid = {
    'module__hidden_size': [10, 20, 30],  # Number of units in the hidden layer
    'optimizer__lr': [0.001, 0.01, 0.1],  # Learning rate
    'batch_size': [32, 64, 128, 256],  # Batch size
}
# Grid Search with Cross-Validation
print("Performing Grid Search...")
grid_search = GridSearchCV(
    estimator=net,
    param_grid=param_grid,
    scoring='neg_mean_squared_error',  # Use negative MSE for scoring
    cv=5,  # 5-fold cross-validation
    n_jobs=-1,  # Use all available CPU cores
    verbose=1,
)
grid_search.fit(X, y)
```

### Reason

[...]

PR #1092, opened 2025-01-27
FIX Unpickling without using torch.load

BenjaminBossan:
Resolves #1090, #1091.

This one got more complicated than I initially thought. So here it goes:

## Description

PyTorch plans to make the switch to `weights_only=True` for torch.load. We already partly dealt with that in #1064 when it comes to `save_params/load_params`. However, we still had a gap. Namely, when using pickle directly, i.e. when going through `__getstate__` and `__setstate__`, we are still using `torch.load` and `torch.save` without handling `weights_only`. This will cause trouble in the future when the default is switched. But it's also annoying right now, because users will get the `FutureWarning` about `weights_only`, even if they correctly pass `torch_load_kwargs` (see #1090).

The reason why we use `torch.save/torch.load` for pickle is that those functions are basically _extended_ pickle functions that have the benefit of supporting the `map_location` argument to handle the device of torch tensors, which we don't have for pickle. The `map_location` argument is important, e.g. when saving a net that uses CUDA and loading it on a machine without CUDA, we would otherwise run into an error.

However, with the move to `weights_only=True`, these `torch.save/torch.load` will become _reduced_ pickle functions, as they will only support a small subset of objects by default. Therefore, we wouldn't be able to rely on `torch.save/torch.load` for pickling the whole skorch object.

## Solution

(thanks ChatGPT o1 for helping with this)

In this PR, we move to using plain pickle for this. However, now we run into the issue of how to handle the `map_location`. The solution I ended up with is now to intercept torch's `_load_from_bytes` using a custom `Unpickler`, and to specifically use `torch.load` there. That way, we can pass the `map_location` and other `torch_load_kwargs`. The remaining unpickling process just works as normal.

Yes, this is a private function, so we cannot be sure if it'll work indefinitely, If there is a better suggestion, I'm open to it. However, the function has existed for 7 years, so it's not very likely that it will change anytime soon:

https://github.com/pytorch/pytorch/blame/0674ab7e33c3f627ca6781ce98468ec1dd4743a5/torch/storage.py#L525

A drawback of the solution is that we cannot just load old skorch nets that were saved with `torch.save` using `pickle.load`. This is because torch uses custom `persistent_load` functions. When trying to load with pickle, we thus get:

> _pickle.UnpicklingError: A load persistent id instruction was encountered, but no persistent_load function was specified.

Therefore, I had to keep `torch.load` as a fallback to avoid backwards incompatibility. The bad news is that the initial problem persists, namely that even when passing `torch_load_kwargs`, users get the `FutureWarning` about `weights_only`. The good news is that users can just re-save their net with the new skorch version and from then on they won't see the warning again.

Note that I didn't add a specific test for this problem of loading nets from before the change, because `test_pickle_load`, which uses a checked in pickled net, already covers this.

## Other considered solutions

1. Why not continue using `torch.save/torch.load` and just pass the `torch_load_kwargs` argument to it? This is unfortunately not that easy. When switching to `weights_only=True`, torch will refuse to load any custom objects, e.g. `class MyModule`. There is a way to prevent that, namely via `torch.serialization.add_safe_globals`, but it is a ton of work to add all required objects there, as even builtin Python types are mostly not supported.
2. We cannot use `with torch.device(map_location):`, as this is not honored during unpickling.
3. During `__getstate__`, we could recursively go through the state, pop all torch tensors, and replace them with, say, numpy arrays and additional meta data like the device, then use this info to restore those objects during `__setstate__`. Even though this looks like a cleaner solution, it is much more complex and therefore, I'd argue, more error prone.
4. _Edit_: Don't do anything and just live with the warning: This will work -- until PyTorch switches the default. Therefore, we had to tackle this sooner or later.

## Notes

While working on this, I thought that we could most likely remove the `cuda_dependent_attributes_` (which contains the `net.module_`, `net.optimizer_`, etc.). Their purpose was to call `torch.load` on these attributes specifically, but with the new `Unpickler`, it should also work without this. However, I kept the attribute for now, mainly for these reasons:

1. I didn't want to change more than necessary, as these changes are delicate and I don't to break any existing skorch code or pickle files.
2. The attribute itself is public, so in theory, users may rely on its existence (not sure if in practice). We would thus have to keep most of the code related to this attribute anyway.

But LMK if you think we should deprecate and eventually remove this attribute.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 ### Changed
+
+- Loading of skorch nets using pickle: When unpickling a skorch net, you may come across a PyTorch warning that goes: "FutureWarning: You are using torch.load with weights_only=False [...]"; to avoid this warning, pickle the net again and use the new pickle file (#1092)
+
 ### Fixed
 
 ## [1.1.0]
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -13,6 +13,7 @@ from itertools import chain
 from collections import OrderedDict
 from contextlib import contextmanager
 import os
+import pickle
 import tempfile
 import warnings
 
@@ -33,6 +34,7 @@ from skorch.exceptions import SkorchAttributeError
 from skorch.exceptions import SkorchTrainingImpossibleError
 from skorch.history import History
 from skorch.setter import optimizer_setter
+from skorch.utils import MapLocationUnpickler
 from skorch.utils import _identity
 from skorch.utils import _infer_predict_nonlinearity
 from skorch.utils import FirstStepAccumulator
@@ -2242,7 +2244,7 @@ class NeuralNet(BaseEstimator):
             state.pop(k)
 
         with tempfile.SpooledTemporaryFile() as f:
-            torch.save(cuda_attrs, f)
+            pickle.dump(cuda_attrs, f)
             f.seek(0)
             state['__cuda_dependent_attributes__'] = f.read()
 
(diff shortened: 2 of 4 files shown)


githubnemo (approved):
LGTM, seems to be a good complexity/impact trade-off.

githubnemo (approved):

PR #1093, opened 2025-01-31
Update: PyTorch v2.6.0

BenjaminBossan:
- Add torch 2.6.0 to CI
- Remove torch 2.2.2
- Update torch install instructions, as they [no longer provide conda packages](https://github.com/pytorch/pytorch/issues/138506)


RoyiAvital:
I hope PyTorch will reconsider this decision.
Anyhow, you may use the PyTorch package provided by `conda-forge`.

BenjaminBossan:
> Anyhow, you may use the PyTorch package provided by `conda-forge`.

For the install instructions, I think it's better to recommend the official way. Of course, users stay free to choose conda-forge if they want to.

BenjaminBossan:
Unfortunately I had to add an instruction to uninstall triton, which appears to be installed via some dependency and actually breaks `torch.compile` with PyTorch < 2.6. Triton [won't/can't fix this](https://github.com/triton-lang/triton/issues/5722).

.github/workflows/testing.yml
```diff
@@ -39,6 +39,7 @@ jobs:
         python -m pip install --force-reinstall -U "numpy<2.0.0"
         python -m pip install pytest-pretty
         python -m pip install torch==${{ matrix.torch_version }} -f https://download.pytorch.org/whl/torch
+        python -m pip uninstall --yes triton # https://github.com/triton-lang/triton/issues/5722
```
thomasjpfan, line 42: Should we use a triton 3.1 instead?

Or tweak it so that we use triton 3.2 for PyTorch 2.6 and then triton 3.1 for everything else?

thomasjpfan: I think `torch.compile` works on CPU but it does not use triton as the backend, so it's really weird to have CPU compile error based on triton versions. https://pytorch.org/tutorials/intermediate/inductor_debug_cpu.html

I mostly want the CI to be as similar to a typical user, which has triton installed. (I suspect users will downgrade if they come across the triton issue)

BenjaminBossan: AFAICT, the error is unrelated to whether triton is actually being used or not, as it happens at import time: https://github.com/triton-lang/triton/issues/5722.

PR #1099, opened 2025-03-11
Nit Fix in LR Scheduler Notebook

ParagEkbote:
I've updated the notebook to clarify about the training data being used in the notebook.

Could you please review the changes?

cc: @BenjaminBossan


BenjaminBossan (approved):
LGTM, thanks.

PR #1104, opened 2025-04-24
Update: PyTorch v2.7.0

BenjaminBossan:
- Dropped 2.3.1
- For PyTorch 2.6, install triton 3.2
- Drop numpy<2.0 install (no longer required for PyTorch)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false  # don't cancel all jobs when one fails
       matrix:
         python_version: ['3.9', '3.10', '3.11', '3.12']
-        torch_version: ['2.3.1+cpu', '2.4.1+cpu', '2.5.1+cpu', '2.6.0+cpu']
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -31,18 +31,19 @@ jobs:
       with:
         python-version: ${{ matrix.python_version }}
     - name: Install dependencies
-      # TODO remove numpy version constraint once we no longer support PyTorch < 2.3
       # TODO remove triton 3.1 install if torch < 2.6 no longer supported, see #1093
+      # TODO remove triton 3.2 install if torch < 2.7 no longer supported
```
thomasjpfan, line 35: I kind of wish that PyTorch would put an upper bound on Triton for releases builds.

thomasjpfan (approved):
LGTM

.github/workflows/testing.yml
BenjaminBossan: I agree, I'm not sure if this has been reported to PyTorch, I saw some issues on the triton repo, but it's not in their scope.

Issue #82, opened 2017-10-12
*Examples* Add an example how to handle data streams with Dataloader


githubnemo:
The example should include

- how data streams such as generators can be used in conjunction with the Dataloader
- how to achieve train/test split with such a setup

ottonemo:
This example should not be bound to any release, we can publish them independently.

benjamin-work:
Does the [image segmentation example](https://github.com/dnouri/skorch/tree/master/examples/nuclei_image_segmentation) solve this issue?

ParagEkbote:
@BenjaminBossan Is this issue open for contribution? What kind of example would be ideal?

githubnemo:
The basic idea was to have examples for generators used to wrap the data loader - this was before there was the `IterableDataset` from PyTorch. This was a usage pattern that I often found helpful when doing small mutations and experiments in jupyter notebook, e.g.:

```python
def my_dataloader(ds, *args, **kwargs):
    dataloader = DataLoader(ds, *args, **kwargs):
    for batch, y in dataloader:
        # do something fancy with batch/y
        yield batch, y

net = NeuralNet(
    ...,
    iterator_train=my_dataloader,
```

This is nice for experimenting because there is not a lot of modification needed and nothing is hidden. But that was then, now we have constructs like `IterableDataset` from PyTorch which I haven't played around with extensively but I'd imagine that there are quite some complications with train/valid splitting. For that it would be nice to have an example IMO.

ParagEkbote:
Can this issue be closed as completed as part of #1105?

cc: @BenjaminBossan

PR #1105, opened 2025-05-29
Add Example for Skorch DataLoader

ParagEkbote:
Refs #82 

In this example, we use the `Iterable Dataset` class from torch for a synthetic streaming dataset. We use a custom callback for validation since `train_split` cannot be used for streaming datasets. I believe I have named the notebook a bit incorrectly and would appreciate feed-back on it.

Could you please review the changes?

cc: @githubnemo, @BenjaminBossan


notebooks/Streaming_Dataset.ipynb
```diff
@@ -0,0 +1,194 @@
...
+    "from skorch.callbacks import Callback"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data Preparation\n",
+    "\n",
+    "In this example, we will use synthetic dataset for training the model. We will utilize the `Iterable Dataset` from PyTorch to create a binary classification dataset."
```
BenjaminBossan, line 48:
```suggestion
    "In this example, we will use synthetic dataset for training the model. We will utilize the `IterableDataset` from PyTorch to create a binary classification dataset."
```

notebooks/Streaming_Dataset.ipynb
```diff
@@ -0,0 +1,194 @@
...
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Model Training and Result\n",
+    "\n",
+    "We will now wrap the neural net classifier with the torch model and use `DataLoader` to create data batches. We will also add the custom validation callback seperately from the training loop. It is important to note that since we are using `IterableDataset`,we will set the train_split to None.\n",
```
BenjaminBossan, line 155: Let's add a short explanation.

```suggestion
    "We will now wrap the neural net classifier with the torch model and use `DataLoader` to create data batches. We will also add the custom validation callback separately from the training loop. It is important to note that since we are using `IterableDataset`, we have to set the `train_split` to `None`, since streaming data cannot be split in advance.\n",
```

notebooks/Streaming_Dataset.ipynb
```diff
@@ -0,0 +1,194 @@
...
+   "source": [
+    "class StreamingDataset(IterableDataset):\n",
+    "    def __init__(self, length=1000, seed=42):\n",
+    "        self.length = length\n",
+    "        self.rng = torch.Generator().manual_seed(seed)\n",
+    "\n",
+    "    def __iter__(self):\n",
+    "        for _ in range(self.length):\n",
+    "            X = torch.randn(20, generator=self.rng)\n",
+    "            y = torch.randint(0, 2, (1,), generator=self.rng).item()\n",
```
BenjaminBossan, line 65: Just a proposal: When `y` is not completely random, the net can actually learn something and improve the loss.

```suggestion
    "            y = (X.sum() > 10).sum()\n",
```

notebooks/Streaming_Dataset.ipynb
```diff
@@ -0,0 +1,194 @@
...
+    "        return self.fc2(X)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Custom Callback\n",
+    "\n",
+    "We will now setup a validation callback for use during model training. The callback is defined to evaluate the model’s accuracy on a validation set at the end of each training epoch. The training dataset and  validation dataset are wrapped in Torch `DataLoader` to enable batch processing."
```
BenjaminBossan, line 101: Let's mention that normally, it is not advised to create callbacks with references to the data. But here it is fine because 1) we cannot use `train_split` and 2) the data is created/loaded lazily so that callback does not reference actual data in memory.

BenjaminBossan (changes requested):
Thanks for this PR to add an example notebook for streaming data. It is clear and precise, nicely done. I have a few small suggestions for improvement, please check.

notebooks/Streaming_Dataset.ipynb
ParagEkbote: This is a good suggestion. Updated as discussed.

notebooks/Streaming_Dataset.ipynb
ParagEkbote: Hmm, I think that adding a rule-based approach for a single variable is a bit unfavorable since we are adding a dependency to a specific variable(X).

Do you think that adding controlled noise to synthetic data with deterministic logic could be useful?

A proposed example could be:

```py
class StreamingDataset(IterableDataset):
    def __init__(self, length=1000, seed=42, noise_prob=0.1, threshold=3.0):
        self.length = length
        self.rng = torch.Generator().manual_seed(seed)
        self.noise_prob = noise_prob
        self.threshold = threshold
```

BenjaminBossan: Sorry, I don't understand the reply, could you please elaborate? My suggestion was simply to make `y` correlate with `X`. That way, when we train the model, we can see the loss improving. When `y` is completely random, the loss is not improving. For the purpose of this notebook, one could say it doesn't matter, but I can imagine some viewers being confused by the stagnating loss, perhaps assuming there is an error, hence my suggestion.

notebooks/Streaming_Dataset.ipynb
```diff
@@ -0,0 +1,196 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Custom Callback\n",
+    "\n",
+    "We will now setup a validation callback for use during model training. The callback is defined to evaluate the model’s accuracy on a validation set at the end of each training epoch. The training dataset and  validation dataset are wrapped in Torch `DataLoader` to enable batch processing.\n",
+    "\n",
+    "It is important to note that callbacks with references to the data are not recommended due to security risks and performance issues. We cannot use the traditional `train_split` for streaming datasets and since the data is created in a synthetic manner and not referenced in memory, a custom callback is implemented."
```
BenjaminBossan, line 103: Thanks for adding the explanation. I reworded it a bit for clarity, WDYT?

```suggestion
    "Normally, callbacks with references to the data are not recommended to keep the model and training data separate. However, as we cannot use the traditional `train_split` for streaming datasets and since the data is streamed, this is okay here."
```

BenjaminBossan (commented):
Thanks for the updates. I just left 2 small comments, please check, the rest looks good.

notebooks/Streaming_Dataset.ipynb
ParagEkbote: This is more concise and conveys the same message. I have updated it.

notebooks/Streaming_Dataset.ipynb
ParagEkbote: What I meant was to add controlled noise/entropy to a synthetic dataset instead of having a correlation that could be seen as less realistic. But I agree that the stagnating loss could be seen as an error. I'll update the variables.

ParagEkbote:
I have been able to execute the notebook completely and the final output is being shown correctly as well:

image

Could you please review?

cc: @BenjaminBossan

BenjaminBossan (approved):
Thanks for contributing this notebook all looks good.

PR #1108, opened 2025-06-15
Add pyproject.toml to Improve CI/CD and Tooling

### LINKED ISSUE #1106: Improve CI/CD by migrating from setup.py to pyproject.toml
Since the release of [PEP 621](https://peps.python.org/pep-0621/), the .toml file has been increasingly adopted for simpler usage and its enhanced functionality.  I believe that by adopting the .toml file for packaging and managing dependencies, we can also make the process of publishing future release versions simpler as well. WDYT?

cc: @BenjaminBossan

ParagEkbote:
Fixes #1106 

As described in the issue, I have chosen to subsume 6 python tooling legacy into a single pyproject.toml file with minimal changes. 

## Motivation

As described in [PEP 639](https://peps.python.org/pep-0639/), [PEP 518](https://peps.python.org/pep-0518/) and [PEP 621](https://peps.python.org/pep-0621/), the pyproject.toml file is a modern python tooling file which can manage multiple tasks of testing, linting, building of package, codecoverage ,etc.  It has key-value configuration format.

## Major Changes

1. Building the source and wheels for the library
This functionality does not have any breaking changes, I've tested it out with the `python -m build` command.

2. Linting with pylint and flake8
Linting with flake8 has no breaking changes. pylint has some changes. version 2.6 has deprecated `no-space-check` option with `no_self_use` being made optional. pylint options `locally-enabled`,` super-on-old-class` were deprecated in version 2.14. I've updated it with the relevant rules to make the lint efficient.

I've tested out both tools, they are working correctly. 

3. Using classifiers for PyPI
When we publish packages on PyPI, classifers help categorize Python packages. These classifiers provide metadata about the package, such as its development status, intended audience, license, supported Python versions, and more. I've current added a standard set of packages for the library. Changes can be made to them as required.

5. Testing and Code Coverage
We can run the tests by running the `pytest` command with no breaking changes.  Testing Code coverage has no breaking changes as well. You can check it with the `pytest -v --cov=./skorch --cov-report=term-missing` command.

6. Deprecating Manifest.in, requirements.txt and requirements-dev.txt
The functions of Manifest.in can be done by the .toml file as described in the setuptools docs [here](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#manifest-in). The optional dependencies (prev. requirements-dev.txt) are listed in [project.optional-dependencies] and dependencies (prev requirements.txt) are listed in the project section.

Lastly, I have updated the gh-actions file to reflect these changes as well. I do know that the changes are a bit sweeping, but keeping the CI/CD and Python tooling up to date is quite essential and lead to less headaches as seen in #1081 . Could you please review the changes?
cc: @BenjaminBossan

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,11 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Add an example of hyper-parameter optimization using [Optuna](https://optuna.org/) [here](https://github.com/skorch-dev/skorch/tree/master/examples/optuna) (#1098)
 ### Added
 
 - Add Contributing Guidelines for skorch. (#1097)
+- Add an example of hyper-parameter optimization using [Optuna](https://optuna.org/) [here](https://github.com/skorch-dev/skorch/tree/master/examples/optuna) (#1098)
 - Add Example for Streaming Dataset(#1105)
+- Add pyproject.toml to Improve CI/CD and Tooling (#1108)
+
 ### Changed
 
 - Loading of skorch nets using pickle: When unpickling a skorch net, you may come across a PyTorch warning that goes: "FutureWarning: You are using torch.load with weights_only=False [...]"; to avoid this warning, pickle the net again and use the new pickle file (#1092)


pyproject.toml
```diff
@@ -0,0 +1,186 @@
+[build-system]
+requires = ["setuptools>=61.1.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.package-data] 
+"*" = ["VERSION", "LICENSE", "*.rst"]
+
+[tool.setuptools.packages.find]
+include = ["skorch*"]
+exclude = ["notebooks", "assets","skorch.tests*"]
```
BenjaminBossan, line 10: Would this result in `skorch/tests` being excluded from the PyPI package? Right now, it is included. It could be argued that the tests are not necessary but they're also not big and who knows if someone is (inadvertently) depending on them being present.

pyproject.toml
```diff
@@ -0,0 +1,186 @@
...
+license-files = ["LICENSE"]
+requires-python = ">=3.9"
+dependencies =[
+    "numpy>=1.13.3",
+    "scikit-learn>=0.22.0",
+    "scipy>=1.1.0",
+    "tabulate>=0.7.7",
+    "tqdm>=4.14.0",
+]
+authors = [
```
BenjaminBossan, line 25: Let's add "Marian Tietz".

.github/workflows/testing.yml
```diff
@@ -35,9 +35,8 @@ jobs:
       # TODO remove triton 3.2 install if torch < 2.7 no longer supported
       run: |
         python -m pip install --upgrade pip
-        python -m pip install -r requirements-dev.txt
-        python -m pip install -r requirements.txt
-        python -m pip install pytest-pretty
+        python -m pip install .[test,docs,dev,others]
+        python -m pip install pytest-pretty pytest-cov
```
BenjaminBossan, line 39: pytest-cov is already included in `test`.

.github/workflows/testing.yml
```diff
@@ -51,4 +50,4 @@ jobs:
         python -m pip install .
     - name: Test with pytest
       run: |
-        pytest
+        pytest
```
BenjaminBossan, line 53:
```suggestion
        pytest
```

pyproject.toml
```diff
@@ -0,0 +1,186 @@
...
+    "flaky",
+]
+
+docs = [
+    "sphinx",
+    "sphinx_rtd_theme",
+    "numpydoc",
+]
+
+others = [
```
BenjaminBossan, line 67: I'm wondering if we can find a better name than `others`. Maybe `full`, which is dev + test + docs + others? Not sure if there are some conventions around it.

BenjaminBossan (changes requested):
Thank you for this fantastic PR, well done cleaning up all these different configs and adopting the new `pyproject.toml`.

I have a few smaller comments, but overall this looks very close to being finished.

There is a big difference for pylint, going from 809 lines of logs to 1536 on my machine, probably because the `no_self_use` rule is not disabled. But honestly, I think it's fine since I don't really use it anymore and I suspect others aren't either. We should move to a more modern linter like black or ruff eventually -- with string normalization disabled, skorch should already be quite close to what they expect.

pyproject.toml
ParagEkbote: There are no particular conventions that I know of.  Updated to `full` for now.

pyproject.toml
ParagEkbote: I actually thought that the tests were excluded. I think that if we are going to add the tests, we could include `pytest` in the core dependencies. Based on the following reference articles, the tests are recommended to not be kept in the packaged source code.

Reference articles:

1. https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#choosing-a-test-layout-import-rules
2. https://www.pyopensci.org/python-package-guide/package-structure-code/python-package-structure.html#don-t-include-test-suite-datasets-in-your-package

BenjaminBossan (approved):
Thanks a lot for the last few changes, the PR LGTM.

Before merging this big change, I'd like a second pair of eyes to check the PR, pinging @thomasjpfan @githubnemo

githubnemo (commented):
Thanks @ParagEkbote!

The changes LGTM except for minor comments and some questions regarding the dependency grouping. 

Regardless of possible changes in the dependency grouping, let's not forget to update the README with the new instructions, e.g., instead of `requirements-dev.txt` we now need to install via `pip install '.[full]'`. This goes for `docs/user/installation.rst` as well.

docs/user/installation.rst
```diff
@@ -38,9 +38,9 @@ If you just want to use skorch, use:
     conda create -n skorch-env python=3.10
     conda activate skorch-env
     # install pytorch version for your system (see below)
-    python -m pip install -r requirements.txt
     python -m pip install .
 
+
```
ParagEkbote, line 43: This is one of the benefits of using pyproject.toml :)

githubnemo:
Thanks a lot, @ParagEkbote, for the very thorough implementation.
This is a long overdue improvement!

PR #1112, opened 2025-07-16
pre-commit setup

ParagEkbote:
This is a draft PR that showcases the diff between files after `pre-commit run --all-files` is executed. flake8 still has some failures:

3     E402 module level import not at top of file
1     E713 test for membership should be 'not in'
5     E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
1     E722 do not use bare 'except'
4     E731 do not assign a lambda expression, use a def
5     E741 ambiguous variable name 'l'
1     F541 f-string is missing placeholders
1     F824 `global _linkcode_git_revision` is unused: name is never assigned in scope

Could you please let me know your thoughts?

cc: @BenjaminBossan


BenjaminBossan (changes requested):
Thanks for the PR. After skimming through the files, overall, the changes look acceptable and I think we will be able to proceed.

Before doing that, could you please:

1. Remove docs/examples from the checked folders? Maybe we can add them later but for now I think they're not necessary.
2. Let's add a step to the `tests.yml` to run the checkers+linters to ensure that the code stays in a good shape.
3. "flake8 still has some failures": I think we should be able to fix those, but it's probably better done in a separate PR. Would you be willing to create that?

PR #1119, opened 2025-08-13
Custom sklearn estimators implementing `__sklearn_is_fitted__`

### LINKED ISSUE #1118: Custom sklearn estimators implementing `__sklearn_is_fitted__ `
A not fitted `NeuralNetClassifier` and `NeuralNetRegressor` have 3 attributes (`['history_', 'initialized_', 'virtual_params_']`) that "trick" `sklearn`'s `check_is_fitted` function into thinking the model is fitted and a `NotFittedError` is not raised.

## Example

```python
net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
    iterator_train__shuffle=True,
)

from sklearn.utils.validation import check_is_fitted

check_is_fitted(net) # this does not raise a NotFittedError
```

It does not raise a `NotFittedError` because the not fitted net has these 3 attributes that end in an underscore which "trick" the check function into thinking the net is fitted. 

## sklearn protocol

[sklearn protocol](https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html#sphx-glr-auto-examples-developing-estimators-sklearn-is-fitted-py) suggest that downstream estimators implement a `__sklearn_is_fitted__` which gets called when `check_is_fitted` runs. 

## Suggestion

Add something like this to the `NeuralNetClassifier` and `NeuralNetRegressor` class as these are the attributes that appear only after the model is fitted. 

```python
def __sklearn_is_fitted__(self):
        """
        Check fitted status and return a Boolean value.
        """
        return all(hasattr(self, attr) for attr in [
                "init_context_",
                "callbacks_",
                "prefixes_",
                "cuda_dependent_attributes_",
                "module_",
                "criterion_",
                "optimizer_",
                "classes_inferred_", # this is only on NeuralNetClassifier
            ])
```

<details>
<summary>Full example code block</summary>

```python
import numpy as np
from torch import nn
from sklearn.model_selection import train_test_split

X, y = make_classification(n_samples=1000, n_features=10, n_informative=4, n_classes=2)
X = X.astype(np.float32)

X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)

class MyModule(nn.Module):
    def __init__(self, num_units=10, nonlin=nn.ReLU()):
        super().__init__()
        self.dense0 = nn.Linear(10, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(0.5)
        self.dense1 = nn.Linear(num_units, num_units)
        self.output = nn.Linear(num_units, 2)

        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = self.nonlin(self.dense1(X))
        X = self.softmax(self.output(X))
        return X

net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
    iterator_train__shuffle=True,
)

from sklearn.utils.validation import check_is_fitted

check_is_fitted(net) # this does not raise a NotFittedError
```

</details>

divakaivan:
Closes #1118 

Things to note:

- I saw the docstring [here](https://github.com/skorch-dev/skorch/blob/44c35271e9b82e136d998c16cc442d3e2204910f/skorch/hf.py#L66) so I made this `__sklearn_is_fitted__` docstring similar
- renamed the existing `test_not_fitted_raises` into `test_not_init_raises` but looking for advice on your preferences. 
- also not sure if a `test_not_fitted_raises` test is needed in all 3 of `tests/test_net.py`, `tests/test_regressor.py` and `tests/test_classifier.py` as the one in `tests/test_net.py` cover it. Again, seeking maintainers' preferences here. 

cc @BenjaminBossan

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1377,6 +1377,14 @@ class NeuralNet(BaseEstimator):
         )
         check_is_fitted(self, attributes, *args, **kwargs)
 
+    def __sklearn_is_fitted__(self):
+        """This method is called when sklearn's ``check_is_fitted`` is used.
+        
+        Explained here: 
+        https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html
+        """
+        return self.check_is_fitted()
+
     def trim_for_prediction(self):
         """Remove all attributes not required for prediction.
 
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1383,7 +1383,15 @@ class NeuralNet(BaseEstimator):
         Explained here: 
         https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html
         """
-        return self.check_is_fitted()
+        from skorch.exceptions import NotInitializedError
+        try:
+            self.check_is_fitted()
+        except NotInitializedError:
+            from sklearn.exceptions import NotFittedError
+            raise NotFittedError(
+                f"This {self.__class__.__name__} instance is not fitted yet. "
+                "Call 'fit' with appropriate arguments before using this estimator."
+            )
 
     def trim_for_prediction(self):
         """Remove all attributes not required for prediction.
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -19,6 +19,7 @@ import warnings
 
 import numpy as np
 from sklearn.base import BaseEstimator
+from sklearn.exceptions import NotFittedError
 import torch
 from torch.utils.data import DataLoader
 
@@ -30,6 +31,7 @@ from skorch.dataset import ValidSplit
 from skorch.dataset import get_len
 from skorch.dataset import unpack_data
 from skorch.exceptions import DeviceWarning
+from skorch.exceptions import NotInitializedError
 from skorch.exceptions import SkorchAttributeError
 from skorch.exceptions import SkorchTrainingImpossibleError
 from skorch.history import History
@@ -1383,11 +1385,9 @@ class NeuralNet(BaseEstimator):
         Explained here: 
         https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html
         """
-        from skorch.exceptions import NotInitializedError
         try:
             self.check_is_fitted()
         except NotInitializedError:
-            from sklearn.exceptions import NotFittedError
             raise NotFittedError(
                 f"This {self.__class__.__name__} instance is not fitted yet. "
                 "Call 'fit' with appropriate arguments before using this estimator."
--- a/skorch/tests/test_classifier.py
+++ b/skorch/tests/test_classifier.py
@@ -253,7 +253,7 @@ class TestNeuralNetBinaryClassifier:
         clone(net_fit)
 
     @pytest.mark.parametrize('method', INFERENCE_METHODS)
-    def test_not_fitted_raises(self, net_cls, module_cls, data, method):
+    def test_not_init_raises(self, net_cls, module_cls, data, method):
         from skorch.exceptions import NotInitializedError
         net = net_cls(module_cls)
         X = data[0]
@@ -263,6 +263,19 @@ class TestNeuralNetBinaryClassifier:
 
         msg = ("This NeuralNetBinaryClassifier instance is not initialized "
                "yet. Call 'initialize' or 'fit' with appropriate arguments "
+               "before using this method.")
+        assert exc.value.args[0] == msg
+
+    def test_not_fitted_raises(self, net_cls, module_cls):
+        from sklearn.utils.validation import check_is_fitted
+        from sklearn.exceptions import NotFittedError
+    
+        net = net_cls(module_cls)
+        with pytest.raises(NotFittedError) as exc:
+            check_is_fitted(net)
+
+        msg = ("This NeuralNetBinaryClassifier instance is not initialized yet. "
+               "Call 'initialize' or 'fit' with appropriate arguments "
                "before using this method.")
         assert exc.value.args[0] == msg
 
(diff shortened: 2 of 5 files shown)


skorch/net.py
```diff
@@ -1377,6 +1377,14 @@ def check_is_fitted(self, attributes=None, *args, **kwargs):
         )
         check_is_fitted(self, attributes, *args, **kwargs)
 
+    def __sklearn_is_fitted__(self):
+        """This method is called when sklearn's ``check_is_fitted`` is used.
+        
+        Explained here: 
+        https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html
+        """
+        return self.check_is_fitted()
```
BenjaminBossan, line 1386: Note that `self.check_is_fitted()` does not return a bool, instead it raises a `NotFittedError`. So for the signature of this method to be correct, we'd have to call `self.check_is_fitted()`, intercept the error, and return based on whether the error was raised or not.

The tests still pass, as they check for the very error that is raised here, but I don't think that's the intent. The error, and the message, should correspond to what sklearn would return, since we explicitly call `check_is_fitted` from sklearn, i.e. the message should be something like:

> `"This NeuralNetBinaryClassifier instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator."`

BenjaminBossan (changes requested):
Thanks for creating this PR so quickly. I think it's not quite correct yet, though, so please check my comment.

divakaivan:
> Thanks for creating this PR so quickly. I think it's not quite correct yet, though, so please check my comment.

I am a mentee in European Summer of Code, contributing to [skore](https://github.com/probabl-ai/skore) and I discovered that [not fitted skorch models pass through our code as if they are fitted](https://github.com/probabl-ai/skore/issues/1966). So OSS is kind of my "job" at the moment haha

skorch/net.py
```diff
@@ -1377,6 +1377,22 @@ def check_is_fitted(self, attributes=None, *args, **kwargs):
...
+        """This method is called when sklearn's ``check_is_fitted`` is used.
+        
+        Explained here: 
+        https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html
+        """
+        from skorch.exceptions import NotInitializedError
+        try:
+            self.check_is_fitted()
+        except NotInitializedError:
+            from sklearn.exceptions import NotFittedError
```
BenjaminBossan, line 1390: Thanks for updating the method. You can make these imports global, it's only in the tests that we use local imports.

BenjaminBossan (changes requested):
Thanks for the update. Only a small comment, the rest looks good.

divakaivan:
Hmm ~ not sure this test fail is related to the changes from this PR, is it? 🤔 
```
________________________________________ TestNeuralNet.test_fit_lbfgs_optimizer ________________________________________

self = <skorch.tests.test_net.TestNeuralNet object at 0x7f1d6182dea0>
net_cls = <class 'skorch.classifier.NeuralNetClassifier'>
module_cls = functools.partial(<class 'skorch.toy.MLPModule'>, output_nonlin=Softmax(dim=-1), input_units=20, hidden_units=10, num_hidden=2, dropout=0.5)
data = (array([[-0.96[58](https://github.com/skorch-dev/skorch/actions/runs/16988488056/job/48164133761?pr=1119#step:6:59)346 , -2.1890705 ,  0.16985609, ..., -0.89645284,
         0.3759244 , -1.0849651 ],
       [-0.454767..., 0, 1,
       1, 0, 0, 1, 1, 0, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 1, 0, 0, 1, 1, 0,
       0, 1, 1, 0, 1, 0, 1, 0, 1, 0]))

    @flaky(max_runs=5)
    def test_fit_lbfgs_optimizer(self, net_cls, module_cls, data):
        # need to randomize the seed, otherwise flaky always runs with
        # the exact same seed
        torch.manual_seed(int(time.time()))
        X, y = data
        net = net_cls(
            module_cls,
            optimizer=torch.optim.LBFGS,
            lr=1.0,
            batch_size=-1,
        )
        net.fit(X, y)
    
        last_epoch = net.history[-1]
>       assert last_epoch['train_loss'] < 1.0
E       assert np.float64(7.054504871368408) < 1.0

skorch/tests/test_net.py:2525: AssertionError
```

skorch/tests/test_classifier.py
```diff
@@ -266,6 +266,19 @@ def test_not_fitted_raises(self, net_cls, module_cls, data, method):
...
+        from sklearn.utils.validation import check_is_fitted
+        from sklearn.exceptions import NotFittedError
+    
+        net = net_cls(module_cls)
+        with pytest.raises(NotFittedError) as exc:
+            check_is_fitted(net)
+
+        msg = ("This NeuralNetBinaryClassifier instance is not fitted yet. "
+                "Call 'fit' with appropriate arguments before "
+                "using this estimator.")
```
BenjaminBossan, line 279: The indentation is a bit off in the 2nd and 3rd line. Overall, I think it's easier and more consistent to use hanging indents here and below:

```suggestion
        msg = (
            "This NeuralNetBinaryClassifier instance is not fitted yet. "
            "Call 'fit' with appropriate arguments before "
            "using this estimator."
        )
```

BenjaminBossan (changes requested):
Thanks for fixing the imports. I only now noticed an issue with the indentation, sorry for not noticing earlier. Please also add a brief entry to the [CHANGES.md](https://github.com/skorch-dev/skorch/blob/master/CHANGES.md).

BenjaminBossan (approved):
Thanks for implementing `__sklearn_is_fitted__`, the PR LGTM.

divakaivan:
Thanks for the help 🙏

PR #1139, opened 2026-04-20
ENH NeuralNet metadata routing

### LINKED ISSUE #1095: Implement sklearn API Metadata routing
The main problem here is the ability to use weird input schemes (mixed lengths, Datasets etc) with complex pipelines, which often await tabular data. This is for example the case with torch_geometric Data objects, and complex, non-GNN architectures receiving multiple types of "X".
see [https://github.com/skorch-dev/skorch/issues/1084](https://github.com/skorch-dev/skorch/issues/1084) for a relevant issue.

proposed solution: metadata routing. If we can successfully route the data (e.g: an array or pandas DataFrame of torch_geometric.data.Data), it would be trivial to pass them as additional fit params, and use a collate_fn to handle stuff from there.
Metadata routing is not implemented for now in skorch NeuralNet, but it looks like it's been partailly done for LLMs?

adrinjalali:
## Implement sklearn metadata routing (router + consumer)

Closes #1095

### Summary

Implements sklearn's [metadata routing](https://scikit-learn.org/stable/metadata_routing.html) protocol for `NeuralNet`. NeuralNet is both a **consumer** (its module's forward method accepts arbitrary metadata) and a **router** (it routes metadata like `groups` to its internal CV splitter).

This follows the same pattern as sklearn's `CalibratedClassifierCV`: `get_metadata_routing()` returns a `MetadataRouter` with `add_self_request(self)` for self-consumed params plus `add(splitter=..., method_mapping=fit→split)` for the CV splitter child.

All behavior is gated on `sklearn.set_config(enable_metadata_routing=True)`. When disabled, behavior is identical to before.

### What this enables

When `enable_metadata_routing=True`, metadata flows correctly through sklearn meta-estimators (Pipeline, GridSearchCV, cross_validate) to NeuralNet and its internal components.

**Routing `groups` through a Pipeline to GroupKFold:**

```python
import numpy as np
import sklearn
from sklearn.base import BaseEstimator, TransformerMixin
from sklearn.model_selection import GroupKFold
from sklearn.pipeline import make_pipeline
from sklearn.preprocessing import StandardScaler
from torch import nn

from skorch.dataset import ValidSplit
from skorch.regressor import NeuralNetRegressor

sklearn.set_config(enable_metadata_routing=True)


class RegressionModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc = nn.Linear(10, 1)

    def forward(self, X, Z, **kwargs):
        return self.fc(X + Z)


class DictScaler(TransformerMixin, BaseEstimator):
    """Scale only the 'X' key of a dict input, pass 'Z' through."""

    def __init__(self):
        self.scaler_ = StandardScaler()

    def fit(self, X, y=None):
        self.scaler_.fit(X["X"])
        return self

    def transform(self, X):
        result = dict(X)
        result["X"] = self.scaler_.transform(X["X"]).astype("float32")
        return result


X, y = make_regression(n_samples=1000, n_features=20, random_state=0)
X, Z = X[:, :10].astype("float32"), X[:, 10:].astype("float32")
y = y.astype("float32").reshape(-1, 1)
groups = np.array([0] * 500 + [1] * 500)

# Per-sample auxiliary data (Z) is packed into a dict so the
# DataLoader batches it alongside X. Metadata routing handles
# getting `groups` through the Pipeline to the GroupKFold splitter.
X_dict = {"X": X, "Z": Z}

net = NeuralNetRegressor(
    RegressionModule,
    max_epochs=3,
    lr=0.01,
    train_split=ValidSplit(GroupKFold(2)),
)

pipe = make_pipeline(DictScaler(), net)
pipe.fit(X_dict, y, groups=groups)
```

### Changes

**`skorch/net.py`:**

- **`get_metadata_routing()`** — returns a `MetadataRouter` (not `MetadataRequest`). Registers self as consumer via `add_self_request` and the inner CV splitter (from `ValidSplit.cv`) as a child. This is what makes NeuralNet a router.
- **`set_fit_request()` / `set_partial_fit_request()`** — custom implementations since `fit(**fit_params)` uses `**kwargs` and sklearn can't auto-generate these. Follows `_BaseScorer.set_score_request` pattern. Uses `__metadata_request__partial_fit` class attribute to suppress conflicting auto-generation (TODO references scikit-learn/scikit-learn#32111).
- **`_get_metadata_request()`** — overrides parent to use class name string as `owner` instead of the instance. Needed because `NeuralNet.__getstate__` uses `pickle.dump` for CUDA-dependent attributes, which causes `deepcopy` to block when the module isn't picklable. An alternative would be implementing a full `__deepcopy__`.
- **`__deepcopy__()`** — bypasses `__getstate__`/`__setstate__` (which use pickle) by deepcopying each attribute individually with a fallback to shallow copy. Needed because sklearn's routing infrastructure calls `deepcopy` on MetadataRouter objects that hold references to the NeuralNet instance (e.g. via Pipeline's `get_metadata_routing`).
- **`_get_splitter_for_routing()`** — extracts the CV splitter from `ValidSplit.cv` when it supports routing (e.g. `GroupKFold`). Returns `None` for int/float cv or custom callables.
- **`fit_loop()`** — when routing is enabled, calls `process_routing()` to validate and route metadata. Extracts `split_params` for `get_split_datasets` from routed params. Following sklearn's router pattern, `forward_params` are the full `fit_params` (the router uses its own params directly).
- **`_get_param_names()`** — excludes `_metadata_request` (set by `set_fit_request`, not `__init__`; sklearn's `clone` handles it separately via `deepcopy`).

### What this does NOT cover (future PRs)

- Predict-path metadata (`set_predict_request` / `set_predict_proba_request`)
- Routing metadata to scoring callbacks (`sample_weight` → `EpochScoring`)

### Test plan

- [x] `set_fit_request` / `set_partial_fit_request` guards and behavior
- [x] `get_metadata_routing` returns `MetadataRouter` with correct children
- [x] Splitter child present for `GroupKFold`, absent for int cv and `None`
- [x] Extra fit_params reach module forward with routing enabled
- [x] Extra fit_params don't break `ValidSplit` with routing enabled
- [x] `groups` routed to `ValidSplit(GroupKFold)` automatically (no `set_fit_request` needed)
- [x] `clone()` preserves metadata requests
- [x] Pipeline integration with routing enabled
- [x] GridSearchCV integration with routing enabled
- [x] Backward compat: legacy fit_params behavior when routing disabled
- [x] Full existing test suite passes (283 tests, 0 regressions)

__Disclaimer__: The code is `claude` generated, but I've reviewed the code, and this is the second iteration of the solution, which I ended up liking, and it matches what we do in sklearn.

cc @BenjaminBossan @tsbinns @DCoupry

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,3 @@ data/
 *.pickle
 *.mat
 *.ipynb
-CLAUDE.md
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ data/
 *.pickle
 *.mat
 *.ipynb
+CLAUDE.md
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -19,10 +19,10 @@ import warnings
 
 import numpy as np
 from sklearn.base import BaseEstimator
-from sklearn.utils._metadata_requests import UNUSED
 from sklearn.utils.metadata_routing import (
     MetadataRouter,
     MethodMapping,
+    UNUSED,
     _routing_enabled,
     process_routing,
 )
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -1460,22 +1460,6 @@ class NeuralNet(BaseEstimator):
         self._metadata_request = requests
         return self
 
-    def _get_splitter_for_routing(self):
-        """Extract the CV splitter from train_split for routing.
-
-        Returns the underlying CV splitter if ``train_split`` is a
-        :class:`.ValidSplit` whose ``cv`` attribute is a sklearn
-        splitter with metadata routing support. Returns ``None``
-        otherwise (e.g. when ``train_split`` is ``None``, a custom
-        callable, or wraps an integer/float cv).
-        """
-        ts = self.train_split
-        if ts is None:
-            return None
-        if hasattr(ts, 'cv') and hasattr(ts.cv, 'get_metadata_routing'):
-            return ts.cv
-        return None
-
     def get_metadata_routing(self):
         """Get metadata routing of this object.
 
@@ -1492,10 +1476,10 @@ class NeuralNet(BaseEstimator):
         router = MetadataRouter(owner=self.__class__.__name__)
         router.add_self_request(self)
 
-        splitter = self._get_splitter_for_routing()
-        if splitter is not None:
+        ts = self.train_split
+        if ts is not None and hasattr(ts, 'cv'):
             router.add(
-                splitter=splitter,
+                splitter=ts.cv,
                 method_mapping=(
                     MethodMapping()
                     .add(caller="fit", callee="split")
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -19,6 +19,13 @@ import warnings
 
 import numpy as np
 from sklearn.base import BaseEstimator
+from sklearn.utils._metadata_requests import UNUSED
+from sklearn.utils.metadata_routing import (
+    MetadataRouter,
+    MethodMapping,
+    _routing_enabled,
+    process_routing,
+)
 import torch
 from torch.utils.data import DataLoader
 
@@ -309,6 +316,13 @@ class NeuralNet(BaseEstimator):
       this list.
 
     """
+    # Suppress auto-generation of set_partial_fit_request that only
+    # allows 'classes'. We provide our own that accepts arbitrary
+    # metadata names, since partial_fit takes **fit_params.
+    # TODO: remove once scikit-learn/scikit-learn#32111 is merged and
+    # provides a public API for this.
+    __metadata_request__partial_fit = {"classes": UNUSED}
+
     prefixes_ = ['iterator_train', 'iterator_valid', 'callbacks', 'dataset', 'compile']
 
     cuda_dependent_attributes_ = []
@@ -1190,8 +1204,26 @@ class NeuralNet(BaseEstimator):
         self.check_training_readiness()
         epochs = epochs if epochs is not None else self.max_epochs
 
+        if _routing_enabled():
+            routed_params = process_routing(self, "fit", **fit_params)
+            if fit_params:
+                split_params = routed_params.get(
+                    "splitter", {"split": {}}
+                )["split"]
+            else:
+                split_params = {}
+            # Following sklearn's router+consumer pattern: the router
+            # uses its own params directly (they're already in
+            # fit_params), while children get theirs from
+            # routed_params. The module's forward method should accept
+            # **kwargs to handle any extra params.
+            forward_params = fit_params
+        else:
+            split_params = fit_params
+            forward_params = fit_params
+
         dataset_train, dataset_valid = self.get_split_datasets(
-            X, y, **fit_params)
+            X, y, **split_params)
         on_epoch_kwargs = {
             'dataset_train': dataset_train,
             'dataset_valid': dataset_valid,
@@ -1205,10 +1237,10 @@ class NeuralNet(BaseEstimator):
             self.notify('on_epoch_begin', **on_epoch_kwargs)
 
             self.run_single_epoch(iterator_train, training=True, prefix="train",
-                                  step_fn=self.train_step, **fit_params)
+                                  step_fn=self.train_step, **forward_params)
 
             self.run_single_epoch(iterator_valid, training=False, prefix="valid",
-                                  step_fn=self.validation_step, **fit_params)
+                                  step_fn=self.validation_step, **forward_params)
 
             self.notify("on_epoch_end", **on_epoch_kwargs)
         return self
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -26,3 +26,4 @@ data/
 *.pickle
 *.mat
 *.ipynb
+CLAUDE.md
```
BenjaminBossan, line 29: Remove, can be handled by global git config.

skorch/net.py
```diff
@@ -2231,6 +2361,24 @@ def _replace_callback(self, name, new_val):
                 break
         setattr(self, 'callbacks_', callbacks_new)
 
+    def __deepcopy__(self, memo):
```
BenjaminBossan, line 2364: So I checked and prior to sklearn 1.8.0, the tests would all pass without this custom `__deepcopy__` method. With 1.8.0, we get:

> AttributeError: Can't pickle local object 'TestMetadataRouting.test_fit_with_extra_params_and_routing.<locals>.RecordingModule'

This seems to stem from `RecordingModule` being defined locally in the test. When I move it to the global scope, the test passes even with sklearn 1.8.0 (other tests with local classes will fail but they all pass when making them global). So it appears that the error comes from a recent change in sklearn that makes something in metadata routing incompatible with locally defined classes.

To me, it would be okay to have a `__deepcopy__` method here but it seems there is something bigger going on here.

skorch/net.py
```diff
@@ -1336,6 +1365,102 @@ def fit(self, X, y=None, **fit_params):
...
+        self._metadata_request = requests
+        return self
+
+    def get_metadata_routing(self):
+        """Get metadata routing of this object.
+
+        NeuralNet is both a :term:`consumer` (its module's forward
+        method accepts arbitrary metadata) and a :term:`router` (it
+        routes metadata like ``groups`` to its internal CV splitter).
+
```
BenjaminBossan, line 1442: I don't think the `:term:` links will juts work, will they? We cross link sklearn docs, but how would rtd know where to map those?

skorch/net.py
```diff
@@ -19,6 +19,13 @@
 
 import numpy as np
 from sklearn.base import BaseEstimator
+from sklearn.utils._metadata_requests import UNUSED
+from sklearn.utils.metadata_routing import (
+    MetadataRouter,
+    MethodMapping,
+    _routing_enabled,
+    process_routing,
+)
```
BenjaminBossan, line 28: I wasn't sure how far back these imports reach, but I checked with sklearn 1.4.0 and the tests passed.

skorch/net.py
```diff
@@ -2231,6 +2361,24 @@ def _replace_callback(self, name, new_val):
...
         setattr(self, 'callbacks_', callbacks_new)
 
+    def __deepcopy__(self, memo):
+        # NeuralNet's __getstate__ uses pickle.dump for CUDA-dependent
+        # attributes, which blocks or fails when the module is not
+        # picklable. We bypass __getstate__/__setstate__ and deepcopy
+        # each attribute individually, falling back to shallow copy
+        # for non-copyable objects (e.g. torch modules with locally
+        # defined classes).
+        import copy
```
BenjaminBossan, line 2371: Can be global import

skorch/net.py
```diff
@@ -1336,6 +1365,102 @@ def fit(self, X, y=None, **fit_params):
...
+        Returns
+        -------
+        routing : MetadataRouter
+            A :class:`~sklearn.utils.metadata_routing.MetadataRouter`
+            encapsulating routing information.
+        """
+        router = MetadataRouter(owner=self.__class__.__name__)
+        router.add_self_request(self)
+
+        ts = self.train_split
```
BenjaminBossan, line 1452: nice

skorch/tests/test_net.py
```diff
@@ -4300,3 +4300,236 @@ def forward(self, input):
...
+        with pytest.raises(RuntimeError, match="metadata routing is enabled"):
+            net.set_partial_fit_request(Z=True)
+
+    def test_set_fit_request_returns_self(
+        self, net_cls, module_cls, routing_enabled
+    ):
+        net = net_cls(module_cls)
+        assert net.set_fit_request(Z=True) is net
+
+    def test_fit_with_extra_params_and_routing(
```
BenjaminBossan, line 4347: Could you please add a similar test for `partial_fit`?

BenjaminBossan (changes requested):
Thanks for adding support for metaddata routing to skorch. Overall, this looks good, but I have a few comments, please check.

> We probably can add a Mixin in scikit-learn so that you could avoid using the private type of API call.

So I assume this hasn't happened (yet)?

adrinjalali:
> > We probably can add a Mixin in scikit-learn so that you could avoid using the private type of API call.
> 
> So I assume this hasn't happened (yet)?

Yeah haven't managed to get it in. It's been not very busy on the metadata routing side on sklearn, since we've been focusing on other projects, but it also means the issues haven't been too urgent, which is a good thing.

The `__deepcopy__` issue, however, is real and I'd like to fix it in sklearn, we shouldn't be deepcopy-ing estimators there.

skorch/net.py
adrinjalali: fix on sklearn side: https://github.com/scikit-learn/scikit-learn/pull/33827

skorch/net.py
```diff
@@ -1290,9 +1323,15 @@ def partial_fit(self, X, y=None, classes=None, **fit_params):
         if not self.initialized_:
             self.initialize()
 
+        # When called from fit(), _routing_method is threaded in via
```
BenjaminBossan, line 1326: Ah, too bad that this is needed, but it is what it is.

BenjaminBossan (approved):
Thanks for your updates, Adrin, and for working on the fix in sklearn too. PR LGTM.

As it's not an urgent change, I'll leave it open for a week or so in case Thomas or someone else wants to review as well.

tsbinns:
> As it's not an urgent change, I'll leave it open for a week or so in case Thomas or someone else wants to review as well.

Thanks both for the work!
Sure, could try in the next days to see if the workarounds I'm currently using are redundant now.

tsbinns:
> Sure, could try in the next days to see if the workarounds I'm currently using are redundant now.

Just had a go and no more workarounds needed. TYSM @adrinjalali & @BenjaminBossan!

tsbinns:
Just so I can sort any env changes: is there a release planned soon, or will this live in main for a while?

BenjaminBossan:
> Just had a go and no more workarounds needed.

Thanks for testing.

> Just so I can sort any env changes: is there a release planned soon, or will this live in main for a while?

I think we can do a release soon. It depends a bit on co-maintainer availability, so I can't promise.

BenjaminBossan:
@tsbinns It took a bit of time but the [v1.4.0 release](https://github.com/skorch-dev/skorch/releases/tag/v1.4.0) is out.

PR #1141, opened 2026-05-17
ENH Add automatic device selection

### LINKED ISSUE #576: device='auto' should select the fastest possible device

adit24dhaya:
## Summary

Add support for `device='auto'` so skorch selects CUDA when it is available and falls back to CPU otherwise.

This centralizes the resolution in `skorch.utils` and applies it to tensor/device movement, map-location handling, module initialization, and safetensors loading.

Closes #576.

## Testing

- `.venv/bin/python -m pytest skorch/tests/test_utils.py::TestToTensor::test_device_setting_auto skorch/tests/test_utils.py::TestToDevice::test_check_device_auto skorch/tests/test_utils.py::TestToDevice::test_get_map_location_auto skorch/tests/test_net.py::TestNeuralNet::test_device_auto`
- `.venv/bin/python -m pytest skorch/tests/test_utils.py skorch/tests/test_net.py::TestNeuralNet::test_device_torch_device skorch/tests/test_net.py::TestNeuralNet::test_pickle_save_load_device_is_none`
- `git diff --check`

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -48,6 +48,7 @@ from skorch.utils import _infer_predict_nonlinearity
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import TeeGenerator
 from skorch.utils import _check_f_arguments
+from skorch.utils import _check_device_auto
 from skorch.utils import check_is_fitted
 from skorch.utils import duplicate_items
 from skorch.utils import get_map_location
@@ -218,7 +219,8 @@ class NeuralNet(BaseEstimator):
     device : str, torch.device, or None (default='cpu')
       The compute device to be used. If set to 'cuda' in order to use
       GPU acceleration, data in torch tensors will be pushed to cuda
-      tensors before being sent to the module. If set to None, then
+      tensors before being sent to the module. If set to 'auto', CUDA
+      is used if available and CPU otherwise. If set to None, then
       all compute devices will be left unmodified.
 
     compile : bool (default=False)
@@ -2719,6 +2721,9 @@ class NeuralNet(BaseEstimator):
             warnings.warn(msg, DeviceWarning)
             return map_device
 
+        if requested_device == 'auto':
+            return map_device
+
         type_1 = torch.device(requested_device)
         type_2 = torch.device(map_device)
         if type_1 != type_2:
@@ -2796,7 +2801,9 @@ class NeuralNet(BaseEstimator):
 
                 if isinstance(f_name, (str, os.PathLike)):
                     state_dict = {}
-                    with safe_open(f_name, framework='pt', device=self.device) as f:
+                    with safe_open(
+                            f_name, framework='pt',
+                            device=_check_device_auto(self.device)) as f:
                         for key in f.keys():
                             state_dict[key] = f.get_tensor(key)
                 else:
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -48,7 +48,7 @@ from skorch.utils import _infer_predict_nonlinearity
 from skorch.utils import FirstStepAccumulator
 from skorch.utils import TeeGenerator
 from skorch.utils import _check_f_arguments
-from skorch.utils import _check_device_auto
+from skorch.utils import _check_device
 from skorch.utils import check_is_fitted
 from skorch.utils import duplicate_items
 from skorch.utils import get_map_location
@@ -2803,8 +2803,9 @@ class NeuralNet(BaseEstimator):
                 if isinstance(f_name, (str, os.PathLike)):
                     state_dict = {}
                     with safe_open(
-                            f_name, framework='pt',
-                            device=_check_device_auto(self.device)) as f:
+                            f_name,
+                            framework='pt',
+                            device=_check_device(self.device)) as f:
                         for key in f.keys():
                             state_dict[key] = f.get_tensor(key)
                 else:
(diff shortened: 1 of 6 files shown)


skorch/utils.py
```diff
@@ -60,6 +60,12 @@ def is_geometric_data_type(x):
 
 
 # pylint: disable=not-callable
+def _check_device_auto(device):
+    if device == 'auto':
+        return 'cuda' if torch.cuda.is_available() else 'cpu'
```
BenjaminBossan, line 65: This function "steals" the `# pylint: disable=not-callable` from `to_tensor`.

BenjaminBossan (changes requested):
Thanks for creating this PR to support `device='auto'` in skorch. It already looks quite complete. I have a small comment, please check.

Moreover, I think we need to add a more complete test than the ones you've added. By that, I mean something like this test:

https://github.com/skorch-dev/skorch/blob/5db0ddc962117a2c3a646910ecd2e70253ff8e87/skorch/tests/test_net.py#L338-L347

The reason is that only checking if the model weight is on the correct device is not enough. There are other places where there could be an error due to mismatched devices, so having a test that is more end-to-end is important.

What I also wondered about while checking this PR is if we can take the opportunity to be more open to non-CUDA devices like XPU. It won't be easy to test, as the CI doesn't have access to those machines and I also don't have one at home. But it could still be a nice addition. LMK what you think.

Regardless of whether that is added, I think it's a good idea to use a more general wording in the documentation, e.g. replacing "If set to ``'auto'``, CUDA is used" with "If set to ``'auto'``, hardware acceleration like CUDA is being used". That way, we don't need to rewrite all these sections if we eventually add support for other devices.

adit24dhaya:
Thanks for the review. I pushed 441d2d1 with the pylint directive moved back to to_tensor, an end-to-end fit/predict test for device=auto, and broader docs wording around hardware acceleration.

On XPU and other accelerators, I would prefer to keep this PR focused on the CUDA-to-CPU fallback and handle additional accelerators in a follow-up once we agree on the detection order.

skorch/net.py
```diff
@@ -2796,7 +2802,9 @@ def _get_state_dict(f_name):
 
                 if isinstance(f_name, (str, os.PathLike)):
                     state_dict = {}
-                    with safe_open(f_name, framework='pt', device=self.device) as f:
+                    with safe_open(
+                            f_name, framework='pt',
```
BenjaminBossan:
```suggestion
                            f_name,
                            framework='pt',
```

skorch/utils.py
```diff
@@ -59,6 +59,12 @@ def is_geometric_data_type(x):
     return isinstance(x, Data)
 
 
+def _check_device_auto(device):
```
BenjaminBossan: Let's rename the function to just `_check_device`, as it may do more than checking for `auto` in the future. Also, let's add a short docstring.

skorch/tests/test_net.py
```diff
@@ -346,6 +346,18 @@ def test_net_learns(self, net_cls, module_cls, data):
         y_pred = net.predict(X)
         assert accuracy_score(y, y_pred) > ACCURACY_EXPECTED
 
+    def test_device_auto_fit_predict(self, net_cls, module_cls, data):
```
BenjaminBossan: Let's parametrize this function similar to the other tests that you added. Also please check the device of the modules. Finally, let's include a call to `y_forward = net.forward(X, device='auto')` and check the device of the returned tensor.

BenjaminBossan (changes requested):
Thanks for the updates. I still found a few parts that could be improved, but otherwise this looks good.

Regarding the support for other devices, I'm fine with only adding CUDA for now. The rest can be worked on in the future.

adit24dhaya:
Thanks again for the follow-up review. I pushed `566a92c` with the requested changes:

- renamed `_check_device_auto` to `_check_device` and added a short docstring
- split the `safe_open` arguments across lines
- parametrized and strengthened the `device='auto'` end-to-end test to check module devices and `forward(..., device='auto')` output device

I also reran the focused and broader affected tests locally:

- `.venv/bin/python -m pytest skorch/tests/test_net.py::TestNeuralNet::test_device_auto_fit_predict skorch/tests/test_net.py::TestNeuralNet::test_device_auto skorch/tests/test_utils.py::TestToTensor::test_device_setting_auto skorch/tests/test_utils.py::TestToDevice::test_check_device_auto skorch/tests/test_utils.py::TestToDevice::test_get_map_location_auto`
- `.venv/bin/python -m pytest skorch/tests/test_utils.py skorch/tests/test_net.py::TestNeuralNet::test_device_auto_fit_predict skorch/tests/test_net.py::TestNeuralNet::test_device_auto skorch/tests/test_net.py::TestNeuralNet::test_device_torch_device skorch/tests/test_net.py::TestNeuralNet::test_pickle_save_load_device_is_none`
- `git diff --check`

BenjaminBossan (approved):
Sorry for the delay. The changes LGTM, thanks for the contribution.

adit24dhaya:
Thanks for the review and merge — the feedback on tests and scoping made the final version much better. Appreciate it!

PR #1144, opened 2026-05-19
Log param group assignments during optimizer initialization

### LINKED ISSUE #291: Print param groups after initialization
We should be verbose about the param groups that are added after initialization (since we support wildcards and it may be intransparent which param groups are created from that).

algojogacor:
This is an independent contribution made in good faith. Not generated by or affiliated with any automated or competitive process.

## Summary

When using wildcard-based param groups (e.g. `optimizer__param_groups=[('dense*', {'lr': 0.03})]`), it is non-obvious which named parameters are matched to each group. This adds logger.info calls that report each assignment at initialization time.

## Root Cause

The `_get_params_for_optimizer` method silently processes wildcard patterns against named parameters. Users have no visibility into how their patterns resolve.

## Fix

Added a `logging` import, a module-level `logger`, and an `info`-level log statement in `_get_params_for_optimizer` that reports the pattern, the group kwargs, and the matched parameter names for each param group assignment.

Changes: 7 lines in `skorch/net.py`.

## Testing

- Unit tests continue to pass
- Manual verification: creating a net with wildcard param groups and enabling INFO-level logging produces output like:

```
Setting param group '{'lr': 0.03}' with dense* for dense0.weight, dense0.bias, dense1.weight, dense1.bias
```

Closes #291

### DIFF
--- a/skorch/net.py
+++ b/skorch/net.py
@@ -7,6 +7,7 @@ sklearn-conforming classes like NeuralNetClassifier.
 """
 
 import fnmatch
+import logging
 from collections.abc import Mapping
 from functools import partial
 from itertools import chain
@@ -16,6 +17,8 @@ import os
 import pickle
 import tempfile
 import warnings
+
+logger = logging.getLogger(__name__)
 
 import numpy as np
 from sklearn.base import BaseEstimator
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
The pattern of using `logging` does not match what skorch does anywhere else. The linked issue discusses idiomatic ways of resolving this. Also, IMO, it's not a high priority issue as users can check what has been matched by inspecting `net.optimizer_.param_groups`.

PR #1149, opened 2026-07-15
feat: print optimizer param groups when verbose

### LINKED ISSUE #291: Print param groups after initialization

aiedwardyi:
Prints which module parameters match each optimizer param group when verbose is set, both at init and on set_params. Fixes #291.

## Review fixes applied (8eeac24)
- Added tests for no / multiple param groups and duplicate-print guards.
- Reverted the unnecessary `optimizer` local in `optimizer_setter`.
- `format_param_group_msg` now reports params not among the module's learnable params instead of hiding them.
- Gated `set_params` printing to a specific param group so global sets (e.g. `optimizer__lr`) stay quiet.

## Review fixes applied (1608a15)
- Truncate the verbose message to `MAX_PARAM_GROUP_MSG_LEN` (200 chars).
- Combined the verbose print tests into one parametrized test that runs a short fit to guard against repeated messages during training.

### DIFF
--- a/skorch/setter.py
+++ b/skorch/setter.py
@@ -4,6 +4,13 @@ import re
 
 def format_param_group_msg(group_config, param_names):
     """Message for which module params a param group config applies to."""
+    if not param_names:
+        return (
+            "Setting param group {} for parameters that are not among the "
+            "module's learnable parameters (this may be unintended).".format(
+                group_config,
+            )
+        )
     return "Setting param group {} for {}.".format(
         group_config,
         ', '.join(param_names),
@@ -85,18 +92,18 @@ def optimizer_setter(
         param_group, param_name = _extract_optimizer_param_name_and_group(
             optimizer_name, param)
 
-    optimizer = getattr(net, optimizer_attr)
     groups = _set_optimizer_param(
-        optimizer=optimizer,
+        optimizer=getattr(net, optimizer_attr),
         param_group=param_group,
         param_name=param_name,
         value=value
     )
 
-    if getattr(net, 'verbose', 0):
+    # only report for a specific param group; a global set (e.g. optimizer__lr)
+    # touches every param and is not what #291 asks to surface
+    if getattr(net, 'verbose', 0) and param_group != 'all':
         tensors = []
         for group in groups:
             tensors.extend(group.get('params', []))
         param_names = _param_names_for_tensors(net, tensors)
-        if param_names:
-            print(format_param_group_msg({param_name: value}, param_names))
+        print(format_param_group_msg({param_name: value}, param_names))
--- a/skorch/setter.py
+++ b/skorch/setter.py
@@ -1,20 +1,27 @@
 """Setter functions for virtual params such as ``optimizer__lr``."""
 import re
 
+# param names can be arbitrarily long, keep the verbose message bounded
+MAX_PARAM_GROUP_MSG_LEN = 200
+
 
 def format_param_group_msg(group_config, param_names):
     """Message for which module params a param group config applies to."""
     if not param_names:
-        return (
+        msg = (
             "Setting param group {} for parameters that are not among the "
             "module's learnable parameters (this may be unintended).".format(
                 group_config,
             )
         )
-    return "Setting param group {} for {}.".format(
-        group_config,
-        ', '.join(param_names),
-    )
+    else:
+        msg = "Setting param group {} for {}.".format(
+            group_config,
+            ', '.join(param_names),
+        )
+    if len(msg) > MAX_PARAM_GROUP_MSG_LEN:
+        msg = msg[:MAX_PARAM_GROUP_MSG_LEN - 3] + '...'
+    return msg
 
 
 def _param_names_for_tensors(net, tensors):
--- a/skorch/setter.py
+++ b/skorch/setter.py
@@ -2,6 +2,27 @@
 import re
 
 
+def format_param_group_msg(group_config, param_names):
+    """Message for which module params a param group config applies to."""
+    return "Setting param group {} for {}.".format(
+        group_config,
+        ', '.join(param_names),
+    )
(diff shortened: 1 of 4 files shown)


skorch/tests/test_net.py
githubnemo, line 1: the tests should also cover a) output for no param groups, b) output for >1 param groups.

we should also make sure that the messages are not duplicated (i.e. if only one group exists, there should only be one print).

skorch/setter.py
```diff
@@ -62,9 +85,18 @@ def optimizer_setter(
         param_group, param_name = _extract_optimizer_param_name_and_group(
             optimizer_name, param)
 
-    _set_optimizer_param(
-        optimizer=getattr(net, optimizer_attr),
+    optimizer = getattr(net, optimizer_attr)
+    groups = _set_optimizer_param(
+        optimizer=optimizer,
```
githubnemo: This change can be reverted.

skorch/setter.py
```diff
@@ -62,9 +85,18 @@ def optimizer_setter(
...
         value=value
     )
+
+    if getattr(net, 'verbose', 0):
+        tensors = []
+        for group in groups:
+            tensors.extend(group.get('params', []))
+        param_names = _param_names_for_tensors(net, tensors)
+        if param_names:
+            print(format_param_group_msg({param_name: value}, param_names))
```
githubnemo: I think hiding the fact that there are parameters set that are not present in `get_all_learnable_params` is misleading. I think we should handle the case of `len(param_names) == 0` in `format_param_group_msg`. For example, stating that param group X was updated and consists of unknown parameters. It could even include a statement that this is possibly an issue and possibly not intended.

githubnemo (commented):
Thanks for taking this on :)

The changes look good and nice to see tests from the start. I've made some comments.

skorch/tests/test_net.py
aiedwardyi: Added `test_optimizer_param_groups_verbose_no_param_groups` and `test_optimizer_param_groups_verbose_multiple_groups`, plus `out.count('Setting param group')` assertions (1 for a single group, 2 for multiple) to catch duplicate prints. Fixed in 8eeac24.

skorch/setter.py
aiedwardyi: Reverted - inlined `getattr(net, optimizer_attr)` back into the `_set_optimizer_param` call. Fixed in 8eeac24.

skorch/setter.py
aiedwardyi: `format_param_group_msg` now handles the empty case: when no names map to learnable params it returns a `... for parameters that are not among the module's learnable parameters (this may be unintended).` message, and the caller prints unconditionally instead of silently skipping. Fixed in 8eeac24.

aiedwardyi:
Thanks for the review, @githubnemo! Addressed all three:

- **Tests** - added no-param-groups and >1-group cases, plus `count()` assertions so duplicate prints are caught.
- **Revert** - inlined `getattr(net, optimizer_attr)` back in `optimizer_setter`.
- **Unmapped params** - `format_param_group_msg` now surfaces them explicitly instead of hiding them.

While adding the tests I also noticed `set_params` printed on global optimizer sets (e.g. `optimizer__lr`), which broke the reinitialization-message tests and would be noisy mid-training - I gated printing to a specific param group, matching #291's intent. All setter and param-group tests pass locally.

skorch/setter.py
```diff
@@ -2,6 +2,34 @@
...
+    if not param_names:
+        return (
+            "Setting param group {} for parameters that are not among the "
+            "module's learnable parameters (this may be unintended).".format(
+                group_config,
+            )
+        )
+    return "Setting param group {} for {}.".format(
+        group_config,
+        ', '.join(param_names),
```
githubnemo: We should probably limit the output of this in case there are a lot of parameters in that group. Since the parameters themselves can be arbitrarily long it might be the best way to truncate the final string to a fixed length. WDYT?

skorch/tests/test_net.py
```diff
@@ -1396,6 +1396,61 @@ def test_optimizer_param_groups(self, net_cls, module_cls):
...
+            ],
+        )
+        net.initialize()
+        out = capsys.readouterr().out
+        # one message per configured group, no duplicates
+        assert out.count('Setting param group') == 2
+        assert "Setting param group {'lr': 0.1} for" in out
+        assert "Setting param group {'lr': 0.5} for" in out
+        assert 'sequential.0.weight' in out
+        assert 'sequential.3.weight' in out
```
githubnemo: I think we can combine `test_optimizer_param_groups_verbose_multiple_groups` and `test_optimizer_param_groups_verbose_prints` into one test and use `pytest.mark.parametrize` for the param groups and expected counts.

I'd also include a (short) training run on dummy data to make sure that there are no accidental repetitions during training.

githubnemo (commented):
Thanks for the changes, they look reasonable!

Two more comments but otherwise LGTM

skorch/setter.py
aiedwardyi: Agreed, fixed-length truncation is the simplest guard. Added `MAX_PARAM_GROUP_MSG_LEN = 200`; `format_param_group_msg` now truncates the final string to that length with a trailing `...`. Fixed in 1608a15.

skorch/tests/test_net.py
aiedwardyi: Combined into one parametrized test covering 0/1/2 groups with expected message counts, and it now runs a short 2-epoch `fit()` on dummy data so repetitions during training would be caught. Fixed in 1608a15.

githubnemo (approved):
LGTM. Thanks :)

PR #1150, opened 2026-07-16
Fix LRScheduler pickling dragging along CUDA optimizer reference

### LINKED ISSUE #1096: Transfer skorch models from CUDA to CPU for inference
Hi all, I am training skorch models locally in a CUDA-enabled torch environment, and, if possible, I would like to transfer the entirety of the model to CPU so that they can be registered and used for inference in a CPU-only environment. Is there a best method for accomplishing this? 

I'm pretty new to skorch and deep learning so I'm not sure if this is even possible, but, if so, a skorch helper method for converting a model to CPU would be a nice-to-have feature.

Edit: Just noticed a very similar (old) issue that is still open at time of posting (https://github.com/skorch-dev/skorch/issues/553). The conversation there didn't seem to completely resolve. Let me know if should post there or if reviving this topic here would be preferable.

pctablet505:
Fixes #1096.

`net.optimizer_` is handled as a CUDA-dependent attribute when pickling a net, so it can be safely loaded on a CPU-only machine. However, when using the `LRScheduler` callback, `lr_scheduler_.optimizer` is the same optimizer object, and this reference lives inside `net.callbacks_`, which is not covered by that CUDA-dependent handling. Pickling the net therefore serialized a second, independent copy of the optimizer (including any device tensors, e.g. Adam's running averages) through the ordinary pickle path, causing CUDA-trained nets with an `LRScheduler` callback to fail to load on CPU-only machines.

`lr_scheduler_` is recreated from `net.optimizer_` in `on_train_begin` anyway, so it doesn't need to be part of the pickled state. This adds a small `__getstate__` to `LRScheduler` that drops it before pickling, following the same pattern already used by `EarlyStopping` and `ProgressBar` for their own transient state.

Added a test that reproduces the reference leak with a stateful optimizer (Adam) and confirms it's gone after a pickle round-trip, and that resuming training afterwards still works correctly.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `ValidSplit` now raises a clear error when it receives an `IterableDataset`, instead of an opaque `TypeError` about a missing length (#594)
 - `Checkpoint`'s sink message now reports the epoch the checkpoint was actually taken in, instead of the following one
+- Fix pickling a net with an `LRScheduler` callback dragging along a reference to the (potentially CUDA-resident) optimizer, which could prevent loading nets trained on CUDA on a CPU-only machine (#1096)
 
 ## [1.4.0]
 
--- a/skorch/callbacks/lr_scheduler.py
+++ b/skorch/callbacks/lr_scheduler.py
@@ -124,6 +124,19 @@ class LRScheduler(Callback):
         self.batch_idx_ = 0
         return self
 
+    def __getstate__(self):
+        # Don't pickle lr_scheduler_: Torch's LR schedulers keep a reference
+        # to the optimizer they were created for, which means that pickling
+        # it would drag along a duplicate of net.optimizer_, including its
+        # device-dependent tensors (e.g. Adam's running averages). This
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
Thanks for fixing this old issue.

> `lr_scheduler_` is recreated from `net.optimizer_` in `on_train_begin` anyway, so it doesn't need to be part of the pickled state.

That was indeed the required ingredient to make the change work. At first I was worried that we'd lose important state by deleting the `lr_scheduler_` attribute, but this not the case.

> Added a test that reproduces the reference leak with a stateful optimizer (Adam) and confirms it's gone after a pickle round-trip, and that resuming training afterwards still works correctly.

The test is a bit indirect, as it doesn't actually test saving a net with CUDA and then loading it on CPU and resuming work. That would be difficult to test though and wouldn't run on the GH CI anyway. I checked that it works locally (also ensuring that it fails without the fix) so at least we have that.

Since the PR is still in draft mode, I'm not sure if it's ready yet. Please let me know.

BenjaminBossan:
@pctablet505 I'm still unsure what the state of the PR is from your side? Is it ready? If yes, please mark it as "ready for review" and ping me.

PR #1151, opened 2026-08-03
fix: raise a clear error when ValidSplit gets an IterableDataset

### LINKED ISSUE #594: torch IterableDataset are not fully supported
With PyTorch 1.2.0 came [`IterableDataset`][1] which only implements `__iter__` but no `__len__` and certainly no `__getitem__`. This is definitely a problem since we are using `Subset` to split the input dataset and wraps the original dataset, introduces `__getitem__` and delegates the call to the wrapped dataset - which doesn't implement that method since it is iterable.

The simplest solution to this is to not split `IterableDataset` in any way. What do you think?

[1]: https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset

aiedwardyi:
Closes #594.

`ValidSplit` needs `len()` and `Subset`, so an `IterableDataset` currently dies with an opaque `TypeError` about a missing length. It now raises a `ValueError` naming the cause and pointing at `train_split=None`.

Per the request on the issue that the error only be raised if really needed, the check sits in `ValidSplit.__call__` rather than in `NeuralNet` - a custom `train_split` that tolerates an `IterableDataset` keeps working, and there is a test covering that.

### Review fixes applied
- 961a68a: intercept the `TypeError` from `get_len` instead of checking `isinstance` up front, so an `IterableDataset` subclass that defines `__len__` still splits normally; any other error is reraised untouched. Added a test for that case.
- 729cdcf: chain the original `TypeError` (`from exc` instead of `from None`) so the traceback keeps it; test asserts `__cause__`.

### DIFF
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -314,14 +314,14 @@ class ValidSplit:
         # pylint: disable=invalid-name
         try:
             len_dataset = get_len(dataset)
-        except TypeError:
+        except TypeError as exc:
             if not isinstance(dataset, torch.utils.data.IterableDataset):
                 raise
             raise ValueError(
                 "Cannot perform a CV split on an IterableDataset because it has "
                 "no length. Set train_split=None to disable the internal "
                 "validation split, or pass a train_split that supports "
-                "IterableDataset.") from None
+                "IterableDataset.") from exc
         if y is not None:
             len_y = get_len(y)
             if len_dataset != len_y:
--- a/skorch/dataset.py
+++ b/skorch/dataset.py
@@ -311,15 +311,17 @@ class ValidSplit:
         if self.stratified and not self._is_stratified(cv):
             raise bad_y_error
 
-        if isinstance(dataset, torch.utils.data.IterableDataset):
+        # pylint: disable=invalid-name
+        try:
+            len_dataset = get_len(dataset)
+        except TypeError:
+            if not isinstance(dataset, torch.utils.data.IterableDataset):
+                raise
             raise ValueError(
                 "Cannot perform a CV split on an IterableDataset because it has "
                 "no length. Set train_split=None to disable the internal "
                 "validation split, or pass a train_split that supports "
-                "IterableDataset.")
-
-        # pylint: disable=invalid-name
-        len_dataset = get_len(dataset)
+                "IterableDataset.") from None
         if y is not None:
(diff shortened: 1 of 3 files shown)


skorch/dataset.py
```diff
@@ -311,6 +311,13 @@ def __call__(self, dataset, y=None, groups=None):
         if self.stratified and not self._is_stratified(cv):
             raise bad_y_error
 
+        if isinstance(dataset, torch.utils.data.IterableDataset):
```
BenjaminBossan: I think this check is too strong, users could implement their own `IterableDataset` subclass and and add a `__len__` method. I would rather intercept the error from `get_len(dataset)`, check if we see an `IterableDataset`, raise the error with the helpful message if yes, otherwise reraise whatever error occurred.

BenjaminBossan (changes requested):
Thanks for the PR. I made a suggestion, please check.

skorch/dataset.py
aiedwardyi: Fixed in 961a68a - now wraps `get_len` and only converts a `TypeError` into the helpful message when the dataset is an `IterableDataset`, reraising otherwise. Added a test with an `IterableDataset` that defines `__len__`/`__getitem__`, which the previous check wrongly rejected.

skorch/dataset.py
```diff
@@ -311,15 +311,17 @@ def __call__(self, dataset, y=None, groups=None):
...
+                raise
             raise ValueError(
                 "Cannot perform a CV split on an IterableDataset because it has "
                 "no length. Set train_split=None to disable the internal "
                 "validation split, or pass a train_split that supports "
-                "IterableDataset.")
-
-        # pylint: disable=invalid-name
-        len_dataset = get_len(dataset)
+                "IterableDataset.") from None
```
BenjaminBossan: Why not raise from the original error?

BenjaminBossan (commented):
Thanks for the update, just one question.

skorch/dataset.py
aiedwardyi: Good catch - fixed in 729cdcf, the ValueError now chains `from exc` and a test asserts `__cause__` is the original TypeError.

BenjaminBossan (approved):
Thanks for the update, LGTM.

PR #1152, opened 2026-08-08
Improve error message and add FAQ entry for stratified CV with Dataset (#1122)

### LINKED ISSUE #1122: NeuralNetClassifier doesn't fit to a `skorch.dataset.Dataset` with the default `train_split`
### Problem

The Skorch docs claim that it is possible to pass a Skorch or PyTorch `Dataset` to the `.fit()` method of a `NeuralNet`, with the caveat that `y=None` must be passed explicitly with the subclasses `NeuralNetClassifier` and `NeuralNetRegressor`.

However, this currently fails with a non-obvious error unless you override the default `train_split` argument.

### Example

Suppose you have this setup:

```python
import numpy as np
from sklearn.datasets import load_breast_cancer
from sklearn.model_selection import train_test_split
from sklearn.preprocessing import minmax_scale

cancer = load_breast_cancer()
X_train, X_test, y_train, y_test = train_test_split(
    cancer.data.astype(np.float32),
    cancer.target,
    stratify=cancer.target
)
train_ds = Dataset(X_train, y_train)
```

Now suppose you have the following simple model:

```python
import skorch
import torch.nn as nn

model = skorch.NeuralNetClassifier(
    MLPModule,
    module__input_units=cancer.data.shape[1],
    module__output_units=len(np.unique(cancer.target)),
    module__hidden_units=1000,
    module__dropout=0.4,
    criterion=nn.CrossEntropyLoss,
)
```

This works:

```python
model.fit(X_train, y_train)
```

But this fails:

```python
model.fit(train_ds, y=None)
```

Error:
```
ValueError: Stratified CV requires explicitly passing a suitable y.
```

### Analysis

The problem arises because the default value for `train_split` is `ValidSplit(5, stratified=True)`, but the code in `ValidSplit.__call__` doesn't handle stratification for a `Dataset` object.

### Possible solution

It would improve usability if Skorch's `NeuralNetClassifier` were just to work with a `Dataset` passed to `.fit()`. Here are some ways to make this happen:

1. If passed a `skorch.dataset.Dataset`, make use of its `.y` attribute for determining class proportions for stratification.
2. If passed some other PyTorch `Dataset`, Skorch could disable stratification and emit an appropriate warning (not an error) to inform the user of the appropriate solution (see (3)).
3. Ideally there would be some way to pass class proportions to Skorch for its validation splitting. All stratification requires is knowing the proportions of each class label in the dataset, which is much less demanding than knowing all the actual `y` values. (Obtaining all the actual `y` values for a Dataset could be very inefficient.) The `y` proportions could potentially be provided as a dictionary like `{0: 0.4, 1: 0.4, 2: 0.2}` or a dense array from calling `np.bincount(y)` as a kwarg `stratify_proportions=` to `ValidSplit()`.

I'm happy to accept feedback on the proposal and create a PR to solve this.

Sweeyya:
## Why

skorch's docs say you can pass a PyTorch or skorch `Dataset` directly to
`.fit()`. Doing so currently fails with a confusing error
(`ValueError: Stratified CV requires explicitly passing a suitable y`),
because the default validation split logic (`ValidSplit`, stratified by
default on `NeuralNetClassifier`) can't pull class labels out of a `Dataset`
object the way it can from raw arrays. Users hitting this had no indication
of what to do next (see the StackOverflow thread linked in the issue).

Per @BenjaminBossan's guidance on the issue, this fix does not guess `y`
from the dataset or silently disable stratification. It only makes the
existing error message and docs more actionable.

## What changed

- `ValidSplit.__call__` now detects when the failing `dataset` is a
  `skorch.dataset.Dataset` or `torch.utils.data.TensorDataset` and extends
  the error message to name the type and point to three workarounds:
  `predefined_split()`, `train_split=ValidSplit(5, stratified=False)`, or
  `train_split=None`. Any other type still gets the original message,
  unchanged.
- Added a FAQ entry to `docs/user/FAQ.rst` covering this failure and the
  same three workarounds.
- Added a `CHANGES.md` entry under Unreleased.

Closes #1122

## Acceptance criteria

- [x] Tests added (`skorch/tests/test_dataset.py`: `Dataset`, `TensorDataset`,
      and generic array cases)
- [x] All tests passing (`python -m pytest`: 1648 passed, 104 skipped,
      1 xfailed)
- [x] Follows existing style and conventions in `dataset.py` and the test
      files
- [x] No breaking changes. Behavior is unchanged except for the two newly
      detected dataset types, where the message is strictly more
      informative

## Before and after

Before:

ValueError: Stratified CV requires explicitly passing a suitable y.

After:

ValueError: Stratified CV requires explicitly passing a suitable y. You
passed a skorch.dataset.Dataset as dataset, which skorch cannot introspect
to obtain labels for stratification. To resolve this, either wrap your
validation data with skorch.helper.predefined_split and pass it as
train_split, disable stratification with train_split=ValidSplit(5,
stratified=False), or disable the internal validation split entirely with
train_split=None.

## Testing

- Ran the full test suite locally: 1648 passed, 104 skipped, 1 xfailed
- Updated the pre-existing `test_net.py` test that hard coded the old
  message text
- Manually reran the reproduction script from the issue and confirmed the
  new message
- Built the docs locally with `sphinx-build` to confirm the FAQ entry
  renders correctly

@BenjaminBossan, ready for review whenever you have a chance. Thank you for
scoping this out and for the mentorship on my first contribution here.

### DIFF
--- a/docs/user/FAQ.rst
+++ b/docs/user/FAQ.rst
@@ -264,6 +264,58 @@ The ``RandomDataset`` can be passed directly to
     net = NeuralNet(MyModule, criterion=torch.nn.MSELoss)
     net.fit(train_ds)
 
+Why do I get an error about stratified CV when I pass a Dataset?
+------------------------------------------------------------------
+
+By default, :class:`~skorch.classifier.NeuralNetClassifier` performs a
+stratified train/valid split internally (``train_split=ValidSplit(5,
+stratified=True)``). Stratification requires ``y`` so that the split can
+preserve the class proportions in each fold. If you call ``net.fit(dataset)``
+with a single object that already bundles ``X`` and ``y`` -- e.g. a
+:class:`skorch.dataset.Dataset` or a
+:class:`torch.utils.data.TensorDataset` -- instead of calling
+``net.fit(X, y)``, skorch has no way to extract the labels from
+``dataset`` to pass on to scikit-learn's stratified splitter, and raises:
+
+.. code:: text
+
+    ValueError: Stratified CV requires explicitly passing a suitable y. ...
+
+skorch will not guess ``y`` from the dataset or silently turn stratification
+off, since that could hide a real problem. Instead, pick one of the
+following workarounds:
+
+* Wrap your own validation set with
+  :func:`skorch.helper.predefined_split` and pass it via ``train_split``,
+  so skorch doesn't need to split (or stratify) anything itself:
+
+  .. code:: python
+
+      from skorch.helper import predefined_split
+
+      net = NeuralNetClassifier(
+          MyModule,
+          train_split=predefined_split(valid_ds),
+      )
+
+* Disable stratification and let skorch split by index instead:
+
+  .. code:: python
+
+      from skorch.dataset import ValidSplit
+
+      net = NeuralNetClassifier(
+          MyModule,
+          train_split=ValidSplit(5, stratified=False),
+      )
+
+* Disable the internal train/valid split entirely, e.g. if you handle
+  validation yourself outside of ``fit``:
+
+  .. code:: python
+
+      net = NeuralNetClassifier(MyModule, train_split=None)
+
 
 How can I deal with multiple return values from forward?
 --------------------------------------------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `ValidSplit` now gives an actionable error message when `stratified=True`
+  and no `y` is passed together with a `skorch.dataset.Dataset` or
+  `torch.utils.data.TensorDataset`, naming the dataset type and pointing to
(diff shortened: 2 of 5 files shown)


docs/user/FAQ.rst
```diff
@@ -264,6 +264,58 @@ The ``RandomDataset`` can be passed directly to
...
+          train_split=ValidSplit(5, stratified=False),
+      )
+
+* Disable the internal train/valid split entirely, e.g. if you handle
+  validation yourself outside of ``fit``:
+
+  .. code:: python
+
+      net = NeuralNetClassifier(MyModule, train_split=None)
+
```
BenjaminBossan, line 318: Yet another option is to explicitly pass `y`, if the user can somehow retrieve it. Something like this would work: `net.fit(dataset, y=y)`

BenjaminBossan (changes requested):
Thanks for this PR to improve the error message. There is one merge conflict, could you please resolve it? Moreover, there is one more option for users if they encounter this error. This would affect the docs and also the error message.

PR #1155, opened 2026-09-08
Drop the bins entry from plot_param_updates' docstring

VenishPaneliya:
`SkorchDoctor.plot_param_updates` documents a `bins` parameter it doesn't accept:

```python
def plot_param_updates(self, match_fn=None, axes=None, figsize=None, **kwargs):
```

```
        bins : np.ndarray or None (default=None)
          Bins to use for the histogram. If left as ``None``, they are inferred
          from the data.
```

It's a line plot of relative parameter updates over time, not a histogram, so there are no bins to configure. The entry looks carried over from the methods that genuinely have it — `plot_activations`, `plot_gradients`, `plot_activations_over_time` and `plot_gradient_over_time` all take a real `bins` argument and use it for `np.linspace(...)`. `plot_param_updates` is the only one of the six that documents it without having it.

Following the docstring doesn't just get ignored, it fails: `bins` falls through `**kwargs` into `ax.plot(xvec, values, label=key, **kwargs)`, and matplotlib rejects it.

```
>>> ax.plot(np.arange(3), np.arange(3), label="x", bins=None)
AttributeError: Line2D.set() got an unexpected keyword argument 'bins'
```

Docs only — just the four lines removed, no code touched. `skorch/tests/test_doctor.py` passes (42 tests).

One thing I left alone: the `**kwargs` entry here (and in the sibling plot methods) describes `figsize` as something you override through `**kwargs`, even though `figsize` is an explicit parameter in each of those signatures. That's consistent across the methods, so it reads like a deliberate house phrasing rather than a mistake, and I didn't want to churn several docstrings on my own reading of it. Happy to tidy that up separately if you'd like it changed.

I also skipped a `CHANGES.md` entry since this doesn't change behaviour — let me know if you'd prefer one under Fixed.

### DIFF
--- a/skorch/_doctor.py
+++ b/skorch/_doctor.py
@@ -533,6 +533,26 @@ class SkorchDoctor(BaseEstimator):
 
         Plots the training loss and, if present, the validation loss over time.
 
+        Parameters
+        ----------
+        ax : AxesSubplot or None (default=None)
+          By default, a new matplotlib plot is created. If you instead want to
+          plot onto an existing plot, pass it here. Only a single plot is
+          created.
+
+        figsize : tuple of int or None (default=None)
+          Size of the figure. Only used when a new plot is created, i.e. when
+          ``ax`` is ``None``.
+
+        **kwargs
+          You can override remaining plotting arguments like ``lw`` (line
+          width).
+
+        Returns
+        -------
+        ax : AxesSubplot
+          The ax of the plot.
+
         """
         self.check_is_fitted()
 
@@ -580,9 +600,25 @@ class SkorchDoctor(BaseEstimator):
           plot onto an existing plot, pass it here. There should be one subplot
           for each top level module (typically 2).
 
+        histtype : str (default='step')
+          The type of histogram to draw, passed to
+          :func:`matplotlib.pyplot.hist`.
+
+        lw : int (default=2)
+          The line width of the histogram, passed to
+          :func:`matplotlib.pyplot.hist`.
+
         bins : np.ndarray or None (default=None)
           Bins to use for the histogram. If left as ``None``, they are inferred
           from the data.
+
+        density : bool (default=True)
+          If ``True``, draw a probability density instead of raw counts.
+          Passed to :func:`matplotlib.pyplot.hist`.
+
+        figsize : tuple of int or None (default=None)
+          Size of the figure. Only used when a new plot is created, i.e. when
+          ``axes`` is ``None``.
 
         **kwargs
           You can override remaining plotting arguments like ``alpha``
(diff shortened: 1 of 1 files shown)


BenjaminBossan (commented):
Thanks for fixing the docstring, including "bins" there was indeed a mistake.

> One thing I left alone: the `**kwargs` entry here (and in the sibling plot methods) describes `figsize` as something you override through `**kwargs`, even though `figsize` is an explicit parameter in each of those signatures.

This was certainly not intentional, more of an oversight. I think it would make sense to mention another argument there instead of "figsize", something like "lw" (line width). If you're up to it, please update the docstrings there as well.

VenishPaneliya:
Done — though `lw` turned out to only be safe in one of them, so I picked per method based on what each one already forwards.

`plot_activations` and `plot_gradients` pass `lw` explicitly into `ax.hist` (along with `histtype`, `bins`, `density`), so suggesting `lw` there would land users in the same trap as `figsize`:

```
>>> ax.hist(y, label=k, histtype='step', lw=2, bins=5, density=True, **{'lw': 3})
TypeError: Axes.hist() got multiple values for keyword argument 'lw'
```

So:

| method | forwards to | already passes | example now used |
|---|---|---|---|
| `plot_activations`, `plot_gradients` | `ax.hist` | `histtype`, `lw`, `bins`, `density` | `alpha` |
| `plot_param_updates` | `ax.plot` | `label` | `lw` |
| `plot_activations_over_time`, `plot_gradient_over_time` | `ax.fill_between` | `alpha`, `color`, `lw` | `linestyle` |

I checked each replacement actually goes through rather than assuming — `alpha` on `hist`, `lw` on `plot` and `linestyle` on `fill_between` all work, and `alpha` on `fill_between` raises the same duplicate-keyword error, which is why the two `*_over_time` methods got `linestyle` instead.

`skorch/tests/test_doctor.py` still passes (42 tests). The rewrap also happens to fix one pre-existing `E501` on the old `figsize` line; the three flake8 warnings left in that file are on untouched lines and are present on master too.

BenjaminBossan (commented):
Thanks for the update. I went through the docstrings again and found that the arguments are still very inconsistently documented. Ideally, we would have documentation exactly for the arguments that are indeed present. So as an example, in `plot_gradient_over_time`, `lw`, `figsize`, and `color` are not documented. Is that something you'd like to tackle?

VenishPaneliya:
Happy to — done. Every plot method now documents exactly the arguments it takes.

| method | added |
|---|---|
| `plot_loss` | had no Parameters section at all — added `ax`, `figsize`, `**kwargs`, plus a Returns section to match its siblings |
| `plot_activations` | `histtype`, `lw`, `density`, `figsize` |
| `plot_gradients` | `histtype`, `lw`, `density`, `figsize` |
| `plot_param_updates` | `figsize` |
| `plot_activations_over_time` | `lw`, `figsize`, `color` |
| `plot_gradient_over_time` | `lw`, `figsize`, `color` |

I checked it both directions rather than by eye — comparing each signature against its parsed Parameters section, there are now no missing entries and no documented-but-absent ones across all six. numpydoc also parses each of them with zero warnings, and the parameter counts line up with the signatures (e.g. `plot_activations` → 9).

One small wording note: I described `figsize` as "only used when a new plot is created, i.e. when ``axes`` is ``None``", since `_get_axes` returns the passed axes untouched and ignores `figsize` in that case — that seemed worth stating explicitly, but say the word if you'd rather keep it shorter.

`skorch/tests/test_doctor.py` passes (42 tests), and flake8 output on the file is unchanged from master.

VenishPaneliya:
Note on the red CI: the failures are the model-convergence assertions (`assert 0.71 > 0.75`, `assert np.float64(7.85) < 1.0`), and they show up across Python 3.10 / 3.11 / 3.13 and several torch builds in the same run. This PR changes only docstring text in `skorch/_doctor.py` (1 file, no code lines), so it can't affect training accuracy — looks like flakiness rather than anything from here. Happy to rebase if you'd like a clean run.

BenjaminBossan (approved):
Thanks for bringing all the docstrings in line, LGTM.

PR #1156, opened 2026-09-08
Update PyTorch to 2.14

BenjaminBossan:
Drop 2.10.0. Local tests on GPU pass. From the release notes, nothing should affect skorch.


thomasjpfan (approved):
lgtm

PR #1157, opened 2026-09-22
Support Accelerator classes and nested parameters

### LINKED ISSUE #960: Allow passing uninitialized Accelerator when using AccelerateMixin
Right now, when using `AccelerateMixin`, the instantiated `Accelerator` instance needs to be passed:

```python
accelerator = Accelerator(...)
net = AcceleratedNet(
    MyModule,
    accelerator=accelerator,
)
net.fit(X, y)
```

This new feature should allow to pass the `Accelerator` class directly, with possible extra arguments using the `__` notation:

```python
net = AcceleratedNet(
    MyModule,
    accelerator=Accelerator,
    accelerator__mixed_precision='fp16',
)
net.fit(X, y)
```

The change should be implemented in a backwards compatible manner, so that passing the instantiated `Accelerator` keeps on working.

The reason for this change is not so much that we want to, for instance, grid search on the `Accelerator` parameters, which makes little sense. Mainly, it would be for consistency, so that the `Accelerator` can be instantiated and configured in the same way as other components like the module and optimizer.

quinnarnold:
Fixes #960.

Allow `AccelerateMixin` to accept the `Accelerator` class with `accelerator__...` settings, while retaining support for initialized instances. Keep the supplied class or instance in `accelerator`, initialize the runtime object as `accelerator_`, and handle parameter updates and callback initialization through the existing mixin. Update the parameter documentation and changelog.

Validation on CPU:

- `pytest skorch/tests/test_hf.py::TestAccelerate -q --no-cov`: 17 passed, 14 hardware-dependent skips.
- Separate local checks passed class-based training/prediction, nested parameters, cloning, parameter updates/refitting, instance identity, pre-initialization iterator usage, callback setup, and device validation. The class-input reproduction fails on upstream.
- File-wide flake8/pylint findings also occur on the upstream baseline; no new findings were identified.

No repository test files were changed. GPU and distributed execution were not independently tested.

AI assistance.


skorch/hf.py
```diff
@@ -924,17 +925,45 @@ def __init__(
...
+        """Set parameters, reinitializing the net when accelerator settings change."""
+        accelerator_params = {
+            key: kwargs.pop(key) for key in list(kwargs)
+            if key == 'accelerator' or key.startswith('accelerator__')
+        }
+        super().set_params(**kwargs)
+        for key, value in accelerator_params.items():
+            setattr(self, key, value)
+        self._params_to_validate.update(accelerator_params)
+        if accelerator_params and self.initialized_:
```
githubnemo: Can you explain why we call `initialize()` here instead of `_initialize_accelerator`?

skorch/hf.py
```diff
@@ -944,56 +973,64 @@ def _initialize_accelerator(self):
...
+                callback.policy_ = self.accelerator_.prepare(callback.policy_)
 
         self._wrapped_with_accelerator = True
         return self
 
     def initialize(self):
         """Initializes all of its components and returns self."""
-        # this should be the same as the parent class, except for the one marked
-        # line
+        # Initialize the accelerator before callbacks, then prepare the components.
```
githubnemo: The previous comment was a hint to state that `initialize` is a 1:1 copy of `NeuralNet.initialize` and only changed in one line. This hint is now lost, please restore it.

skorch/hf.py
```diff
@@ -944,56 +973,64 @@ def _initialize_accelerator(self):
...
         self.check_training_readiness()
 
+        self.initialize_accelerator()
+        self._wrapped_with_accelerator = False
```
githubnemo: Why do we need `_wrapped_with_accelerator = False` here?

I also don't understand why we need to call `initialize_accelerator` here when it is also done in `_initialize_accelerator` further down?

githubnemo (changes requested):
Hey, thanks for the contribution.

This is still in early stages, there are no tests that test the new behavior and there are some places that I don't understand. 

A quick remark: if you vibecoded this, please refrain from submitting a PR. We'd like our PR authors to understand what they produced and how their changes integrate into the project. If you just used some AI assistance, go on and let's finish this PR up.

skorch/hf.py
quinnarnold: `_initialize_accelerator()` only prepares the existing components with `accelerator_`; it doesn't create a replacement accelerator, and it returns early if the components are already marked as wrapped. That's why I used full initialization when the accelerator settings change.

I've kept that behavior explicit in the docs, but fixed the unnecessary reset when the same instance is passed again. All parameters are now applied before rebuilding, so combined accelerator/module/optimizer updates use the new settings together. Added regression tests for these cases in fcc1fa7.

skorch/hf.py
quinnarnold: Restored the synchronization comment and marked the accelerator-specific lines in fcc1fa7. There are now three differences from `NeuralNet.initialize`, so the comment points those out instead of saying only one line differs.

skorch/hf.py
quinnarnold: The flag can still be true after an earlier `initialize()` or a fit with `unwrap_after_train=False`. A new initialization creates fresh components, so leaving the flag true would skip preparing them. I've added a repeated-initialization test that fails if this reset is removed.

The two similarly named methods do different jobs: `initialize_accelerator()` creates or reuses the instance before callbacks bind its print method; `_initialize_accelerator()` prepares the modules and optimizer after they've been created. The latter doesn't call the former. I've marked these steps in the code to make the order clearer.

quinnarnold:
Thanks for the review. To clarify, this was AI-assisted, not vibe-coded. I've tested and confirmed the changes on CPU, including how they interact with initialization and callbacks.

You were right about the missing tests. I've added regression coverage in fcc1fa7 for class input, cloning, parameter updates, repeated initialization, and actual scheduler steps. I also caught and fixed a scheduler integration bug: it was still reading the class in `accelerator` instead of the runtime instance in `accelerator_`, which silently prevented the learning rate from changing.

The Accelerate/scheduler tests now report 74 passed and 14 hardware-dependent skips. Related NeuralNet tests report 61 passed and one expected failure. GPU/distributed execution hasn't been validated locally. I've replied to the inline questions and updated the PR description.

skorch/hf.py
```diff
@@ -944,56 +989,65 @@ def _initialize_accelerator(self):
...
         return self
 
     def _initialize_callbacks(self):
-        if self.callbacks__print_log__sink == 'auto':
-            print_func = getattr(self.accelerator, 'print', print)
+        """Use the initialized accelerator's print function for automatic logging."""
+        sink = self.callbacks__print_log__sink
+        if sink == 'auto':
+            if not hasattr(self, 'accelerator_'):
+                self.initialize_accelerator()
```
githubnemo: I think it is a bad idea to have initialization cross-contamination since this increases the complexity of the code by a lot. Make sure that the necessary components are available beforehand through the order defined in `initialize()` instead.

skorch/hf.py
```diff
@@ -944,56 +989,65 @@ def _initialize_accelerator(self):
...
     def initialize(self):
         """Initializes all of its components and returns self."""
-        # this should be the same as the parent class, except for the one marked
-        # line
+        # Keep this in sync with NeuralNet.initialize, except for the marked
+        # accelerator-specific lines below.
         self.check_training_readiness()
 
+        self.initialize_accelerator()  # before callbacks bind accelerator.print
+        self._wrapped_with_accelerator = False  # new components need preparation
```
githubnemo: Why do we need to set `_wrapped_with_accelerator = False` here? It's not set by `initialize_accelerator`. If this is a case of what to do when `initialize` is called anew, then there's a missing `_unwrap_accelerator` call here.

githubnemo (commented):
Thanks for the quick response.

> Thanks for the review. To clarify, this was AI-assisted, not vibe-coded. I've tested and confirmed the changes on CPU, including how they interact with initialization and callbacks.

I'm not concerned about if you tested these changes. I'm concerned whether you **engineered** the changes. I want to review changes that are engineered with intent and understanding. What I don't want to review is stuff that falls out of a language model because you prompted it so with no understanding what the implications are and how well they fit into the existing code. Right now, the changes do not fit well and I think that's because they are not engineered to fit in.

Please give this a consideration before responding with new changes. I'm happy to assist you getting this PR into the repo but they need to be well engineered and not just stuff glued on top. My apologies if I am misjudging what is happening.

quinnarnold:
Ah, got it. Appreciate you explaining further, and sorry for misunderstanding your concerns and taking up extra review time. I’ll close this for now. Thanks for being willing to help!

